PR #18, opened 2026-08-05
fix: execute pnpm with shell on Windows

NediarSK8:
Fixes ENOENT and EINVAL errors when running pnpm via child_process execFileSync on Windows by ensuring it runs with shell: true.

This fixes an issue where running `pnpm run-local` fails immediately on Windows with `ENOENT` or `EINVAL` errors coming from `child_process.execFileSync`.

**Why it was broken**
On Windows, `pnpm` is installed as a batch script (`pnpm.cmd`). Node.js's `execFileSync` cannot execute `.cmd` files natively without invoking a shell (like `cmd.exe`).

**The fix**
I updated the 5 instances of `execFileSync("pnpm", ...)` across the codebase to include `{ shell: process.platform === "win32" }`.

### DIFF
--- a/packages/gatekeeper-context/build-app.mjs
+++ b/packages/gatekeeper-context/build-app.mjs
@@ -15,5 +15,5 @@ console.log(
 execFileSync(
   "pnpm",
   ["exec", "vite", "build", "-c", "vite.config.ts", ...(watch ? ["--watch"] : [])],
-  { cwd: pkgDir, stdio: "inherit" },
+  { cwd: pkgDir, stdio: "inherit", shell: process.platform === "win32" },
 );
--- a/packages/gatekeeper-scheduler/build-app.mjs
+++ b/packages/gatekeeper-scheduler/build-app.mjs
@@ -8,5 +8,5 @@ const watch = process.argv.includes("--watch");
 execFileSync(
   "pnpm",
   ["exec", "vite", "build", "-c", "vite.config.ts", ...(watch ? ["--watch"] : [])],
-  { cwd: packageDirectory, stdio: "inherit" },
+  { cwd: packageDirectory, stdio: "inherit", shell: process.platform === "win32" },
 );
--- a/run-dev-server.js
+++ b/run-dev-server.js
@@ -323,7 +323,7 @@ console.log(`\nStarting: wrangler dev ${args.join(" ")}\n`);
 
 try {
   execFileSync("pnpm", ["exec", "wrangler", "dev", ...args],
-      { stdio: "inherit", cwd: ROOT });
+      { stdio: "inherit", cwd: ROOT, shell: process.platform === "win32" });
 } catch (e) {
   // wrangler was killed or exited with an error; the output was already shown
   // via stdio: "inherit", so just propagate the exit code.
--- a/scripts/release/build-release.mjs
+++ b/scripts/release/build-release.mjs
@@ -42,7 +42,7 @@ function parseArgs(argv) {
 
 function run(command, argv, options = {}) {
   console.log(`running: ${command} ${argv.join(" ")} ${options.cwd ? `(in ${options.cwd})` : ""}`);
-  execFileSync(command, argv, { stdio: "inherit", cwd: ROOT, ...options });
+  execFileSync(command, argv, { stdio: "inherit", cwd: ROOT, shell: process.platform === "win32", ...options });
 }
 
 function gitCommit() {
--- a/scripts/run-local.mjs
+++ b/scripts/run-local.mjs
@@ -115,7 +115,7 @@ const needsInstall = needsBuild || !existsSync(NODE_MODULES);
 
 function run(cmd, args) {
   console.log(`\n> ${cmd} ${args.join(" ")}`);
-  execFileSync(cmd, args, { stdio: "inherit", cwd: ROOT });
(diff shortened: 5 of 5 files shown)


NediarSK8:

snowyukitty:
I hit the same bug independently and filed #19 before spotting this PR — sorry for the duplicate
report, this got here first. One finding from that investigation that I think matters for this
change:

`shell: true` makes Node build a command line by joining the file and arguments with spaces,
without quoting them, so any argument containing a space is re-split by the shell:

```
$ node -e "..." "C:\...\space test dir\hello.txt"

no shell      -> ["C:\\...\\space test dir\\hello.txt"]
shell: true   -> ["C:\\...\\space", "test", "dir\\hello.txt"]
```

That is reachable from the `run-dev-server.js` hunk here. Its `args` are absolute paths built from
`gk.dir`, which comes from `ROOT`:

https://github.com/cloudflare/cloudflare-os/blob/aedcda8/run-dev-server.js#L297-L303

So on a checkout under, say, `C:\Users\Jane Smith\src\cloudflare-os`, wrangler would receive
`-c C:\Users\Jane` and two stray arguments. `scripts/release/build-release.mjs` has the same
exposure through `--outdir <tmpdir>`.

An alternative that avoids the quoting question entirely: `npm_execpath` holds pnpm's own JS entry
point, which `node` can run directly, so no shell is involved and arguments keep their exact values.

```js
      : ["pnpm", args];
}
```

used as `const [file, argv] = pnpmCommand([...]); execFileSync(file, argv, opts);`.

The regex test is worth keeping: under `npm run` that same variable points at npm
(`...\npm\bin\npm-cli.js`), and substituting it unchecked would quietly run `npm install` against a
pnpm workspace. Falling through to plain `"pnpm"` keeps the failure loud instead.

I verified the end-to-end result on Windows 11 / Node 22 — `pnpm run-local` gets from the original
`spawnSync pnpm ENOENT` all the way to `[wrangler:info] Ready on http://127.0.0.1:8787`. Also worth
noting `execFileSync("pnpm.cmd", ...)` is not an option on current Node: it returns `EINVAL` since
the CVE-2024-27980 fix.

Happy for this PR to be the one that lands — just flagging the argument-splitting so it does not
trade one Windows failure for a subtler one. If it is useful I can push my version somewhere for
comparison, or leave a suggestion on the diff.

NediarSK8:
> Happy for this PR to be the one that lands — just flagging the argument-splitting so it does not
> trade one Windows failure for a subtler one. If it is useful I can push my version somewhere for
> comparison, or leave a suggestion on the diff.

Thanks for the heads up. To be honest, this is my first public PR, and you spotted an flaw I'm not sure how to handle properly right now.

Feel free to push your fix directly to this PR branch or suggest on the diff if you'd like. That would help me learn and get this working for everyone.

kentonv:
Hmm, I was considering merging this PR as falling under the "small, trivially-verified" rule from [our contributing policy](https://github.com/cloudflare/cloudflare-os/blob/main/CONTRIBUTING.md). However, from the discussion, it sounds like it's actually more subtle. Nothing is trivial it seems. :(

So this also looks like something we're going to want to throw our own agents at to try to solve. The issue report #19 looks like it covers the details, so let's move the discussion there?

snowyukitty:
Thanks for tackling this. I reproduced the whole flow from a checkout whose path contains a space
(`C:\Users\snowy\cf os\repo`; Windows 11, Node 22.22.3, pnpm 11.17.0).

On unpatched `main`, `pnpm run-local` dies at the first nested call, which is the bug you're fixing:

```
> pnpm install
Error: spawnSync pnpm ENOENT
    at run (.../scripts/run-local.mjs:118:3)
  spawnargs: [ 'install' ]
```

With this PR applied, install and build both succeed — but the run then fails at wrangler:

```
✘ [ERROR] Unknown arguments: os\repo\packages\gatekeeper-confluence\wrangler.dev.jsonc,
  os\repo\packages\gatekeeper-context\wrangler.dev.jsonc, os\repo\packages\gatekeeper-email\...
```

`run-dev-server.js` builds those with `join(gk.dir, "wrangler.dev.jsonc")`, and `gk.dir` descends
from `ROOT`, so each `-c` gets an absolute path. In this run `shell: true` had cmd.exe reparse the
command line and break each one at the space in the checkout path, so `-c` received
`C:\Users\snowy\cf` and the rest of the path arrived as a separate argument. Exit code 1.

I also tested a local alternative that keeps the arguments as an argv array with no shell involved.
With that, the same flow completed and `http://localhost:8787/` returned HTTP 200.

So the spaced-path case still looks like a blocker to me. It's your PR, so rather than hand you an
implementation, here's what the measurements say a fix has to satisfy:

- don't rely on the shell to parse the command line — argument boundaries have to survive
- it has to work from a checkout path containing a space
- it should handle pnpm being either a JS launcher or a native executable, since installs differ
- don't assume any `npm_execpath` value is pnpm's — under `npm run` it points at npm

I'd also avoid quoting the arguments for cmd.exe by hand: beyond spaces there are quotes,
backslashes and metacharacters, and getting the space case right doesn't restore argv semantics.

Happy to go into the approach I tested, or share the rest of the measurements — just ask. Either
way, glad to review another revision.

AI tools assisted this investigation; I ran the reproductions myself and the output above is what
they actually printed.

PR #20, opened 2026-08-05
Fix Cloudflare Access logout

maxwellpeterson:
Fixes logout for deployments behind Cloudflare Access. Clicking "Sign out" currently drops users into an infinite "Authenticating..." state without actually logging them out:
image

image

Instead, we should send them to the [Access logout URL](/cdn-cgi/access/logout) so their session is properly revoked:
image


maxwellpeterson:

ndisidore (approved):

PR #23, opened 2026-08-05
Swap regex for dedicated trailing slash stripper function

ndisidore:
Addresses a few issues flagged by codeql, mostly the same fix applied across the gatekeeper suite


ndisidore:

PR #24, opened 2026-08-05
feat: alchemy deployment — the cloudflare-os package with the OperatingSystem resource

sam-goodwin:
Deploy Cloudflare OS to your own account with [alchemy](https://alchemy.run). Everything is alchemy code, owned by the package it deploys — **the wrangler.jsonc files, the release-manifest machinery, and everything that parsed them are deleted**; each package's `src/alchemy.ts` is the single source of truth for deploy, local dev, and tests.

```typescript
// alchemy.run.ts — your company's repo
import * as Alchemy from "alchemy";
import * as Cloudflare from "alchemy/Cloudflare";
import * as Config from "effect/Config";
import * as Effect from "effect/Effect";
import { OperatingSystem, OperatingSystemData } from "cloudflare-os";
import Context from "cloudflare-os/context";
import GitHub from "cloudflare-os/github";
import Mcp from "cloudflare-os/mcp";

export default Alchemy.Stack(
  "AcmeOS",
  { providers: Cloudflare.providers(), state: Cloudflare.state() },
  Effect.gen(function* () {
    // User data is a separate, protected resource — retained on destroy.
    const data = yield* OperatingSystemData("Data");

    const github = GitHub({
      clientId: Config.string("GITHUB_CLIENT_ID"),
      clientSecret: Config.redacted("GITHUB_CLIENT_SECRET"),
    });

    const os = yield* OperatingSystem("OS", {
      domain: "os.acme.com",
      admins: ["sam@acme.com"],
      data,
      // sign-in providers are references — a typo can't lock you out
      auth: { providers: [github], disablePasswordAuth: true },
      gatekeepers: [
        github,
        Context({ sharingDomain: "production" }),
        Mcp({ enabled: false }),
      ],
      // presence of `ai` = gateway mode; the gateway, its token, and the
      // stored provider keys are provisioned, not referenced
      ai: { providers: { anthropic: Config.redacted("ANTHROPIC_API_KEY") } },
    });

    return { url: os.url };
  }),
);
```

`alchemy deploy` provisions the router (public origin + assets + domain), the workshop-backend, one Worker per enabled gatekeeper, the storage, and the AI gateway. `alchemy destroy` tears it all down — except the data.

### Each package owns its deployment, as code

```typescript
// packages/gatekeeper-github/src/alchemy.ts — the package's root export
export const workerConfig = {
  name: "gatekeeper-github",
  id: "GitHub",
  entry: "github.ts",
  prebuild: "build:configurator",
  compatibilityFlags: ["allow_irrevocable_stub_storage", "nodejs_als"],
  migrations: [
    { tag: "v0", newSqliteClasses: ["UserAccount", "GitHubGatekeeperImpl"] },
  ],
} as const satisfies WorkerConfig;

export const GitHub = (config: GitHubConfig): Gatekeeper =>
  Gatekeeper({
    name: "github",
    oauth: true, // usable as a sign-in provider
    worker: Gatekeeper.Worker(workerConfig, {
      dir: packageRoot(import.meta),
      env: {
        CLIENT_ID: Gatekeeper.secret(config.clientId),
        CLIENT_SECRET: Gatekeeper.secret(config.clientSecret),
      },
    }),
  });

export default GitHub;
```

- The deploy contract (`Gatekeeper`, `Gatekeeper.Worker`, `toWranglerConfig`) lives in `@gadgets/workshop-shared/alchemy`; backend/router/frontend/typed-storage export `<pkg>/alchemy` the same way.
- `cloudflare-os/<name>` re-exports each package's module — same file inside the monorepo and from npm.
- `run-dev-server.js` (local dev, unchanged UX), the integration-test harness, and the pool-workers vitest configs all derive wrangler-shaped config in memory from the exported `workerConfig` — no config files, one source of truth. `scripts/release/` and the golden manifest are deleted.

### A custom gatekeeper is just a Worker

```typescript
const worker = yield* Cloudflare.Worker("Acme", {
  main: "./src/worker.ts",
  compatibility: { date: "2026-02-02" },
  env: { AcmeThing: Cloudflare.DurableObject("AcmeThing") },
});

gatekeepers: [worker]; // GATEKEEPER_ACME derives from the id
```

### Protected user data

```typescript
const data = yield* OperatingSystemData("Data"); // retained on `alchemy destroy`
const data = yield* OperatingSystemData("Data", {
  blueprints: existingKv,     // adopt existing storage
  removalPolicy: "destroy",   // ephemeral stacks (tests, previews) only
});
```

Verified live end to end by `packages/cloudflare-os/test/OperatingSystem.test.ts` (deploy → HTTP probes → destroy), on alchemy's own test harness.

### Status

Draft until an alchemy release ships `Cloudflare.WorkerEntrypoint` (alchemy-run/alchemy#1097, merged); this checkout consumes a local build via pnpm overrides. Intended end state, per review discussion: CI publishes this package as a prebuilt release (worker bundles, frontend variants, checksums), so `npm install cloudflare-os` needs no monorepo checkout — the per-package alchemy code here is the source of truth that pipeline drives.

### DIFF
--- a/packages/alchemy/package.json
+++ b/packages/alchemy/package.json
@@ -1,0 +1,69 @@
+{
+  "name": "cloudflare-os",
+  "version": "0.1.0",
+  "description": "Deploy Cloudflare OS to your own Cloudflare account with alchemy \u2014 one OperatingSystem resource provisions the router, backend, gatekeepers, storage, and AI gateway.",
+  "license": "MIT",
+  "type": "module",
+  "main": "./src/index.ts",
+  "exports": {
+    ".": "./src/index.ts",
+    "./gatekeepers": "./src/gatekeepers/index.ts",
+    "./package.json": "./package.json"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cloudflare/cloudflare-os",
+    "directory": "packages/alchemy"
+  },
+  "scripts": {
+    "types:check": "tsc --noEmit",
+    "clean": "rm -rf dist",
+    "test:e2e": "vitest run",
+    "deploy": "alchemy deploy ./examples/0-quickstart/alchemy.run.ts",
+    "destroy": "alchemy destroy ./examples/0-quickstart/alchemy.run.ts"
+  },
+  "peerDependencies": {
+    "alchemy": ">=2.0.0-beta.67",
+    "effect": ">=4.0.0-beta.100 || >=4.0.0"
+  },
+  "dependencies": {
+    "@gadgets/cloudflare-gatekeeper": "workspace:*",
+    "@gadgets/confluence-gatekeeper": "workspace:*",
+    "@gadgets/gatekeeper-context": "workspace:*",
+    "@gadgets/gatekeeper-scheduler": "workspace:*",
+    "@gadgets/github-gatekeeper": "workspace:*",
+    "@gadgets/google-gatekeeper": "workspace:*",
+    "@gadgets/homeassistant-gatekeeper": "workspace:*",
+    "@gadgets/linear-gatekeeper": "workspace:*",
+    "@gadgets/mcp-gatekeeper": "workspace:*",
+    "@gadgets/mcp-portal-gatekeeper": "workspace:*",
+    "@gadgets/notion-gatekeeper": "workspace:*",
+    "@gadgets/router": "workspace:*",
+    "@gadgets/slack-gatekeeper": "workspace:*",
+    "@gadgets/spotify-gatekeeper": "workspace:*",
+    "@gadgets/supabase-gatekeeper": "workspace:*",
+    "@gadgets/workshop-backend": "workspace:*",
+    "@gadgets/workshop-frontend": "workspace:*",
+    "@gadgets/zoominfo-gatekeeper": "workspace:*",
+    "esbuild": "^0.28.1",
+    "jsonc-parser": "^3.3.1"
+  },
+  "devDependencies": {
+    "@cloudflare/workers-types": "^4.20260702.1",
+    "@effect/vitest": "4.0.0-beta.103",
+    "@gadgets/workshop-shared": "workspace:*",
+    "@types/node": "26.1.0",
+    "alchemy": ">=2.0.0-beta.67",
+    "effect": "4.0.0-beta.103",
+    "typescript": "^5.9.3",
+    "vitest": "^4.1.10",
+    "@effect/platform-node": "4.0.0-beta.103",
+    "ws": "^8.20.0"
+  },
+  "files": [
+    "src",
+    "scripts",
+    "examples",
+    "README.md"
+  ]
+}
--- a/packages/alchemy/src/OperatingSystem.ts
+++ b/packages/alchemy/src/OperatingSystem.ts
@@ -1,0 +1,763 @@
+import * as Cloudflare from "alchemy/Cloudflare";
+import * as Command from "alchemy/Command";
+import type { Input, InputProps } from "alchemy/Input";
+import * as Namespace from "alchemy/Namespace";
+import * as Output from "alchemy/Output";
+import * as Effect from "effect/Effect";
+import { cast } from "effect/Function";
+import * as Redacted from "effect/Redacted";
+import {
+  isWorkerGatekeeper,
+  PUBLIC_BASE_URL,
+  type GatekeeperDeployment,
+} from "./Gatekeeper.ts";
+import { Context } from "./gatekeepers/context.ts";
+import { Mcp } from "./gatekeepers/mcp.ts";
+import { Scheduler } from "./gatekeepers/scheduler.ts";
+import { ownScript, packageDir } from "./paths.ts";
+import { readWranglerConfig } from "./wrangler.ts";
+
+/** The `Cloudflare.AI.Gateway` resource (derived — no deep import). */
+type AIGateway = Effect.Success<ReturnType<typeof Cloudflare.AI.Gateway>>;
+/** The props accepted by `Cloudflare.AI.Gateway`, passed through verbatim. */
+type GatewayProps = Parameters<typeof Cloudflare.AI.Gateway>[1];
+/** A deployed `Cloudflare.Worker` resource. */
+type Worker = Effect.Success<ReturnType<typeof Cloudflare.Worker>>;
+/** A deployed `Cloudflare.KV.Namespace` resource. */
+type KvNamespace = Effect.Success<
+  ReturnType<typeof Cloudflare.KV.Namespace>
+>;
+/** A deployed `Cloudflare.R2.Bucket` resource. */
+type R2Bucket = Effect.Success<ReturnType<typeof Cloudflare.R2.Bucket>>;
+/** The `env` prop shape of `Cloudflare.Worker`. */
+type WorkerEnv = Cloudflare.WorkerBindingProps;
+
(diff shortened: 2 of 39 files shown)


packages/alchemy/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "cloudflare-os",
```
sam-goodwin, line 2: This name might be wrong. Should itbe `@gadgets/alchemy` or something else? Up to you guys how you want to distribute

sam-goodwin:
This term `gatekeepers` doesn't make sense to me. An MCP is a gatekeeper?

Please provide feedback on the desired DX

oiwa-coder:
would love this as well

packages/alchemy/src/wrangler.ts
```diff
@@ -0,0 +1,71 @@
...
+  kvNamespaces: string[];
+  /** Plain vars from `vars`. */
+  vars: Record<string, string>;
+  /** Static-assets routing config from `assets`, when declared. */
+  assets:
+    | { notFoundHandling: string | undefined; runWorkerFirst: string[] | boolean | undefined }
+    | undefined;
+}
+
+interface RawWranglerConfig {
```
ndisidore: This looks like a second, incomplete deployment-config parser. It silently ignores several things the existing release manifest handles or explicitly rejects, so the two deployment paths can drift without anyone noticing.

Could this consume the existing release manifest instead? Otherwise, I think it needs to share the same strict parser and fail on anything it doesn’t understand.

packages/alchemy/src/wrangler.ts
```diff
@@ -0,0 +1,71 @@
...
+    const file = NodePath.join(dir, "wrangler.jsonc");
+    const raw = parseJsonc(readFileSync(file, "utf8")) as RawWranglerConfig;
+    if (typeof raw?.main !== "string") {
+      throw new Error(`${file} has no "main" — not a deployable worker config`);
+    }
+    return {
+      main: NodePath.join(dir, raw.main),
+      compatibilityDate: raw.compatibility_date,
+      compatibilityFlags: raw.compatibility_flags ?? [],
+      durableObjects: (raw.migrations ?? []).flatMap(
```
ndisidore: Possible 🐛 
Flattening the migration history down to class names loses tags, ordering, renames, deletes, and transfers. That’s probably enough for a fresh deployment today, but not for upgrades or adoption.

I think we need to preserve the full DO lifecycle and test deploy → write data → upgrade.

packages/alchemy/src/OperatingSystem.ts
```diff
@@ -0,0 +1,763 @@
...
+      backendEnv[service.binding] = Cloudflare.WorkerEntrypoint(
+        service.worker,
+        {
+          entrypoint: service.entrypoint,
+          ...(service.props ? { props: substituteProps(service.props) } : {}),
+        },
+      );
+    }
+    // Durable Object classes are reached via `ctx.exports`
+    // (`enable_ctx_exports`); the bindings exist to declare the classes and
```
ndisidore: I don't think any services actually use this compat flag so this may require updates to all the gatekeepers

packages/alchemy/src/OperatingSystem.ts
```diff
@@ -0,0 +1,763 @@
...
+  // other prop accepts Inputs.
+  props: InputProps<
+    OperatingSystemProps,
+    "domain" | "auth" | "gatekeepers" | "ai" | "backend" | "frontend"
+  > = {},
+): Effect.Effect<OperatingSystem, never, Cloudflare.Providers> =>
+  Effect.gen(function* () {
+    const gatekeeperSet = resolveGatekeepers(props.gatekeepers);
+
+    // ── Storage ────────────────────────────────────────────────────────
```
ndisidore: Does this mean that they’re owned internally by the convenience component?

I’d rather have a separate protected `OperatingSystemData` resource, with support for caller-owned existing storage and explicit destructive deletion.

This would own:
- Blueprint KV
- Avatar KV
- Blueprint R2
- Any long-lived shared storage

packages/alchemy/package.json
ndisidore: Probably `@cloudflare/cloudflare-os-alchemy` or the like (I know its a mouthful)

packages/alchemy/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "clean": "rm -rf dist",
+    "test:e2e": "vitest run",
+    "deploy": "alchemy deploy ./examples/0-quickstart/alchemy.run.ts",
+    "destroy": "alchemy destroy ./examples/0-quickstart/alchemy.run.ts"
+  },
+  "peerDependencies": {
+    "alchemy": ">=2.0.0-beta.67",
+    "effect": ">=4.0.0-beta.100 || >=4.0.0"
+  },
+  "dependencies": {
```
ndisidore, line 29: This package depends on unpublished workspace packages and then runs their builds in the consumer’s environment. Those builds rely on dev dependencies and monorepo-root scripts that won’t be present after a normal npm install.

What are your thoughts on the npm package carrying prebuilt, checksummed Worker and frontend artifacts instead of rebuilding the monorepo in customer projects?

packages/alchemy/src/OperatingSystem.ts
```diff
@@ -0,0 +1,763 @@
...
+    // Durable Object classes are reached via `ctx.exports`
+    // (`enable_ctx_exports`); the bindings exist to declare the classes and
+    // their SQLite migrations, derived from wrangler.jsonc's history.
+    for (const className of backendWrangler.durableObjects) {
+      backendEnv[className] = Cloudflare.DurableObject(className);
+    }
+    if (props.admins !== undefined) backendEnv.ADMINS = props.admins;
+    const auth = props.auth;
+    if (auth?.gatekeepers?.length) {
+      backendEnv.AUTH_GATEKEEPERS = auth.gatekeepers.join(",");
```
ndisidore: Can we make this typed against the current gatekeeper list? Or at least validated?

I'm worried Aatypo here can leave the whole installation without a login method: the list is non-empty, password auth gets disabled, but no usable auth gatekeeper appears

ndisidore (changes requested):
I would love if CI could produce and publish whatever we call the npm package containing:
- Prebuilt Worker module bundles.
- Prebuilt frontend variants for password/OAuth and Access.
- Static asset manifests and blobs.
- The complete deployment manifest.
- Full DO migration/lifecycle declarations.
- Checksums and source commit metadata.
- Supported configuration schema.
- 

It seems like composable primitives would work well with a convenience facade, i.e.

```ts
const data = yield* OperatingSystemData("Data", {
  protect: true,
});

const github = yield* BuiltinGatekeeper("GitHub", {
  release: CloudflareOs.release.gatekeepers.github,
  credentials: { clientId, clientSecret },
});

const os = yield* OperatingSystem("OS", {
  release: CloudflareOs.release,
  data,
  auth: OAuthAuth({ providers: [github.auth] }),
  admins: ["admin"],
  gatekeepers: [
    github,
    BuiltinGatekeeper.Context(),
    BuiltinGatekeeper.Scheduler(),
  ],
});
```

packages/alchemy/src/wrangler.ts
sam-goodwin: If i was doing it my way, i would not have the deployment.jsonc/wrangler.json and just export typescript with the alchemy code in it. I did it this way conservatively because the repository is currently based on wrangler.

PR #33, opened 2026-08-06
refactor: unify the toolchain on Vite+ (vp)

ubugeeei:
Replaces the split oxlint/vitest/vite setup with the Vite+ toolchain and moves tool configuration into `vite.config.ts`.
Toolchain
- Adds `vite-plus` (0.2.7); `vite` becomes a workspace catalog alias for @voidzero-dev/vite-plus-core, so plugins keep resolving `vite` as before.
- Scripts call `vp lint` / `vp test` / `vp build`; pnpm still drives the recursive orchestration, so `pnpm lint`, `pnpm build` and `pnpm test` keep their meaning and CI needs no changes -- `vp` comes from node_modules/.bin.
- `vitest` imports rewritten to `vite-plus/test`, `vitest/config` to `vite-plus`. No formatting changes: the repo is not oxfmt-clean, so `check.fmt` is false and `vp check` runs lint only. Configuration
- `.oxlintrc.json` moves into the `lint` block of a new root `vite.config.ts`, rules and comments intact. `vp lint` reports the same 0 errors / 64 warnings.
- Every `vitest*.config.ts` folds into its package's `vite.config.ts`. The two packages that ran two suites (workshop-backend, gatekeeper-scheduler) now use `test.projects` instead of a second config file and a second `vitest` call. TypeScript 7
- Most packages move to TypeScript 7. It has no JS compiler API -- the npm package exports `version` and nothing else -- so code that drives the API uses a `typescript-5` alias, and workshop-frontend/workshop-backend/ gatekeeper-scheduler stay on 5.9 (capnweb's recursive Stub types exceed tsgo's instantiation depth; capnweb-validate builds a ts.Program).
- tsgo requires explicit `rootDir` under declaration emit and drops `baseUrl`. `baseUrl` is removed from 21 tsconfigs (their `paths` were already relative). The 17 packages that compile sibling-package sources get `noEmit: true` -- nothing consumes their `dist`, and AGENTS.md already describes `pnpm build` as a type check -- and the 5 that only compile their own `src` get `rootDir`. Known gap: `vp test` does not yet work for workshop-backend and gatekeeper-scheduler. Under Vite+ their @cloudflare/vitest-pool-workers suites fall back to Node's vm instead of workerd. Bumping the pool to 0.20.2 does not help. Every other package's tests pass, as do `pnpm build` and `pnpm lint`.


ndisidore:
Thanks for your interest and contribution! We have our own upgrade process in the works and are putting careful thought into introducing new dependencies

PR #34, opened 2026-08-06
Auto-title chats/workspaces without requiring a manual "Quick model"

HarvardChin:
Title generation (generateThreadTitle/generateGadgetTitle) was gated on userMeta.quickModel, which is null until a user finds the "Quick model" toggle on the Providers settings page. AI Gateway mode already sidesteps this by always resolving a hardcoded quick model; direct/BYOK mode had no equivalent fallback, so titles silently stayed "New Chat"/"Untitled Workspace" forever unless a user opted in.

Falls back to the chat's own model when no quick model has been explicitly chosen, mirroring AI Gateway mode's behavior.

### DIFF
--- a/packages/workshop-backend/src/user.ts
+++ b/packages/workshop-backend/src/user.ts
@@ -692,6 +692,12 @@ export class UserDurableObject extends DurableObject<Cloudflare.Env> {
           result.quickModel = quickModel.config;
         }
       }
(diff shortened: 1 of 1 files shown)


kentonv:
Sorry, I'm not sure this is what we want. If the user chose no quick model, perhaps they prefer not to have auto-generated names at all. They can edit the names manually. That said we'll think about this more as we work on polishing the UI. Thanks for the feedback.

PR #37, opened 2026-08-06
Don't send the developer role from the Ollama provider slot

### LINKED ISSUE #36: Ollama-slot models send role: "developer", which most OpenAI-compatible endpoints reject
## Summary

A model added through the **Ollama** provider slot pointing at a self-hosted OpenAI-compatible endpoint fails on the very first message with HTTP 400:

```
{"message":"developer is not one of ['system', 'assistant', 'user', 'tool', 'function'] - 'messages.[0].role'","type":"invalid_request_error","param":"","code":null}
```

The system prompt goes out with `role: "developer"`, which the endpoint rejects.

## Cause

The `ollama` branch of `getModelDirect()` in `packages/workshop-backend/src/ai-models.ts` hardcodes `reasoning: true` and passes no `compat`:

```ts
case "ollama":
  return makeHandle({
    model: {
      id: config.model,
      name: config.model,
      api: "openai-completions",
      provider: "ollama",
      baseUrl: `${stripTrailingSlashes(config.apiUrl ?? "http://localhost:11434")
          .replace(/\/(api|v1)$/, "")}/v1`,
      reasoning: true,
      input: ["text", "image"],
      cost: ZERO_COST,
      ...window,
    },
    ...
```

pi's `openai-completions` API chooses the system-prompt role like this:

```js
const useDeveloperRole = model.reasoning && compat.supportsDeveloperRole;
const role = useDeveloperRole ? "developer" : "system";
```

and `detectCompat()` infers the flag from provider id / hostname:

```js
supportsDeveloperRole: isOpenRouterDeveloperRoleModel || (!isNonStandard && !isOpenRouter)
```

`isNonStandard` covers nvidia, cerebras, xai, together, chutes, deepseek.com, zai, moonshot, opencode, Workers AI, AI Gateway and ant-ling — **it does not include ollama**, and a self-hosted hostname matches none of the `baseUrl.includes(...)` checks either. So an arbitrary OpenAI-compatible gateway is assumed to accept the `developer` role, and the hardcoded `reasoning: true` supplies the other half of the condition.

This matters more than the provider name suggests: the Ollama slot is the only entry in `AddModelModal` that combines a user-supplied base URL with the chat-completions API (`openai` uses `openai-responses`, `google` uses `google-generative-ai`, `cloudflare` is account-scoped), so it is what users reach for with vLLM, LM Studio, llama.cpp or a self-hosted gateway. Any such endpoint that validates message roles strictly is unusable.

The sibling branch in the same file already handles this — `workersAiCompat()` explicitly sets `supportsDeveloperRole: false`.

## Reproduction

1. **Add Model → "Other Ollama…"**
2. **API URL**: an OpenAI-compatible endpoint that validates `role` (in our case an internal gateway serving a DeepSeek model)
3. **Model ID** / **API Token**: whatever that endpoint serves
4. Send any message → 400 as above, on the first request.

Verified independently with `curl` that the endpoint itself is fine: non-streaming, streaming, `reasoning_content` deltas, assistant-message replay and function calling all return 200. Only the `developer` role is rejected.

## Suggested fix

Pin the flag on the ollama branch, mirroring what `workersAiCompat()` already does:

```ts
compat: { supportsDeveloperRole: false },
```

After this change, chat, streaming, reasoning deltas, `executeCode` tool calls and multi-turn tool-result replay all work against the same gateway.

Happy to open a PR if useful.

## Environment

- cloudflare-os `aedcda8`
- macOS 15 (arm64), Node 24.18.0, pnpm 11.17.0
- run via `pnpm run-local`

warmwind:
Fixes #36.

A model added through the **Ollama** provider slot against a self-hosted OpenAI-compatible endpoint fails on the very first message:

```
400 {"message":"developer is not one of ['system', 'assistant', 'user', 'tool', 'function'] - 'messages.[0].role'","type":"invalid_request_error"}
```

pi chooses the system-prompt role with

```js
const useDeveloperRole = model.reasoning && compat.supportsDeveloperRole;
```

and `detectCompat()` infers `supportsDeveloperRole` from the provider id and base URL. `ollama` is not in its `isNonStandard` list, and a self-hosted base URL matches none of its hostname checks, so the endpoint is assumed to accept `role: "developer"`. The branch also hardcodes `reasoning: true`, which supplies the other half of the condition — so the developer role went out on every request.

This slot is the only one in `AddModelModal` that pairs a user-supplied base URL with the chat-completions API (`openai` uses `openai-responses`, `google` uses `google-generative-ai`, `cloudflare` is account-scoped), so it is what vLLM, LM Studio, llama.cpp and self-hosted gateways get added through, not just Ollama itself. Any of them that validates message roles strictly was unusable.

`workersAiCompat()` in the same file already pins this flag; this brings the ollama branch in line.

## Verification

Against an internal OpenAI-compatible gateway serving a DeepSeek model, before the change every chat failed with the 400 above. After it:

- plain chat ✅
- streaming, including `reasoning_content` deltas rendered as thinking ✅
- `executeCode` tool calls ✅
- multi-turn with tool-result replay ✅ (asked it to compute the sum of squares 1..100 in code; it ran `executeCode` and returned 338350)

`pnpm --filter @gadgets/workshop-backend types:check` passes.

I also checked with `curl` that the endpoint itself accepts everything else pi sends — non-streaming, streaming, assistant-message replay with and without `reasoning_content`, and function calling all return 200. The `developer` role was the only rejected part.

### DIFF
--- a/packages/workshop-backend/src/ai-models.ts
+++ b/packages/workshop-backend/src/ai-models.ts
@@ -563,6 +563,11 @@ function getModelDirect(config: AiModelConfig, sessionAffinity?: string): ModelH
           input: ["text", "image"],
           cost: ZERO_COST,
           ...window,
+          // pi detects the `developer` system role from the hostname, so an OpenAI-compatible
+          // gateway on an unrecognized domain is assumed to accept it. Most don't — they reject
+          // the request with "developer is not one of [...] - 'messages.[0].role'". Reasoning is
+          // hardcoded on above, which is the other half of pi's condition, so pin the role here.
+          compat: { supportsDeveloperRole: false },
         },
         ...(config.apiToken === ""
             ? { apiKey: "unused", headers: { Authorization: null } }


warmwind:

kentonv:
Hi @warmwind,

Apologies but, [per our contributing guide](https://github.com/cloudflare/cloudflare-os/blob/main/CONTRIBUTING.md), we prefer not to receive external pull requests. We'd rather receive issue reports, which we can then prompt our own agents to solve. Could you please file an issue instead?

I realize this one may seem trivial, but I think it's not really -- presumably Pi is using the developer role for a reason here, and by accepting this change we might lose somtehing. The ollama provider is not intended to be a general "OpenAI compatible" provider; it is specifically intended for ollama, and in my testing it does in fact work correctly against ollama.

PR #38, opened 2026-08-06
fix: honor run-local port option

kiyo-e:
## Summary

- Honor `pnpm run-local --port <port>` when starting Wrangler.
- Keep the static frontend WebSocket and every gatekeeper-generated URL on that selected port.

## Root cause

`run-local` forwarded `--port`, but `run-dev-server.js` derived Wrangler's port only from
`VITE_BACKEND_HOST`. The frontend and gatekeepers also retained their `localhost:8787` defaults.

## Validation

- `node --check scripts/run-local.mjs`
- `node --check run-dev-server.js`
- `node --test scripts/dev-server-config.test.js`
- `pnpm --filter @gadgets/workshop-frontend types:check`
- `pnpm --filter @gadgets/workshop-frontend exec vite build`
- `pnpm lint:check run-dev-server.js packages/workshop-frontend/src/main.tsx scripts/run-local.mjs`

### DIFF
--- a/packages/workshop-frontend/src/main.tsx
+++ b/packages/workshop-frontend/src/main.tsx
@@ -63,8 +63,10 @@ function getBackendHost(): string {
   if (backendHost) return backendHost;
 
   // When opening the Vite dev server directly (localhost:3000), the backend is at localhost:8787.
-  // Otherwise, the API is on the same host as the frontend.
-  return window.location.hostname === 'localhost' ? 'localhost:8787' : window.location.host;
+  // Otherwise, including run-local, the API is on the same host as the frontend.
+  return window.location.hostname === 'localhost' && window.location.port === '3000'
+    ? 'localhost:8787'
+    : window.location.host;
 }
 
 function startConnection(): RpcStub<PublicApi> {
--- a/packages/workshop-frontend/src/main.tsx
+++ b/packages/workshop-frontend/src/main.tsx
@@ -62,11 +62,9 @@ function getBackendHost(): string {
   const backendHost = import.meta.env.VITE_BACKEND_HOST?.trim();
   if (backendHost) return backendHost;
 
-  // When opening the Vite dev server directly (localhost:3000), the backend is at localhost:8787.
-  // Otherwise, including run-local, the API is on the same host as the frontend.
-  return window.location.hostname === 'localhost' && window.location.port === '3000'
-    ? 'localhost:8787'
-    : window.location.host;
+  // When opening the Vite dev server directly, the backend is at localhost:8787. Otherwise,
+  // including run-local, the API is on the same host as the frontend.
+  return import.meta.env.DEV ? 'localhost:8787' : window.location.host;
 }
 
 function startConnection(): RpcStub<PublicApi> {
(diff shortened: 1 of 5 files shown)


kiyo-e:

snowyukitty:
Useful fix — the gatekeeper `BASE_URL` half in particular, since without it the OAuth callback URLs
stay on 8787 while everything else moves. One thing I think is worth a second look before this
lands, in the `main.tsx` hunk.

`vite.config.ts` sets `server.port: 3000` but not `strictPort`, so Vite takes the next free port
when 3000 is busy — another checkout already running `pnpm dev`, or anything else on 3000. Running
two dev servers with this repo's exact server options:

```
server A bound to 3000
Port 3000 is in use, trying another one...
server B bound to 3001
```

Today that case still works, because the check is only on the hostname: `localhost:3001` is still
`localhost`, so the frontend correctly looks for the backend on 8787. With
`window.location.port === '3000'` added, it instead returns `window.location.host`, and
`startConnection()` opens `ws://localhost:3001/api` — a path the Vite dev server does not proxy
(only `/api/client-errors`, `/blueprint-screenshot` and `/api/site-logo` are), so the app drops into
the `handleBroken` reconnect backoff with no hint why.

`import.meta.env.DEV` is exactly the distinction the comment above the line is describing, and it
does not involve the port at all: it is `true` under the Vite dev server, and `vite build` — which
is what `run-local` runs before serving `dist` through wrangler — inlines it as `false`:

```
$ vite build   # of `export const isDev = import.meta.env.DEV`
const isDev = false;
```

So:

```js
// When opening the Vite dev server directly, the backend is at localhost:8787. Otherwise,
// including run-local, the API is on the same host as the frontend.
return import.meta.env.DEV ? 'localhost:8787' : window.location.host;
```

That keeps the case this PR is fixing and leaves the moved-port case working. (It is not a complete
answer for `host: true` — opening the dev server from another machine on the LAN still guesses
wrong — but that is true of the current line too, so it is not a regression either way.)

Much smaller, in `scripts/run-local.mjs`: `indexOf("--port")` plus `cliArgs[portIndex + 1]` means
`pnpm run-local --port` with the flag last silently drops it (`port` is `undefined`, so the
`port ? …` branches all take the default) while still filtering it out of `passthroughArgs`, and
`--port=8788` is not recognised at all. Given the PR is specifically about honouring that flag, both
are probably worth an explicit error rather than a silent fall back to 8787.

kiyo-e:
@snowyukitty Thanks for the careful review — both points were spot on. I addressed them in e82062b:

- Switched the frontend distinction to `import.meta.env.DEV`, so Vite moving from 3000 to another port still connects to the backend on 8787, while `run-local` uses the same host as its static frontend.
- Added explicit parsing for both `--port 8999` and `--port=8999`, with errors for missing, non-numeric, and out-of-range values, plus unit coverage. I also fixed the related explicit port 80 normalization case.

`pnpm lint` and `pnpm test` pass, and I re-tested `run-local` on an alternate port successfully. Thanks again for catching these.

kentonv:
Hi @kiyo-e,

Issue #10, opened 2026-08-05
GitHub connection cannot recover from stale credentials through reauthorization


martijngonlag:
## Description

We installed the GitHub gatekeeper by following the documented Cloudflare OS setup steps.

After connecting our GitHub account, the repository selector returned `Bad credentials`. We have not determined what originally caused the credential to become invalid, so this report is not claiming that the initial credential failure was necessarily caused by Cloudflare OS.

The reproducible problem was that completing the GitHub authorization flow again with the same account did not repair the connection, and there is no way to "remove" an account (at least that we could find in the admin UI).

## Steps observed

1. Installed and configured the GitHub gatekeeper using the documented process.
2. Connected a GitHub account successfully through OAuth.
3. Opened the GitHub Repository connector.
4. Repository search returned `Bad credentials`.
5. Selected **Use another GitHub account** and authorized the same identity again.
6. OAuth completed successfully, but repository search continued to return `Bad credentials`.

The UI did not provide an apparent way to remove the invalid connected account.

## Expected behavior

Reauthorizing the same GitHub identity should refresh or replace the stale credential, regardless of what originally caused it to become invalid.

## Fix confirmed in our deployment

We made two code changes:

1. Updated `putConnectedAccount()` so reauthorizing the same vendor and identity replaces the existing account capability with the newly authorized one while preserving the account ID.
2. Changed GitHub revocation from the grant-wide `/applications/{client_id}/grant` endpoint to the token-specific `/applications/{client_id}/token` endpoint.

After deploying these changes, we repeated the same authorization flow and repository search worked immediately.

We did not change the OAuth application, callback URL, requested permissions, client credentials, or gatekeeper bindings between the failing and successful attempts.

The original cause of `Bad credentials` remains unknown, but Cloudflare OS’s inability to recover through reauthorization appears reproducible.

martijngonlag:
Aha, I found that I can disconnect a GitHub account on https://cloudflare-os-deployment-url/gatekeepers. It just can't be done via the "add resource flow" (screenshot). 

<img width="789" height="466" alt="Image" src="https://github.com/user-attachments/assets/711562c1-23e2-4ea7-a8ce-a25141723a8c" />

Would it make sense to allow users to disable the account through this flow as well and/or expose this under `/admin`?

snowyukitty:
Chasing this from the code side: one contributing cause sits in the GitHub gatekeeper and is a bug
on its own, separate from the account-replacement half of your fix.

[...]

PR #39, opened 2026-08-06
Revoke only this account's GitHub token, not the whole app grant

snowyukitty:
`UserAccount.revoke()` in the GitHub gatekeeper calls `DELETE /applications/{client_id}/grant`.
That endpoint deletes the app *authorization*: GitHub documents that it revokes **every** OAuth
token the user holds for that app, not just the one in the request body. Those sibling tokens are
real — GitHub issues a new token on each authorization, up to ten concurrent per
user/application/scope.

`putConnectedAccount()` depends on the opposite. When a user adds a second account and the provider
returns the identity they are already connected as — the case its own comment calls out — it keeps
the existing record and revokes "the newly-created duplicate grant":

https://github.com/cloudflare/cloudflare-os/blob/aedcda8/packages/workshop-backend/src/user.ts#L1586-L1595

Under a grant-wide revoke that call also destroys the **existing** account's still-valid token and
drops the app from the user's GitHub authorizations, so trying to add a duplicate silently breaks a
working connection.

`DELETE /applications/{client_id}/token` is the token-scoped counterpart. It takes the identical
basic auth and `{ access_token }` body, so the change is the path plus a rename for accuracy.

This is also what the rest of the codebase already does. `github.ts` avoids the revoke endpoint
entirely on the ephemeral sign-in path for exactly this reason ("it could invalidate the user's
other grants for this OAuth app"), and among the other connectors Linear (`/oauth/revoke`), Google
(`oauth2.googleapis.com/revoke`) and Slack (`auth.revoke`) are token-scoped, while Cloudflare,
Spotify and Confluence only drop local state. `gatekeeper-github` is the only one that deauthorizes
the app.

Related to #10, but deliberately **not** claimed as fixing it: that reporter also needed
`putConnectedAccount()` to replace the capability in place, which is a separate change. This PR is
just the endpoint, which breaks a working connection on its own.

One judgement call worth flagging: `disconnectAccount()` also goes through `revoke()`, and there a
full deauthorization is arguably what the user wants. `user.ts` describes `revoke()` as the
gatekeeper's cleanup hook rather than a deauthorization, and no other connector deauthorizes, so
this PR keeps them consistent — happy to add a separate call for disconnect instead if you would
rather keep that behaviour.

Verified: `pnpm lint:check` and `pnpm --filter @gadgets/github-gatekeeper types:check` pass. Not
exercised against a live GitHub OAuth app.

### DIFF
--- a/packages/gatekeeper-github/src/github-api.ts
+++ b/packages/gatekeeper-github/src/github-api.ts
@@ -340,14 +340,18 @@ export async function exchangeAuthCode(
   };
 }
 
-export async function revokeOAuthGrant(
+// Revokes one access token. Deliberately the token-scoped endpoint: deleting the *grant*
+// (`/applications/{client_id}/grant`) revokes every token the user holds for this OAuth app, so
+// revoking one account would also invalidate the others sharing it -- including the fresh grant a
+// re-authorization just issued.
+export async function revokeOAuthToken(
   accessToken: string,
   clientId: string,
   clientSecret: string,
 ): Promise<void> {
   await request<void>(
     "DELETE",
-    `/applications/${encodeURIComponent(clientId)}/grant`,
+    `/applications/${encodeURIComponent(clientId)}/token`,
     {
       auth: "basic",
       basicAuth: {
--- a/packages/gatekeeper-github/src/github.ts
+++ b/packages/gatekeeper-github/src/github.ts
@@ -21,7 +21,7 @@ import {
   GitHubApi,
   GitHubApiError,
   exchangeAuthCode,
-  revokeOAuthGrant,
+  revokeOAuthToken,
   type ConditionalRequestResult,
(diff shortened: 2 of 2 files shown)


snowyukitty:

kentonv:
Hi @snowyukitty,

Thanks, but I'm not sure if this is the right fix. The intent of revoke() is to revoke the refresh token, not the access token. We will need to take a closer look at the issue and decide what to do here.

Per the [contributing guidelines](https://github.com/cloudflare/cloudflare-os/blob/main/CONTRIBUTING.md), we prefer not to receive external pull requests -- we'd rather prompt our own agents to write the code. No offense, it's just easier for us that way.

To that end, please raise an issue for any bugs (if #10 doesn't already cover it), or a discussion for any feature requests. Thank for trying to help, and sorry if this policy comes as a disappointment.

Issue #27, opened 2026-08-06
webFetch's 30s timeout does not cover the response body read


snowyukitty:
## Summary

`webFetch` declares a 30s timeout, but the timer is cancelled as soon as response *headers* arrive,
so it never bounds the body read. A server that answers immediately and then trickles the body keeps
the agent turn pending indefinitely, and the user's Stop cannot end it either, because the abort
signal the agent framework hands the tool is not passed on.

## Why the timer is dead by the time it matters

[`web-fetch.ts#L271-L295`](https://github.com/cloudflare/cloudflare-os/blob/e1ab8fbd4f609aff7ede9d490bafe1bcf9b2a682/packages/workshop-backend/src/web-fetch.ts#L271-L295)
arms the timer, `await fetch(...)` resolves once headers are in, and the `finally` clears it. The
body is only read afterwards, at
[`#L316`](https://github.com/cloudflare/cloudflare-os/blob/e1ab8fbd4f609aff7ede9d490bafe1bcf9b2a682/packages/workshop-backend/src/web-fetch.ts#L316),
by [`readBodyCapped`](https://github.com/cloudflare/cloudflare-os/blob/e1ab8fbd4f609aff7ede9d490bafe1bcf9b2a682/packages/workshop-backend/src/web-fetch.ts#L85-L137),
which is a bare `while (true) { await reader.read() }` with no signal and no deadline. Nothing
between the two re-arms anything, and `AbortController` cannot fire after `clearTimeout`.

The 1 MiB cap does not help: it bounds total bytes, not time, and a server that stalls below the cap
never reaches it.

## Reproduced on workerd

A server that flushes headers and then writes one byte every 2s, against four variants of the same
code with the timeout shortened to 5s:

| variant | result |
| --- | --- |
| `clearTimeout` in `finally` after `fetch` — **what the code does today** | still pending when the client gave up at **25s** |
| timer cleared only after the body read | `AbortError` after **5057ms** |
| `signal: AbortSignal.timeout(5000)`, no manual timer | `TimeoutError: The operation was aborted due to timeout` after **5046ms** |
| the patch below | `Error: Fetch timed out after 5000ms` after **5039ms** |

So the abort *does* reach an in-flight `reader.read()` in workerd — the mechanism works, it is just
switched off before the body is touched. Note the middle two rows: simply keeping the timer alive
stops the hang but surfaces a raw abort, because the `catch` that produces the friendly message
wraps only the `fetch` call.

[...]

PR #40, opened 2026-08-06
Keep webFetch's timeout armed until the response body has been read

snowyukitty:
Refs #27 — this covers the timeout half only, so please keep that issue open for the rest (see the
list at the bottom).

`fetch()` resolves as soon as the response headers arrive, and the `finally` cleared
`FETCH_TIMEOUT_MS` right there — before `readBodyCapped()` ran. The declared 30s bound therefore
only covered connect and time-to-headers, and the body read had no deadline at all.
`readBodyCapped` is a bare `while (true) { await reader.read() }` with no signal, so a server that
answers promptly and then stalls holds the agent turn open with nothing to stop it.

The 1 MiB cap does not substitute for a deadline: it bounds total bytes, so a server trickling
below the cap never trips it.

**Measured on workerd**, against a server that flushes headers then writes one byte every 2s, with
the timeout shortened to 5s:

| variant | result |
| --- | --- |
| before this PR | still pending when the client gave up at **25s** |
| just keeping the timer armed | `AbortError` after **5057ms** |
| this PR | `Error: Fetch timed out after 5000ms` after **5039ms** |

The middle row is why the diff is not a one-line move: the abort does reach an in-flight
`reader.read()`, but the `catch` that produces the friendly message wraps only the `fetch` call, so
a timeout during the body read would otherwise surface as a bare `AbortError`. The change clears
the timer on each exit path and maps an abort during the body read onto the same error.

Two things this deliberately leaves alone, both because they change the function's signature or its
caller, and both noted in #27:

- The `AbortSignal` `pi-agent-core` passes to a tool as `execute`'s third argument is still dropped
  by the `webFetch` tool, so Stop still cannot cancel a slow fetch.
- `convertToMarkdown` runs after the body read, so it stays outside the deadline.

Verified: `pnpm lint:check` and `pnpm --filter @gadgets/workshop-backend types:check` pass.

### DIFF
--- a/packages/workshop-backend/src/web-fetch.ts
+++ b/packages/workshop-backend/src/web-fetch.ts
@@ -283,6 +283,7 @@ export async function webFetch(
       signal: abortController.signal,
     });
   } catch (err) {
+    clearTimeout(timeoutId);
     if (
       err instanceof Error &&
       (err.name === "AbortError" || /abort/i.test(err.message))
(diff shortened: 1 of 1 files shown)


snowyukitty:
recheck

snowyukitty:
Closing this in deference to CONTRIBUTING.md and your note on #39.

The finding and the workerd measurements are in #27, which stands on its own — same diff, the
four-variant timing table, and the two related findings this PR deliberately left out. No review
requested here; issues only from here on.

PR #46, opened 2026-08-06
chore: align Vite, Vitest, and Wrangler tooling

ndisidore:
Updates a few of our dev dependencies and reduces version sprawl across packages/manually maintained types 

- Align Wrangler at 4.119.0 and @cloudflare/vitest-pool-workers at 0.18.8 across every consumer; upgrade Vitest users to 4.1.10 and the relevant Workers types to 5.x
- Pin Vite 7.3.6 monorepo-wide with the compatible React plugin (Vite 8/Oxc leaves Stage-3 decorators unlowered; workers-sdk#12626)
- Remove the obsolete CodeMirror overrides, update and deduplicate its packages, and remove the now-unused js-yaml override
- Add a shared `types:generate` command that regenerates and checks Wrangler-owned Worker declarations

All tested and working locally
Screenshot from 2026-08-05 16-50-00


scripts/generate-worker-types.mjs
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env node
+/**
+ * Regenerate worker-configuration.d.ts for every package that has wrangler.jsonc.
```
Maximo-Guk, line 3: Maybe we could regenerate these/manage these with turbo or something in the future ( including our build commands too )

Maximo-Guk (approved):

PR #49, opened 2026-08-06
Fix GitHub issue search repository scope

ndisidore:
GitHub issue search was relying on an appended `repo:` qualifier, but caller-controlled search syntax could escape that scope.

This keeps the repository boundary local by quoting search inputs, validating results before caching or returning them, and retaining the source URL for cache revalidation. It also logs blocked scope mismatches and avoids caching issue bodies.

Added focused regression tests for injection attempts, case handling, and malformed result URLs.

Tested with the full test suite, type checks, and lint.


ndisidore:
Temporarily closing to re-trigger stuck CI checks.

PR #52, opened 2026-08-06
fix: expand oauth scopes for observer verification if needed

maxwellpeterson:
Ensures that the OAuth scopes requested during observer verification cover the resource types that need to be verified.

Fixes situations like these:
1. A user receives a shared Gadget that has observed data from a Google service such as Sheets or Docs
2. The user reaches the observer verification page without an existing Google authentication grant
3. The authentication flow requests only the Gmail scope
4. Observer verification attempts to check the Sheets, Docs, or other relevant Google resource using that insufficient grant
5. The user is rejected as having no access, even though their Google account does have access to the observed resource

### DIFF
--- a/packages/workshop-frontend/src/ObserverConfigModal.tsx
+++ b/packages/workshop-frontend/src/ObserverConfigModal.tsx
@@ -43,13 +43,10 @@ function accountLabel(account: AccountInfo | undefined, accountId: number): stri
   return account?.description.uniqueName || account?.description.displayName || `Account ${accountId}`
 }
 
-/**
- * Returns the grantable resource type needed to verify one observer binding.
- *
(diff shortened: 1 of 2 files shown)


packages/workshop-frontend/src/ObserverConfigModal.tsx
```diff
@@ -36,6 +43,37 @@ function accountLabel(account: AccountInfo | undefined, accountId: number): stri
   return account?.description.uniqueName || account?.description.displayName || `Account ${accountId}`
 }
 
+/**
```
ndisidore: super nit: (i've gotten called out here) inconsistent comment style with the rest of this file

ndisidore (approved):

Maximo-Guk (approved):

PR #53, opened 2026-08-06
Let admins choose which built-in AI models are offered to users

xsloba:
Adds a disabledAiModels set to the deployment admin config. Built-in AI Gateway models can now be disabled individually from a new "AI models" tab in /admin, so a deployment can offer a provider without offering every one of its models (resolves the SUGGESTED_MODELS TODO about needing an admin option to disable specific models).

A disabled model is hidden from user model lists, refused as a chat or preferred model, and cannot be re-added as a custom model (which would otherwise route through the platform gateway and undo the disable). User-added custom models are otherwise unaffected, as is the internal quick model. Follows the disabledGatekeepers pattern: everything stays enabled by default and the admin UI opts models out.

### DIFF
--- a/packages/workshop-backend/__tests__/admin-config.test.ts
+++ b/packages/workshop-backend/__tests__/admin-config.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { DEFAULT_ADMIN_CONFIG, defaultOutputFormatId, parseAdminConfig, reorderFormats, resolveFormatOutput, sanitizeOutputOverrides, serializeAdminConfig } from "../src/admin-config.js";
+import { DEFAULT_ADMIN_CONFIG, defaultOutputFormatId, isAiModelDisabled, parseAdminConfig, reorderFormats, resolveFormatOutput, sanitizeOutputOverrides, serializeAdminConfig } from "../src/admin-config.js";
 
 describe("parseAdminConfig", () => {
   it("backfills fields missing from a config persisted before they existed", () => {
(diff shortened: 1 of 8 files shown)


ndisidore:
Your head is absolutely in the right place here! I have local work pending that does nearly exactly this.

One of the tricky parts of this work is what happens to existing workspaces using a model that become disabled. Especially on workspaces with background tasks that may possibly have ai models attached: I feel like as an admin, I would expect "off" to be "no usage", but breakage/jarring ux also seems less then ideal
In my own work, I have the admin required to choose a fallback model.
Though it is fair to say that with gatekeeper disablement, breakage is expected.

Let me see if I can get what I have cleaned up and pushed and we can compare notes

xsloba:
Perfect! Also I apologize for jumping the gun here. I just wanted to contribute, as this is one of the last remaining items before our large rollout, and company policy restricts the use of specific models company-wide.

Open questions I hit while writing it:

- Should an existing chat pinned to a since-disabled model fail hard (current behavior in my branch: refuse with "disabled by administrator") or silently fall back to the deployment default?
- Is per-model curation the right altitude, or would you rather see an allowlist env var (CF_AI_GATEWAY_MODELS) so it stays deployment-controlled like auth config? I went with /admin because model curation feels like runtime policy (same tier as connector availability), not a trust boundary.

kentonv:
Hi @xsloba,

Apologies but, [per our contributing guide](https://github.com/cloudflare/cloudflare-os/blob/main/CONTRIBUTING.md), we prefer not to receive external pull requests. We prefer instead to receive feature requests via discussions, which we can then direct our own agents to implement (if we think it makes sense -- though in this case I think it does). Could you please open a discussion instead?

xsloba:
[Opened](https://github.com/cloudflare/cloudflare-os/discussions/50#discussion-10565155). Thank you!

PR #56, opened 2026-08-07
Backend observability and hygiene for Durable Object resets

ndisidore:
## Overview

Backend-only: restore a lost observability signal, fix a day-one debug constant, and switch on the instrumentation that upcoming reset-resilience work relies on. No request-handling behavior changes.

image

## What the data showed

From internal logs (identifiers scrubbed, figures rounded):

- Tens of thousands of "lost connection to workspace DO" aborts on `GET /api` — then abruptly zero after a deploy swapped `ctx.abort(error)` for a bare `webSocket.close()` (`b2a215ec`). The aborts almost certainly continue; we just stopped logging them.
- Thousands of storage-timeout DO resets, mostly from a single workspace hammering `startGatekeeperSession` — pathological gatekeeper traffic, not organic editing. What's slow *inside* such a DO is unknown; that's the tracing gap.
- `MIN_SNAPSHOT_THRESHOLD` shipped as `256; //65536` against a comment saying 64k (`b10f1ba5`), so small workspaces rebuild their full Yjs snapshot on nearly every edit burst.

## Commits

- **Restore visibility of API session aborts** — `abortSession` closes the socket without failing any invocation, so nothing logged it; one `warn` with the discarded reason restores the signal.
- **Restore intended 64 KiB snapshot threshold** — one line; hygiene, not claimed as the dominant reset cause.
- **Enable Workers Traces** — auto-instruments fetch/binding/DO invocations (incl. the loopback hops behind the concentrated resets); 0.5 sampling, free during beta.
- **Custom spans + heavy-op duration logs** — spans on the two interior ops auto-instrumentation can't see (`agent.run`, `code.snapshot.rebuild`); 100%-sampled duration logs on rare heavy ops that trace sampling could miss.

## Verification

Unit suite + `tsc` green; `wrangler deploy --dry-run` validates the config. Post-deploy: `session.abort` volume should match the pre-regression rate.

### DIFF
--- a/packages/workshop-backend/wrangler.jsonc
+++ b/packages/workshop-backend/wrangler.jsonc
@@ -94,7 +94,10 @@
   "observability": {
     "enabled": true,
     "head_sampling_rate": 1,
-    "logs": { "invocation_logs": false }
+    "logs": { "invocation_logs": false },
+    // Traces are free during the beta; spans bill against the Logs quota from 2026-10-01,
+    // so revisit the sampling rate before then.
+    "traces": { "enabled": true, "head_sampling_rate": 0.5 }
   }
 
 }
--- a/packages/backend-utils/__tests__/tracing.test.ts
+++ b/packages/backend-utils/__tests__/tracing.test.ts
@@ -1,0 +1,49 @@
+// Runs against the real workerd tracing API (vitest-pool-workers), pinning the platform
+// contract `traced` depends on: enterSpan ends its span when the returned promise settles, not
+// when the synchronous callback returns (docs: workers/observability/traces/custom-spans/).
+//
+// Caveats:
+// - Invocations are only traced when a tail consumer is attached, so vitest.config.ts wires a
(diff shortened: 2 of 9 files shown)


packages/workshop-backend/wrangler.jsonc
```diff
@@ -94,7 +94,10 @@
   "observability": {
     "enabled": true,
     "head_sampling_rate": 1,
-    "logs": { "invocation_logs": false }
+    "logs": { "invocation_logs": false },
+    // Traces are free during the beta; spans bill against the Logs quota from 2026-10-01,
+    // so revisit the sampling rate before then.
+    "traces": { "enabled": true, "head_sampling_rate": 0.5 }
```
Maximo-Guk, line 100: Are we going to adjust this to head_sampling_rate: 1 on our instance?

ndisidore: Yep we'll override for our deploy :+1: 
Didn't want to throw too much out at once here

Maximo-Guk (approved):

PR #57, opened 2026-08-07
Fix gatekeeper app navigation after OAuth

ShibinCo:
## What

Refresh the cached gatekeeper app list when the connected-account subscription adds an account that provides a management UI.

OAuth connections arrive through this subscription after the connect popup opens. Without the refresh, the account appears under Gatekeepers but its app remains absent from navigation until a full page reload. The ambient provisioning path already performs the same cache invalidation explicitly.

## Verification

- `pnpm --filter @gadgets/workshop-frontend types:check`
- `pnpm --filter @gadgets/workshop-frontend test` (24 files, 115 tests)
- `pnpm lint:check`

### DIFF
--- a/packages/workshop-frontend/src/routes/gatekeepers.tsx
+++ b/packages/workshop-frontend/src/routes/gatekeepers.tsx
@@ -540,6 +540,8 @@ function ConnectorsPage() {
           credentialsValid,
         })
         setAccounts(Array.from(accountMap.values()))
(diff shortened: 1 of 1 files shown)


kentonv:
Hi @ShibinCo,

(I realize that the guide says that trivial PRs are allowed. However, I honestly couldn't tell at a glance what this is doing, and I also noticed you did not accept the CLA. So, I think it's best to proceed with an issue report.)

PR #58, opened 2026-08-07
Classify RPC errors and recover quietly from Durable Object resets

ndisidore:
Stacked on #56.

Addresses this class of error
Screenshot from 2026-08-07 17-08-52
Screenshot from 2026-08-07 17-09-00

Routine DO resets (storage timeouts, overload aborts i.e. the stuff we saw in the logs) were surfacing in the frontend as scary terminal errors. The reject frames actually arrive with structured flags (`durableObjectReset`, `overloaded`, etc., thanks to `enhanced_error_serialization`), but nothing ever read them: every call site was just `catch(console.error)`.

Two details made it worse than it needed to be:

- A user-DO reset rejects the in-flight RPC but the WebSocket stays healthy, so nothing re-fetched. Model list, sidebar, onboarding check just stayed broken until you reloaded the page.
- The API session cached its user-DO stub for its whole lifetime. A stub is bound to one incarnation of the object and is permanently broken after a reset (see the [DO error handling docs](https://developers.cloudflare.com/durable-objects/best-practices/error-handling/)), so a retry through the cached one could never succeed anyway.

## What changed

- New RPC error classifier. Flags first; message matching only as a fallback for errors that lose them in transit.
- Idempotent reads and subscribes retry once (~1.5s, jittered) after a reset. Never writes. The retry lives at one chokepoint: a `withReadRetries` proxy installed where `useAuth` creates the API stub, keyed by a method-level allowlist — idempotency is a property of the method, not the call site, so every consumer (present and future) gets the policy without per-call-site wrapping.
- The backend keeps **one user-DO stub per session** — e-order (in-order delivery) is guaranteed per stub, so this preserves request ordering (see [the review discussion](https://github.com/cloudflare/cloudflare-os/pull/58#discussion_r3739300221)). A Proxy around the stub observes rejections: flagged errors (`durableObjectReset`/`retryable`) drop the cached stub and propagate — the call may have executed, so the frontend owns recovery — while a flagless dead-capability rejection proves the call never reached the object and is transparently re-issued exactly once on a fresh stub, which is safe even for writes. Net effect: an idle reset costs the user nothing at all.
- Transient failures (ones a retry or reconnect will cure) log at debug instead of toasting, and `logRpcFailure` reports do-resets on action paths (chat send/new, workspace create) to the client-errors endpoint so they stay visible in telemetry. State fallbacks unchanged.
- Failed chat sends show an inline "may not have been sent — check the thread" hint instead of the old toast. Deliberately not auto-retried: if the reset lands after the write commits, a retry double-sends.
- Auth errors get stable codes (same pattern as the workspace-open ones, both now built by one `codedErrorFamily` helper). Canary tests pin the capnweb message strings we match against and round-trip the flags through capnweb's real serializer; the workerd dead-capability message lives once in `workshop-shared`, referenced by both the backend matcher and the frontend classifier, so the canary guards both.

## Intentionally Deferred
There remains a gap in this work around "Established subscriptions remain dead after a User DO reset." - subscriptions present a slightly more challenging workflow so that work is intentionally left out here.

## Testing

- Integration tests (`abortAllDurableObjects()`, the non-graceful teardown): the same session recovers transparently on the first call after a reset (this also canaries the dead-capability message — if workerd's string drifts, the never-sent retry stops firing and the test fails loudly), and the stub cache re-arms instead of churning or staying poisoned.
- Unit tests pin the chokepoint contract (listed reads retry once, writes never) alongside the existing classifier/retry/canary suites. 279 backend + 142 frontend tests green.
- Manual fault-injection re-run (chrome-devtools against a local dev stack, temporary `debugAbort()` injector): a single idle reset followed by page navigation, and a 20-abort storm at 150ms intervals across five SPA navigations, both completed with every page functional and zero application console output. Server-side spans confirmed exactly the injected aborts as exceptions, the only casualties being long-lived subscriptions (the deferred gap above); every user-facing read and write in the window succeeded, including writes issued after resets. One in-flight write during a burst failed loudly and succeeded on manual retry — the intended contract.

### DIFF
--- a/packages/workshop-backend/src/server.ts
+++ b/packages/workshop-backend/src/server.ts
@@ -1,7 +1,7 @@
 import { RpcStub, RpcTarget, newWorkersRpcResponse } from "capnweb";
 import { validateRpc } from "capnweb-validate";
 import type { JWTPayload } from "jose";
-import { PublicApi, AuthenticatedApi, Overseer, GadgetMetadataWithTimestamps, AiChatAuthorInfo, AiModelConfig, AiGatewayInfo, AiModelProvider, ConnectedAccountsSubscriber, ConnectedAccountsFilter, GatekeeperVendorFilter, ObserverConfigCallback, BlueprintLibrarySummary, BlueprintPublicInfo, BlueprintUserSummary, BlueprintBindingAssignment, AgentSpawnerConfig, WorkpieceId, BLUEPRINT_SCREENSHOT_PATH_PREFIX, BLUEPRINT_SCREENSHOT_R2_PREFIX, blueprintScreenshotUrl, ServerConfig, CloudflareUsageInfo, CloudflareAccountOption, LoginAttempt, GatekeeperAppInfo, AdminApi, GatekeeperVendorInfo, OutputFormatOffer, ListOutputsResult, createOpenGadgetError, getOpenGadgetErrorCode, OPEN_GADGET_ERROR_CODES } from '@gadgets/workshop-shared/api';
+import { PublicApi, AuthenticatedApi, Overseer, GadgetMetadataWithTimestamps, AiChatAuthorInfo, AiModelConfig, AiGatewayInfo, AiModelProvider, ConnectedAccountsSubscriber, ConnectedAccountsFilter, GatekeeperVendorFilter, ObserverConfigCallback, BlueprintLibrarySummary, BlueprintPublicInfo, BlueprintUserSummary, BlueprintBindingAssignment, AgentSpawnerConfig, WorkpieceId, BLUEPRINT_SCREENSHOT_PATH_PREFIX, BLUEPRINT_SCREENSHOT_R2_PREFIX, blueprintScreenshotUrl, ServerConfig, CloudflareUsageInfo, CloudflareAccountOption, LoginAttempt, GatekeeperAppInfo, AdminApi, GatekeeperVendorInfo, OutputFormatOffer, ListOutputsResult, createOpenGadgetError, getOpenGadgetErrorCode, OPEN_GADGET_ERROR_CODES, AUTH_ERROR_CODES, createAuthError } from '@gadgets/workshop-shared/api';
 import type { UiFeatureFlags } from "@gadgets/workshop-shared/feature-flags";
 import { getServerConfig } from "./deployment-config.js";
 import { isPasswordAuthEnabled, getAuthGatekeeperAllowlist } from "./auth/config.js";
@@ -671,7 +671,7 @@ class PublicApiImpl extends RpcTarget implements PublicApi {
   async authenticate(token: string): Promise<AuthenticatedApi> {
     let split = token.split(':');
     if (split.length !== 2) {
-      throw new Error("Invalid session token.");
+      throw createAuthError(AUTH_ERROR_CODES.invalidSessionToken);
     }
 
     let userId = this.users.idFromName(split[0]);
@@ -687,7 +687,7 @@ class PublicApiImpl extends RpcTarget implements PublicApi {
 
   async authenticateFromCfAccess(): Promise<AuthenticatedApi> {
     if (!this.accessPayload) {
-      throw new Error("Not authenticated with Access.");
+      throw createAuthError(AUTH_ERROR_CODES.notAuthenticatedWithAccess);
     }
 
     let email = this.accessPayload.email as string;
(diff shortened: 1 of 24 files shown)


packages/workshop-backend/src/server.ts
```diff
@@ -74,7 +74,7 @@ type Env = Cloudflare.Env & {
 @validateRpc()
 class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
   constructor(private ctx: ExecutionContext, private env: Env,
-      private user: DurableObjectStub<UserDurableObject>,
```
ndisidore, line 77: This is likely a controversial decision.
Its driven by https://developers.cloudflare.com/durable-objects/best-practices/error-handling/ specifically the block

> Many exceptions leave the [DurableObjectStub](https://developers.cloudflare.com/durable-objects/api/stub) in a "broken" state, such that all attempts to send additional requests will just fail immediately with the original exception. To avoid this, you should avoid reusing a DurableObjectStub after it throws an exception. You should instead create a new one for any subsequent requests.

When the the user DO resets e.g. storage timeout, overloaded abort (which is exactly what we saw in the logs) that stub is permanently poisoned. Even if we retry it will fail.

This is not super obvious because the premise does hold for the workspace path: Overseer stubs get re-resolved through the namespace on each open, so a retried openGadget genuinely reaches the restarted object. The UserDO path is the only one where a stub is cached across calls.

This should be cheap and safe: namespace.get(id) is not a network call. Stub creation is local and lazy.

Why not just force a re-load? a reload doesn't avoid the retry; _it is the retry,_ multiplied by everything else and makes blast radius wildly disproportionate. It give a worse UX as well as the entire page resets (as opposed to trying to recover silently where possible)

packages/workshop-frontend/src/rpcErrors.ts
```diff
@@ -0,0 +1,110 @@
...
+
+// Retries an idempotent call once after a backend-side transient failure: a DO reset (the object
+// restarts on its next request, reached via a fresh stub) or a `retryable`-flagged invocation
+// failure. Flags only survive on errors that round-tripped from the backend, so a flagged error
+// proves the socket was healthy and a retry-in-place can succeed. Local transport errors carry no
+// flags and are deliberately not retried: the connection manager owns that recovery, and a retry
+// through the closure-captured dead stub could never succeed anyway. Deliberately retries even
+// when `overloaded` is set alongside the reset — the reset destroyed the queue that was
+// overloaded, and one jittered attempt is not a retry loop. Never use for writes.
+export async function withDoResetRetry<T>(fn: () => Promise<T>, delayMs = 1500): Promise<T> {
```
ndisidore, line 107: This comment goes more in depth here, but I want get ahead of any guttural reactions: Retry is confined to idempotent reads, bounded to exactly 1 attempt.

packages/workshop-backend/src/server.ts
Kieran-Hulsman: i don't think it should be a problem, but do different stubs mean we lose request ordering to the userDO? i couldn't find any instances were that would be a big problem though, so it seems like a worthwhile tradeoff.

packages/workshop-frontend/src/ChatInterface.tsx
```diff
@@ -3050,6 +3064,11 @@ export const ChatInput = ({
           </div>
         )}
         {draftUpdateBanner}
+        {sendHiccup && (
+          <div className="px-4 pt-2 text-xs text-kumo-warning">
+            Connection hiccup — your message may not have been sent. Check the thread, then try again.
```
Kieran-Hulsman, line 3066: @ndisidore what happens in this case?

1. The user tries to send a message to chat
2. The workspace reads the user’s profile and model configuration through its cached User DO connection.
3. The User DO has reset, so that connection fails.
4. The message is not written to the Overseer DO.

it looks like the message's body & attachments aren't thrown away, but retrying won't get a fresh User DO stub yet. so when the user tries to resend, will it work?

Kieran-Hulsman (approved):
i'm a bit out of my depth here, but i spent an hour thinking through this with an agent, and everything seems sensible to me!

packages/workshop-backend/src/server.ts
ndisidore: haha when writing this I told claude "have some pre-canned responses ready for the inevitable push back"
and it had one for e-order! the concern is valid. but practically nothing in our code relies on cross-call ordering through the user stub

preserving ordering while recovering poisoned stubs would require caching and centrally invalidating the stub after native RPC failures across every UserDO operation, adding substantial kernel complexity

PR #68, opened 2026-08-07
fix: handle malformed percent-escapes in Gmail and Calendar URLs

### LINKED ISSUE #55: Gmail URLs containing a bare % throw an unhandled URIError when binding a resource

Ashutosh0x:
Fixes #55

### Problem
Gmail URLs containing a bare \%%\ (e.g. \50%%off\, \Q1 100%%\) throw an unhandled \URIError\ from bare \decodeURIComponent()\ calls in the \#search/\ and \#label/\ branches. The rest of \getGatekeeperClassFor\ provides friendly errors for bad input, but this path crashes with a raw stack trace.

### Fix
Add a \safeDecodeURI\ helper that catches \URIError\ and falls back to the raw string. Applied to the Gmail search, label, and Calendar URL decode sites.

8 insertions, 3 deletions.

cc @jonesphillip @kentonv

### DIFF
--- a/packages/gatekeeper-google/src/google.ts
+++ b/packages/gatekeeper-google/src/google.ts
@@ -865,8 +865,13 @@ export class GatekeeperUserImpl extends WorkerEntrypoint<Env, GatekeeperUserImpl
       };
     }
 
+    // Safe decode that falls back to the raw string for malformed percent-escapes (e.g. "50%off").
+    const safeDecodeURI = (value: string): string => {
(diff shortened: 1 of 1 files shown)


Ashutosh0x:

kentonv:
Hi @Ashutosh0x,

PR #70, opened 2026-08-07
fix: run pnpm through a shell in execFileSync/spawnSync on Windows

unknownhad:
## What

On Windows, `pnpm` is a `.CMD` shim, not a real executable. Node's `execFileSync` does not apply `PATHEXT`, so `execFileSync("pnpm", ...)` fails with `spawnSync pnpm ENOENT`. (`git`/`node` calls are unaffected because they resolve to real `.exe` files.)

This breaks `pnpm run-local` on Windows: the first `pnpm install` succeeds, but every later step that shells out to `pnpm` aborts.

## Fix

Pass `shell: process.platform === "win32"` at the four `pnpm` call sites so the shim resolves on Windows. POSIX behavior is unchanged (`shell` stays `false` off-Windows).

- `scripts/run-local.mjs` — `pnpm install` + typed-storage/frontend builds
- `run-dev-server.js` — `pnpm exec wrangler dev`
- `packages/gatekeeper-context/build-app.mjs` — `pnpm exec vite build` (incl. `--watch`)
- `packages/gatekeeper-scheduler/build-app.mjs` — `pnpm exec vite build` (incl. `--watch`)

## Testing

With this patch, `pnpm run-local -- --use-workers-ai-binding` completes install + build and serves at `http://localhost:8787` on Windows 11 (Node 23, pnpm 11.17.0). Verified the app loads, a Workers AI model can be configured, and the agent builds and runs a gadget end to end.

### DIFF
--- a/packages/gatekeeper-context/build-app.mjs
+++ b/packages/gatekeeper-context/build-app.mjs
@@ -15,5 +15,5 @@ console.log(
 execFileSync(
   "pnpm",
   ["exec", "vite", "build", "-c", "vite.config.ts", ...(watch ? ["--watch"] : [])],
-  { cwd: pkgDir, stdio: "inherit" },
+  { cwd: pkgDir, stdio: "inherit", shell: process.platform === "win32" },
 );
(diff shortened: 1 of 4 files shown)


unknownhad:

kentonv:
This looks to be the same thing as #18, but #18 covers more files, so I'm closing this one. Thanks for reporting, though.

PR #72, opened 2026-08-07
fix: reject unsupported URLs before Gmail fallthrough in Google Gatekeeper

### LINKED ISSUE #63: Google Gatekeeper falls through unsupported URLs to Gmail
## Summary

The Google Gatekeeper treats every URL that does not match its Docs, Sheets, Calendar, or BigQuery branches as a Gmail resource. Unsupported hosts and unsupported paths therefore receive a Gmail capability instead of being rejected.

## Evidence

`packages/gatekeeper-google/src/google.ts:832-955` checks the recognized resource branches, then falls through to `// Default: Gmail` and returns `GMAIL_RESOURCE` without checking `parsed.hostname`, protocol, or whether the URL matches the Gmail resource pattern (`https://mail.google.com/*`).

The capability boundary does not independently reject the mismatch. `packages/workshop-backend/src/overseer.ts:7447-7458` passes the caller-supplied `resourceUrl` to `UserDurableObject.getGatekeeperClassFor()` and stores that original URL in `GatekeeperCreationSpec`. `packages/workshop-backend/src/user.ts:1621-1645` trusts the Gatekeeper-returned resource metadata for policy checks and returns it without comparing it to the requested URL.

## Reproduction from the current source

For a connected Google account, request a Gatekeeper with either of these resource URLs:

```text
https://evil.example/resource
https://docs.google.com/document/not-a-document-id
```

`GatekeeperUserImpl.getGatekeeperClassFor()` falls through to the Gmail branch and returns `GmailGatekeeperImpl` plus `GMAIL_RESOURCE`, even though neither URL is a Gmail URL. A direct client, blueprint, or any caller that reaches `OverseerClientInterface.newGatekeeper()` can supply the URL; the core stores the mismatched original URL as the binding's resource URL.

## Impact

This is not a cross-user account escalation: the capability still comes from the caller's own connected Google account. It is a resource-URL validation and audit-integrity bug. A non-Gmail request silently becomes a Gmail mailbox capability, and the returned resource metadata no longer describes the URL the caller requested. That weakens the intended resource-pattern boundary and can make admin/resource policy and audit records misleading.

The agent request path normally pre-validates URLs using `resolveRequestedResource()`, but the capability boundary is also reached by the frontend and blueprint instantiation paths, so the Gatekeeper should reject unsupported URLs itself rather than rely on callers.

## Suggested direction

Reject unsupported Google hosts and paths before the Gmail fallback, or make each resource parser explicitly recognize its canonical host/pattern and return an error otherwise. Add a regression test that calls `getGatekeeperClassFor()` with a non-Google URL and an unsupported Google path and asserts rejection.

I am filing this as an issue rather than a PR because the repository's `CONTRIBUTING.md` asks external contributors to report bugs and avoid design-level patches.

Ashutosh0x:
Fixes #63

### Problem
\getGatekeeperClassFor()\ falls through to the Gmail branch for **any** URL that doesn't match Docs, Sheets, Calendar, or BigQuery — including non-Google URLs like \https://evil.example/resource\. This silently returns a Gmail capability for an unrelated resource URL, weakening the resource-URL boundary and making audit records misleading.

### Fix
Add a hostname check for \mail.google.com\ before the Gmail fallthrough. Unsupported URLs now get a clear error listing the supported resources.

9 insertions, 0 deletions.

cc @jonesphillip @kentonv

### DIFF
--- a/packages/gatekeeper-google/src/google.ts
+++ b/packages/gatekeeper-google/src/google.ts
@@ -927,6 +927,15 @@ export class GatekeeperUserImpl extends WorkerEntrypoint<Env, GatekeeperUserImpl
       };
     }
 
+    // Reject URLs that don't belong to Gmail before falling through to the default.
+    // Without this check, any URL that doesn't match Docs/Sheets/Calendar/BigQuery
+    // silently receives a Gmail capability, which weakens the resource-URL boundary.
+    if (parsed.hostname !== "mail.google.com") {
+      throw new Error(
+        `Unsupported Google resource URL: ${parsed.hostname}${parsed.pathname}. ` +
+        "Supported resources are Google Docs, Sheets, Calendar, BigQuery, and Gmail.");
+    }
+
     // Default: Gmail
     let props: GmailGatekeeperImplProps = {...this.ctx.props};
 


Ashutosh0x:

kentonv:
Hi @Ashutosh0x,

PR #73, opened 2026-08-07
Fix Enter during IME composition sending the message

kote2kote:
## Problem

When typing with an IME (Japanese, Chinese, Korean input methods), pressing Enter to confirm a text conversion is interpreted as a send action. This makes the chat input nearly unusable for CJK users: every conversion confirmation sends a half-written message.

## Fix

Guard the `onKeyDown` handlers with the standard IME check:

```ts
if (e.nativeEvent.isComposing || e.keyCode === 229) return;
```

`keyCode === 229` covers browsers (notably Safari) that fire the confirming keydown without `isComposing` set.

Applied to the three text inputs in `ChatInterface.tsx` that submit on Enter:
- the main chat message input
- the chat list rename input
- the chat title edit input

## Testing

- With a Japanese IME: type text, press Enter to confirm the conversion → message is not sent; press Enter again → message sends.
- Without an IME: Enter sends, Shift+Enter inserts a newline, slash-command picker Enter/Tab/arrow behavior unchanged (the guard returns before those branches only during active composition).
- `tsc --noEmit` passes in `packages/workshop-frontend`.

### DIFF
--- a/packages/workshop-frontend/src/ChatInterface.tsx
+++ b/packages/workshop-frontend/src/ChatInterface.tsx
@@ -3146,6 +3146,10 @@ export const ChatInput = ({
                 }
               }}
               onKeyDown={(e) => {
+                // While an IME composition is active (e.g. Japanese input), Enter confirms
+                // the conversion and must not send the message. keyCode 229 covers browsers
+                // that fire the confirming keydown without isComposing set (e.g. Safari).
+                if (e.nativeEvent.isComposing || e.keyCode === 229) return;
                 if (slashCommandPicker.open && e.key === "Escape") {
                   e.preventDefault();
                   slashCommandPicker.dismiss();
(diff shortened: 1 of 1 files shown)


kote2kote:
After opening this I noticed #31 addresses the same IME composition bug in the chat composer and was submitted first — apologies for the overlap.

For the maintainers, the differences in this PR:

- It includes the `keyCode === 229` fallback for the Safari/WebKit event-ordering case raised in [this comment on #31](https://github.com/cloudflare/cloudflare-os/pull/31#issuecomment-5214240555) ([WebKit bug 165004](https://bugs.webkit.org/show_bug.cgi?id=165004)), where `compositionend` can fire before the confirming Enter keydown so `isComposing` is already false.
- It applies the same guard to two other Enter-submitting inputs with the same bug: the chat list rename input and the chat title edit input.

Happy to defer to #31 and close this one if you'd prefer to land that first — either way, whichever base you pick, the 229 fallback and the two extra inputs are probably worth including. Your call.

kote2kote:

kentonv:
Hi @kote2kote,

kote2kote:
Understood, thanks for the explanation — filed as #81. It includes repro steps, the root cause (three `onKeyDown` handlers in `ChatInterface.tsx`), and the suggested guard incl. the Safari `keyCode === 229` edge case, so hopefully it's easy for your agents to pick up.

PR #82, opened 2026-08-08
Don't treat IME composition keys as Enter

arakawayasuaki:
## Problem

Typing Japanese (or any other IME-composed text) sends the prompt half-written. The Enter that commits a kana-kanji conversion reaches our `onKeyDown` handlers as a plain Enter, so the chat input submits mid-word.

The same bug hits every rename and inline-edit field across the app, and in the ⌘K command palette the arrow keys that pick a conversion candidate move the result selection instead.

There was no `isComposing` guard anywhere in the frontend.

## Fix

Add `isComposingKeyEvent()` in `src/utils/imeComposition.ts` — checks `nativeEvent.isComposing`, plus the legacy `keyCode === 229` that Safari still reports on the commit key — and bail out of the affected handlers while composition is in progress.

Guarded inputs:

- `ChatInterface.tsx` — prompt input, chat-list rename, title edit
- `CommandPalette.tsx` — ⌘K palette (Enter *and* the candidate-selection arrow keys)
- `FileSidebar.tsx` — new-file name, file rename
- `ShareModal.tsx` — add collaborator, create share link, link name
- `GadgetEditor.tsx`, `WorkpiecePicker.tsx`, `GadgetList.tsx`, `SidebarGadgetRow.tsx` — title / rename fields
- `SettingsPage.tsx`, `Connections.tsx`, `ConnectionConfigModal.tsx`, `AdminFormatsPanel.tsx`

Left alone: the Enter/Space handlers on `role="button"` divs in `AdminPage.tsx`, `routes/outputs.tsx`, `routes/gatekeepers.tsx`, `routes/providers.tsx`. Those activate buttons rather than edit text, so no IME is ever composing over them.

## Testing

`tsc --noEmit` and `oxlint` pass with no new findings. Verified manually against a local `pnpm run-local` build: Japanese conversion now commits without submitting.

### DIFF
--- a/packages/workshop-frontend/src/ChatInterface.tsx
+++ b/packages/workshop-frontend/src/ChatInterface.tsx
@@ -115,6 +115,7 @@ import { useActionEntries } from "./useActions";
 import { useAlwaysApproveTag } from "./useAlwaysApproveTag";
 import { useResolveAction } from "./useResolveAction";
 import { safeExternalUrl } from "./utils/safeExternalUrl";
+import { isComposingKeyEvent } from "./utils/imeComposition";
 import { useAuthenticatedApi } from "./AuthContext";
 import { useVendorBranding } from "./useVendorBranding";
 import OutOfCreditsModal from "./components/billing/OutOfCreditsModal";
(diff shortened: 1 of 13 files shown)


maxwellpeterson:

Issue #81, opened 2026-08-07
Enter during IME composition sends the message (Japanese/Chinese/Korean input)


kote2kote:
## Bug

When typing with an IME (Japanese, Chinese, Korean input methods), pressing Enter to confirm a text conversion is interpreted as a send action. Every conversion confirmation sends a half-written message, which makes the chat input nearly unusable for CJK users.

Filing as an issue per the maintainer guidance in #73 (and #31, which reported the same bug as a PR).

## Steps to reproduce

1. Open the chat input with a Japanese IME active (e.g. macOS standard Japanese input).
2. Type some text — the IME shows it as an underlined pre-edit with conversion candidates.
3. Press Enter to confirm the conversion.
4. The message is sent immediately, even though the user was still composing.

Reproduced on macOS with Chrome and Arc (see also the reports from other users in #31). Safari has an additional quirk, noted below.

## Root cause

The `onKeyDown` handlers in `packages/workshop-frontend/src/ChatInterface.tsx` treat any unmodified Enter as submit without checking whether an IME composition is active. Three inputs are affected:

- the main chat message input (the `e.key === "Enter" && !e.shiftKey` branch)
- the chat list rename input
- the chat title edit input

## Suggested fix

Guard each affected `onKeyDown` at the top with the standard IME check:

```ts
if (e.nativeEvent.isComposing || e.keyCode === 229) return;
```

The `keyCode === 229` fallback matters for Safari/WebKit, which can dispatch `compositionend` before the confirming Enter keydown ([WebKit bug 165004](https://bugs.webkit.org/show_bug.cgi?id=165004)) — by the time the handler runs, `isComposing` is already false.

The (now closed) #73 contains a working implementation of exactly this, covering all three inputs, with `tsc --noEmit` passing — feel free to lift it from there or have your agents re-derive it.

## Expected behavior

- With an IME: Enter confirms the conversion; a subsequent Enter sends the message.
- Without an IME: unchanged (Enter sends, Shift+Enter inserts a newline, slash-command picker behavior intact).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

NightMustEnd:
I can reproduce this on macOS with Japanese IME as well.

Pressing Enter to confirm a Japanese IME conversion immediately submits the message, so half-composed text can be sent accidentally.

Expected behavior:
- Enter during IME composition should only confirm the conversion.
- A subsequent Enter, after composition has ended, should submit the message.

Thanks for tracking this.

PR #94, opened 2026-08-08
Preserve connection capsules during IME composition

### LINKED ISSUE #92: Connection capsule disappears when confirming IME text on iOS Safari
## Bug

When composing Japanese text in the chat composer on iOS Safari, confirming an IME conversion can remove an attached Connection capsule from the input.

This is related to, but distinct from, #81: that issue covers Enter incorrectly sending during IME composition, while this issue covers loss of a structured connection token.

## Steps to reproduce

1. Open Cloudflare OS in iOS Safari.
2. Add a Connection capsule to the chat composer.
3. Type Japanese text using the system IME.
4. Confirm a conversion.

## Actual behavior

The Connection capsule can disappear from the composer.

## Expected behavior

Confirming IME text should commit only the composed text. Existing Connection capsules must remain attached.

## Root cause

Mobile Safari emits controlled-textarea changes while composition is still in progress and can report a stale selection range. The composer currently runs token overlap bookkeeping for every provisional value. A provisional edit can therefore be misclassified as crossing the capsule range, causing the capsule to be removed.

## Proposed fix

Track composition explicitly. Continue rendering provisional text, but defer capsule and token bookkeeping until compositionend, then apply the committed edit once against the value captured at compositionstart. Also avoid handling Enter as a composer command while composition is active.

## Validation

- Frontend TypeScript build
- Production Vite build
- Manual reproduction on iOS Safari with a Connection capsule and Japanese IME

tktm8:
## Summary

- keep provisional IME text rendered without running connection-capsule overlap bookkeeping
- apply the committed edit once at composition end against the pre-composition value
- prevent composer commands and submission while the key event is part of active composition

## Root cause

Mobile Safari can emit controlled-textarea changes during IME composition with a stale selection range. Processing each provisional value as a normal edit can make the edit appear to overlap a structured Connection capsule, removing it from the composer.

## User impact

Japanese and other IME users can confirm text conversion in iOS Safari without losing an attached Connection capsule.

## Validation

- pnpm --filter @gadgets/workshop-frontend build

Closes #92.
Related to #81.
Supersedes #93 after renaming the fork branch.

### DIFF
--- a/packages/workshop-frontend/src/ChatInterface.tsx
+++ b/packages/workshop-frontend/src/ChatInterface.tsx
@@ -1892,6 +1892,13 @@ export const ChatInput = ({
   // Keep inputValue in a ref so handleCursorChange can read it without re-binding.
   const inputValueRef = useRef(inputValue);
   inputValueRef.current = inputValue;
+  // Mobile Safari emits controlled-textarea changes while an IME composition is still in
+  // progress. Treating each provisional value as an ordinary edit can make its occasionally stale
+  // selection range look as though the edit crossed a connection capsule, which removes the
+  // capsule. Keep rendering the provisional text, but defer token bookkeeping until composition
+  // ends and compare the committed value with the value from before composition began.
+  const isComposingRef = useRef(false);
+  const compositionStartValueRef = useRef("");
 
   // Seed the composer from an external suggestion (Home task cards). Re-runs whenever the nonce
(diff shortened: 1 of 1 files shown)


dcartertwo:
thank you for your contribution! this should be addressed now with https://github.com/cloudflare/cloudflare-os/pull/280

PR #98, opened 2026-08-09
Don't replay image attachments to models without image input

### LINKED ISSUE #97: Image attachment sent to a non-vision model permanently locks the chat
## Summary
Sending an image attachment to a model that does not support image input fails with provider's 400. Subsequent requests in the same chat replay the image on every request, effectively locking the chat.

## Example error:
```
400: {"message":"this model does not support image input (ref: ...)","type":"invalid_request_error","param":null,"code":null}
```

Image

## Steps to reproduce
1. Add a model that does not support image input (e.g. GLM-5.2)
2. Start a chat with that model suggested, attach and image and send a message. Upload validation will accept it, but request will fail with a 400 (as above).
3. Send any follow-up message in the same chat - no attachment this time.
4. *Bug*: the new requests will fail.

### Primary
Stop re-sending images the model can't accept.  When the agent rebuilds the message history for a request, check whether the current model supports image input, if it doesn't, don't include the image bytes; instead replace the image with a short text note like [Attached file photo.png (image/png) omitted — this file type is not supported by the current model].

### Secondary
Reject the upload up front. We should reject images being uploaded if the selected model does not support images (alternate — some sort of warning). This is a nice-to-have, but the primary fix would actually make it so that we both unstick chats AND allow multi-model chats with some image models and some non-image models to work well.

hmk:
Sending an image attachment to a model without image input fails with the provider's 400 — and because history replay rebuilds the model request from the chat log on every turn, the image is re-sent on every subsequent message, permanently wedging the chat.

This gates the image content part on the model's declared input modalities (`model.input`), degrading to the existing `[Attached file … omitted]` text marker instead — the same path other unsupported attachment types already take. PDFs get the same gate since they ride an image content part. The conversion is extracted into `chat-attachment-parts.ts` with tests, following the existing `chat-attachment-validation.ts` / `chat-attachment-pdf.ts` pattern.

Since replay rebuilds from the log each turn, chats that are already stuck recover on their next message.

Known limitation: custom/uncatalogued models default to `input: ["text", "image"]`, so a text-only custom model can still hit the 400. Discussion of how to handle that is on the issue.

Fixes #97

### DIFF
--- a/packages/workshop-backend/__tests__/chat-attachment-parts.test.ts
+++ b/packages/workshop-backend/__tests__/chat-attachment-parts.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, it } from "vitest";
+import type { Api, Model } from "@earendil-works/pi-ai";
+import { chatAttachmentModelParts } from "../src/chat-attachment-parts.js";
+
+// Verifies that an attachment the current model can't accept is replaced with a text marker
+// rather than sent as an image part; see the header of chat-attachment-parts.ts for why.
+
+function testModel(api: Api, input: ("text" | "image")[]): Model<Api> {
+  return {
+    id: "test-model",
+    name: "Test Model",
+    api,
+    provider: "anthropic",
+    baseUrl: "https://example.com",
+    reasoning: false,
+    input,
+    cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+    contextWindow: 128_000,
+    maxTokens: 4096,
+  };
+}
+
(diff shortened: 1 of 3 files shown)


hmk:

hmk:
for posterity: this PR was built with the assistance of Claude Code and Fable 5

hmk:
Closing this myself — rereading CONTRIBUTING, it's clear this is bigger than what you want in external PRs, and the preference is for issue reports over patches. #97 has the full report; leaving this up as reference in case any of it is useful.

PR #102, opened 2026-08-09
Apply the webFetch timeout to the response body

RenKoya1:
`webFetch` arms a 30s abort timer, but clears it in the `finally` of the `fetch()` call — which resolves as soon as the response *headers* arrive. The body read that follows therefore runs with no deadline: a response whose headers arrive promptly and whose body then stalls holds the agent's `webFetch` tool call open until the Worker itself gives out. `HARD_MAX_BYTES` bounds the size, not the time.

The change keeps the timer armed across the body read and clears it in a single `finally`, so an abort mid-body surfaces as the same `Fetch timed out after 30000ms` error the header path already produces. The diff looks larger than it is — most of it is the existing final-URL / Content-Signal / `readBodyCapped` block moving inside the `try` unchanged.

### Verification

Added a regression test that stubs `fetch` with a body that yields one chunk and then only ends on abort, and advances fake timers past the deadline.

- Without the fix: the test hangs and fails on the 5s vitest timeout; the captured `AbortSignal` is still un-aborted after 60 simulated seconds.
- With the fix: `webFetch` rejects with `Fetch timed out after 30000ms`.

`pnpm exec tsc --noEmit` clean, `workshop-backend` suite 280/280 passing, oxlint reports nothing new.

### DIFF
--- a/packages/workshop-backend/__tests__/web-fetch.test.ts
+++ b/packages/workshop-backend/__tests__/web-fetch.test.ts
@@ -261,6 +261,32 @@ describe("webFetch document conversion", () => {
     // only cares that the call didn't go through toMarkdown.
     expect(typeof result.body).toBe("string");
   });
+
+  it("times out a response whose body stalls after the headers arrive", async () => {
+    vi.useFakeTimers();
+    try {
+      // Headers arrive at once; the body yields one chunk and then never completes, as the
+      // runtime would present a stalled origin. Only an abort ends the read.
+      globalThis.fetch = vi.fn(async (_url: string, init: RequestInit) =>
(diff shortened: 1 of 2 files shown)


RenKoya1:

snowyukitty (approved):
Verified against current main (3562627): the commit cherry-picks cleanly, the targeted workshop-backend build (including tsc) passes, and the web-fetch test file passes 18/18 on Windows 11 with Node 22.22.3 and pnpm 11.17.0. The new stalled-body case exercises the gap described in #27, and keeping readBodyCapped inside the existing abort timeout scope fixes it without broadening the public API.

PR #104, opened 2026-08-09
Absorb user-DO resets in the Worker and self-heal subscriptions

ndisidore:
Supersedes #58 — same goal, restarted and reworked. #56 (the observability groundwork) is already on main, and the client-side half of this work — error classification, quieting, auth error codes, and the shared connected-accounts subscriber adapter — was extracted and merged separately as #110. This PR is the Worker-side recovery half, plus one small frontend subscription-hygiene fix.

Routine Durable Object resets — storage timeouts, overload aborts; ~3k over a recent 4-day window in production logs — were surfacing in the frontend as scary terminal errors, and anything subscribed to the user DO stayed silently broken until a page reload. The reject frames carry structured flags (`durableObjectReset`, `retryable`, … thanks to `enhanced_error_serialization`), but nothing read them.

terminal error screenshot
terminal error screenshot

## What changed

**The Worker absorbs user-DO resets.** Stubs are resolved fresh per call (a stub dies with its incarnation), and every user-DO RPC goes through one of three chokepoints: reads retry once on reset (jittered, same-colo), writes never retry — the reset can land after the commit — and serialize per session to keep their order, and subscription registration gets neither.

**Subscriptions self-heal.** The DO holds registrations in memory only, so the session stamps each with the DO's incarnation id and re-subscribes any whose stamp goes stale (detected via error flags, a throttled check after successful calls, or a mismatched new subscribe). The catch-up replay is pure upsert; the shared client adapter that consumes it landed in #110.

**ResourcePicker disposes subscriptions that resolve after unmount** — a small standalone fix for a leaked registration the self-heal work surfaced.

## Accepted gaps (deliberate)

- An account disconnected during the seconds a registration is dead ghosts in the UI until the next re-subscribe. Removes have exactly one source today (the owner's own disconnect), and the alternative was worse: a replay-reconciliation protocol was built and then deleted, because the DO's replay is lossy by design and pruning against it removed live accounts more often than ghosts.
- A fully idle session learns nothing until its next user-DO call; the client's socket-level reconnect remains the backstop.
- Workspace (Overseer) DO resets keep their pre-existing session-abort handling — out of scope here; a follow-up PR (workspace reopen) addresses them client-side.

## Testing

- **Integration** (`abortAllDurableObjects()`, the non-graceful teardown): the same session recovers transparently after a reset, live subscriptions recover via incarnation drift, and the local-aborts-reject-flagless shape is pinned so the flag paths can graduate to real integration tests if a future pool upgrade attaches the production flags.
- **Unit**: reset predicate/retry and the incarnation/replay behavior. (The frontend classifier canaries and subscriber-adapter contract tests merged with #110.) 290 backend unit + 4 integration + 136 frontend green on this branch.
- **Manual fault injection** (chrome-devtools against a local dev stack, temporary `debugAbort()` injector, composed with the #110 client): 72 aborts across idle, live-subscription, reload-race, and SPA-navigation-storm scenarios — every page functional, zero application console output, and an OAuth connect performed after an abort arrived live through the recovered subscription (no remount). A chat send racing a workspace-DO abort showed the inline hint (#110's UI), preserved the draft, and did not double-send. Server logs: 3× `subscription.recovered`, 0× `recover_failed`.

### DIFF
--- a/packages/workshop-backend/__integration__/open-gadget-rpc.test.ts
+++ b/packages/workshop-backend/__integration__/open-gadget-rpc.test.ts
@@ -1,5 +1,6 @@
+import { abortAllDurableObjects } from "cloudflare:test";
 import { exports } from "cloudflare:workers";
-import { newWebSocketRpcSession, type RpcStub } from "capnweb";
+import { newWebSocketRpcSession, RpcTarget, type RpcStub } from "capnweb";
 import {
   createOpenGadgetError,
   getOpenGadgetErrorCode,
@@ -8,7 +9,7 @@ import {
   type OpenGadgetErrorCode,
   type PublicApi,
 } from "@gadgets/workshop-shared/api";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 
 type CodedError = Error & { code?: unknown };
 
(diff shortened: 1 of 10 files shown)


packages/workshop-frontend/src/accountsSubscriber.ts
ndisidore, line 1: This looks like a lot of new work, but if you look at the diff, you'll see OnboardingWizard, ConnectionChips, ... many files (about 7) implemented this same pattern - this just pulls it out in a way that it can be both reused and tested

packages/workshop-backend/src/server.ts
```diff
@@ -87,8 +116,196 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
+  async #userRead<T>(operation: string,
+               fn: (user: DurableObjectStub<UserDurableObject>) => Promise<T>): Promise<T> {
+    let result = await retryOnDoReset(
+        () => fn(this.#user),
+        info => this.#onUserDoReset(operation, info.error, info.attempt, info.delayMs));
+    this.#maybeRevalidateSubscriptions();
+    return result;
+  }
+
+  /** Non-idempotent call: NO retry — a reset can land after the write committed but before the
```
ndisidore: I audited every frontend write call site before adding this chain, and several overlap. The quick-model row (`handleSetQuick` in `providers.tsx`) is a whole-row click target with no busy state, so a double-click issues two concurrent `setQuickModel` writes to a single-valued field and reversed arrival persists the wrong one. Blueprint pin/unpin from the kebab menu has the same shape. Per-call stubs have no cross-call delivery-order guarantee, so this chain is what makes those safe.

packages/workshop-backend/src/server.ts
```diff
@@ -87,8 +116,196 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
+    logger.warn("user DO reset observed", {
+      event: "user_do.reset.recovered",
+      operation, attempt, durationMs: delayMs,
+      durableObjectId: this.#userId.toString(),
+      error,
+    });
+    this.#triggerRecovery();
+  }
+
+  // ── subscription recovery ─────────────────────────────────────────────────────────────────
```
ndisidore, line 237: The obvious alternative to all this is what the workspace DO already does: `abortSession()` and let the client's reconnect rebuild everything. I didn't do that because these resets are routine (~3k over a recent 4-day window in prod logs, mostly storage timeouts) and each abort is a visible reconnect blip plus a refetch storm. The incarnation-stamp design is also a stopgap that deletes itself if native stubs ever grow `onRpcBroken`; see the TODO below.

We could still go this route, for subscriptions, but it gives up reset invisibility for sessions with live subscriptions and because ConnectionChips sits in the app shell, essentially every authenticated session has a live subscription at all times. So "abort when subscriptions died" is in practice "abort on every reset the session detects."

packages/workshop-backend/src/server.ts
```diff
@@ -107,56 +324,56 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
   }
 
   whoami(): Promise<AiChatAuthorInfo> {
```
ndisidore, line 406: Everything from here down is mechanical: each API method declares whether it's a read (one retry on reset) or a write (no retry, ordered). The interesting code is the three chokepoints above. The only judgment calls are the methods that look like reads but are writes; I flagged those where they occur (`listGatekeeperApps`, `getGatekeeperApp`).

packages/workshop-backend/src/server.ts
```diff
@@ -549,9 +804,11 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
   // accounts are auto-provisioned singletons (one per vendor), so the vendor id identifies them.
   async listGatekeeperApps(): Promise<GatekeeperAppInfo[]> {
-    // listProvidedAccounts provisions auto-provisioned accounts first (idempotent), so their apps
-    // appear in the nav even before the user opens a gadget — in a single round trip.
-    let accounts = await this.user.listProvidedAccounts();
+    // listProvidedAccounts provisions auto-provisioned accounts first, so their apps appear in
+    // the nav before the user opens a gadget. #userWrite, not #userRead: provisioning calls the
+    // vendor's createAccount before the record persists, so a reset-then-retry in that window
+    // would create a duplicate vendor-side account. Same in getGatekeeperApp.
+    let accounts = await this.#userWrite("listGatekeeperApps", u => u.listProvidedAccounts());
```
ndisidore: This is the subtlest classification in the file. `listProvidedAccounts` provisions auto-provisioned accounts, and the vendor's `createAccount` runs before the record persists. A reset in that window followed by a retry would create a duplicate vendor-side account (the DO's dedup promise is in-memory and dies with the reset). So it's a write despite the name.

packages/workshop-backend/src/do-retry.ts
```diff
@@ -0,0 +1,48 @@
...
+// whenever the object's incarnation died — the production storage-timeout reset arrives as
+// `{remote, overloaded, durableObjectReset}`. The flags are attached natively in the calling
+// Worker, so no message matching is needed. Local vitest-pool-workers aborts reject FLAGLESS
+// (pinned by the "user-DO reset flags" integration test), so these paths are unit-tested with
+// synthetic errors and integration tests assert recovery behaviorally.
+
+/** True for rejections caused by a DO reset or lost connection — cases where a retry of an
+ * IDEMPOTENT call through a fresh stub is expected to succeed against the restarted object.
+ * `overloaded` alone is deliberately excluded (the object is alive and shedding load; a retry
+ * adds to it — the docs' "never retry .overloaded" case). An overloaded RESET is retried
```
ndisidore, line 14: This knowingly deviates from the [DO error-handling docs](https://developers.cloudflare.com/durable-objects/best-practices/error-handling/), which say never retry `.overloaded`. The carve-out only applies when `durableObjectReset` is also set (the production storage-timeout shape), where the overloaded queue died with the incarnation. It's bounded to one jittered retry against an object that serves a single user, so there's no herd to amplify.

packages/workshop-backend/__integration__/open-gadget-rpc.test.ts
```diff
@@ -133,3 +134,130 @@ describe.skip("openGadget errors across native RPC and Cap'n Web", () => {
...
+
+    const nativeErr = await rejection(userStub.listModels());
+    expect({
+      message: nativeErr.message,
+      durableObjectReset: (nativeErr as Record<string, unknown>).durableObjectReset,
+      retryable: (nativeErr as Record<string, unknown>).retryable,
+      overloaded: (nativeErr as Record<string, unknown>).overloaded,
+    }).toEqual({
+      message: "Application called abortAllDurableObjects().",
+      durableObjectReset: undefined,
```
ndisidore, line 173: This assertion looks backwards on purpose. It pins that local aborts reject flagless, which is why the flag-driven paths can only be unit-tested with synthetic errors. If a future vitest-pool-workers upgrade starts attaching the production flags, this test fails and the flag paths can graduate to real reset integration tests.

Maximo-Guk:
From a first pass with GPT:

```
1. High: ResourcePicker.tsx:100-136 no longer reconciles cached accounts on initial ready(). Disconnecting an account while unmounted leaves it selectable indefinitely. Reconcile only the initial census, not recovery replays.
2. Medium: server.ts:321-328 iterates a live Set. A subscription added after a second reset can be mistaken as stale and disposed; if replacement fails, its current stamp causes future recovery passes to skip the dead registration.
3. Medium: server.ts:455-456 expects reset retries, but connection-service.ts:142-151 catches reset errors and returns []. The query neither retries nor triggers subscription recovery, falsely showing no Cloudflare accounts.
```

PR #105, opened 2026-08-09
Codex/fix gatekeeper app accent theme

deloreyj:
After setting up my Cloudflare OS instance with a custom accent color, I noticed that the primary call to action buttons in the Schedule and Context tabs were using the default orange. This pull request propagates the accent colors into the built-in gatekeeper iframe. 

The change is narrowly focused on the accent color and follows the same pattern as the theme subscription.

### DIFF
--- a/packages/gatekeeper-context/app/main.tsx
+++ b/packages/gatekeeper-context/app/main.tsx
@@ -32,6 +32,7 @@ interface HostCapability extends RpcTarget {
   setPresenting(active: boolean): Promise<PresentAck>
   // Returns the current resolved theme mode and calls back on `receiver` whenever it changes.
   subscribeTheme(receiver: AppIframe): Promise<ResolvedThemeMode>
+  subscribeAccent(receiver: AppIframe): Promise<Record<string, string> | null>
 }
 
 function main() {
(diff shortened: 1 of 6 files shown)


deloreyj:

Maximo-Guk:
Would it be possible to provide some before and after screenshots of your change?

ndisidore (approved):
This is looking good to me! I looked briefly into this and came to a similar changeset.
Agree before/after pictures would be nice to have

deloreyj:
Sure thing! cc @Maximo-Guk and @ndisidore 

Before: New collection and create buttons button are default Cloudflare Orange despite the app theme being changed. However the New Workspace button correctly adopts the new theme
image
image
image


After: The app passes through the user-specified app theme to the built-in gatekeepers' iframes
image
image

ndisidore:
@ask-bonk review this

ndisidore:
Checking this out and testing locally I found a few issues e.g. ServerConfigContext, which is likely only populated at boot, while the admin color picker updates the host DOM directly

Also I think it probably makes sense to roll into the single theme RPC:
I took a stab at what I think may be a better approach here: https://github.com/cloudflare/cloudflare-os/pull/155 utilizing [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore)

PR #107, opened 2026-08-09
fix: invoke pnpm via npm_execpath on Windows (fixes #19)

### LINKED ISSUE #19: pnpm run-local cannot start on Windows: spawnSync pnpm ENOENT
## Summary

`pnpm run-local`, the command the README gives for trying the Workshop locally, cannot start on
Windows. It fails on the first nested pnpm call, before anything is installed or built.

```
> gadgets@1.0.0 run-local D:\...\cloudflare-os
> node scripts/run-local.mjs

> pnpm install
node:internal/child_process:1120
    result.error = new ErrnoException(result.error, 'spawnSync ' + options.file);
                   ^
<ref *1> Error: spawnSync pnpm ENOENT
    at execFileSync (node:child_process:911:15)
    at run (file:///D:/.../cloudflare-os/scripts/run-local.mjs:118:3)
```

Environment: Windows 11, Node v22.12.0, pnpm 10.6.3, clean clone of `main` (e1ab8fb).

## Cause

On Windows the `pnpm` on `PATH` is a `.cmd` shim. `execFileSync` starts a process directly instead
of going through a shell, so there is no extensionless `pnpm` for it to execute and the call fails
with `ENOENT`. Naming the shim explicitly does not help either — since the fix for CVE-2024-27980,
Node refuses to spawn `.bat`/`.cmd` without a shell and returns `EINVAL`:

```
execFileSync('pnpm.cmd', ['--version'])   ->  EINVAL  spawnSync pnpm.cmd EINVAL
execFileSync('pnpm',     ['--version'])   ->  ENOENT  spawnSync pnpm ENOENT
```

## Affected call sites

Four sites are on the `pnpm run-local` / `pnpm dev-server` path. Fixing only the first just moves
the failure to the next one:

| File | What it runs |
| --- | --- |
| `scripts/run-local.mjs` (`run()`) | `pnpm install`, and both builds |
| `run-dev-server.js` | `pnpm exec wrangler dev` |
| `packages/gatekeeper-context/build-app.mjs` | `pnpm exec vite build` |
| `packages/gatekeeper-scheduler/build-app.mjs` | `pnpm exec vite build` |

`scripts/release/build-release.mjs` has the same pattern in its own `run()` helper, but that path is
CI-only, so it is not part of the failure above.

Everything else in these scripts already spawns `process.execPath` or `git`, which are real
executables and unaffected.

## Two things worth knowing before fixing it

**`shell: true` is not a safe fix here.** It makes the call work, but a shell re-splits the command
line, and `run-dev-server.js` passes wrangler absolute config paths built from `gk.dir`. On a
checkout whose path contains a space — `C:\Users\Some Name\...` — those arguments break apart.

**`npm_execpath` needs a guard.** It holds pnpm's own JS entry point, which `node` can run directly
with no shell, so arguments keep their exact values. But under `npm run` the same variable points at
npm instead:

```
pnpm run ...  ->  npm_execpath = ...\pnpm\bin\pnpm.cjs
npm  run ...  ->  npm_execpath = ...\npm\bin\npm-cli.js
```

Substituting it unchecked would silently run `npm install` against a pnpm workspace.

## Patch

Happy to send this as a PR if you want it — flagging it here first because it lands at about 30
changed lines across five files, which is past the guideline in CONTRIBUTING.md.

A small shared helper keeps it to one copy (`packages/*` already reference `../../scripts/`, e.g.
`build:configurator`):

```js
// scripts/pnpm-command.js
      : ["pnpm", args];
}
```

used as `const [file, argv] = pnpmCommand([...]); execFileSync(file, argv, opts);` at the four sites.
It is a no-op off Windows, and on Windows it falls back to today's behaviour whenever the launcher
was not pnpm, so the failure stays loud rather than turning into the wrong package manager.

## Verification

On Windows 11 / Node v22.12.0, from a clean clone:

- before: `pnpm run-local` dies at `pnpm install` with the `ENOENT` above;
- after: it completes install, `@gadgets/typed-storage` build and the frontend `vite build`, runs
  both gatekeeper `build-app.mjs` bundles, and reaches
  `[wrangler:info] Ready on http://127.0.0.1:8787` with the Workshop serving.
- `pnpm lint:check` passes.
- The helper was exercised under `pnpm run`, `npm run`, and bare `node` to confirm the guard picks
  pnpm's entry point only in the first case.

Not verified: Linux/macOS behaviour is unchanged by inspection only (the platform check makes it a
no-op), and I did not run `pnpm test` — the release-manifest golden test already fails on Windows
before any change, from CRLF line endings.

Sainigurnoor511:
## Summary

- Add \scripts/pnpm-command.mjs\ to resolve pnpm invocations on Windows without a shell: when \
pm_execpath\ points at pnpm's \.cjs\/\.mjs\ entry (as under \pnpm run\), run \
ode <execpath> ...\ so argv boundaries survive intact.
- Reject \
pm-cli.js\ via the path guard so \
pm run\ does not silently use the wrong package manager.
- Update all nested pnpm call sites on the local dev path, plus \scripts/generate-worker-types.mjs\ (flagged in #19).
- Surface spawn failures in \generate-worker-types.mjs\ when \
esult.error\ is set (misleading wrangler error today).

Fixes #19

## Approach

Follows the approach discussed in #19 / @snowyukitty's analysis: no shell, no \.cmd\ shim, arguments stay intact (including checkout paths with spaces).

## Test plan

- [x] Windows 10 (10.0.26200), Node v24.16.0, pnpm 11.17.0, path \E:\\Projects\\cloudflare-os\ (no spaces): \pnpm run-local\ completes install/build and reaches wrangler ready on port 8787
- [ ] Checkout path containing a space (regression case from #19)
- [ ] \pnpm types:generate\ on Windows

## CLA

### DIFF
--- a/packages/gatekeeper-context/build-app.mjs
+++ b/packages/gatekeeper-context/build-app.mjs
@@ -1,8 +1,8 @@
 // Build the Context Library SPA into generated single-file HTML for startAppUi().
 
-import { execFileSync } from "node:child_process";
 import { resolve } from "node:path";
 import { fileURLToPath } from "node:url";
+import { execPnpm } from "../../scripts/pnpm-command.mjs";
 
 const pkgDir = resolve(fileURLToPath(import.meta.url), "..");
 const watch = process.argv.includes("--watch");
@@ -12,8 +12,7 @@ console.log(
     ? "watching context library app for changes…"
     : "building context library app single-file bundle…",
 );
-execFileSync(
-  "pnpm",
+execPnpm(
   ["exec", "vite", "build", "-c", "vite.config.ts", ...(watch ? ["--watch"] : [])],
   { cwd: pkgDir, stdio: "inherit" },
 );
--- a/packages/gatekeeper-scheduler/build-app.mjs
+++ b/packages/gatekeeper-scheduler/build-app.mjs
@@ -1,12 +1,11 @@
-import { execFileSync } from "node:child_process";
 import { resolve } from "node:path";
 import { fileURLToPath } from "node:url";
+import { execPnpm } from "../../scripts/pnpm-command.mjs";
 
 const packageDirectory = resolve(fileURLToPath(import.meta.url), "..");
 const watch = process.argv.includes("--watch");
 
-execFileSync(
-  "pnpm",
+execPnpm(
   ["exec", "vite", "build", "-c", "vite.config.ts", ...(watch ? ["--watch"] : [])],
   { cwd: packageDirectory, stdio: "inherit" },
 );
--- a/run-dev-server.js
+++ b/run-dev-server.js
@@ -15,6 +15,7 @@ import { execFileSync, spawn } from "node:child_process";
 import { join, dirname } from "node:path";
 import { fileURLToPath } from "node:url";
 import { parse } from "jsonc-parser";
+import { execPnpm } from "./scripts/pnpm-command.mjs";
 import { getWranglerPortFromBackendHost } from "./scripts/dev-server-config.js";
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
@@ -322,7 +323,7 @@ if (backendHost) {
 console.log(`\nStarting: wrangler dev ${args.join(" ")}\n`);
 
 try {
-  execFileSync("pnpm", ["exec", "wrangler", "dev", ...args],
+  execPnpm(["exec", "wrangler", "dev", ...args],
       { stdio: "inherit", cwd: ROOT });
 } catch (e) {
   // wrangler was killed or exited with an error; the output was already shown
--- a/scripts/generate-worker-types.mjs
+++ b/scripts/generate-worker-types.mjs
@@ -14,10 +14,10 @@
  *
  * `--check` is non-mutating: writes a sibling temp file, compares, deletes it.
  */
-import { spawnSync } from "node:child_process";
 import { readdir, readFile, rm, writeFile } from "node:fs/promises";
 import { dirname, join, relative, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
+import { spawnPnpmSync } from "./pnpm-command.mjs";
 
 const root = resolve(dirname(fileURLToPath(import.meta.url)), "..");
 const packagesDir = join(root, "packages");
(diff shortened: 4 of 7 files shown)


Sainigurnoor511:

snowyukitty:
I tested the exact PR head (`b289c1a`) against the regression case from #19:

- Windows, Node 22.22.3, pnpm 11.17.0
- checkout path containing a space
- cold `pnpm run-local`, through install, both app builds, Wrangler custom builds, and startup
- `GET /` returned HTTP 200 (653 bytes)

So the core `npm_execpath` approach in this revision does preserve argv boundaries in a path containing a space. I did not run `pnpm types:generate` separately.

Two current-main deltas are worth knowing before anyone rebases it:

1. `main` is now at `ba4036b`, and GitHub reports this branch as conflicting. The Vite+ dev-server work added two asynchronous bare-pnpm preflight calls at [`run-dev-server.js:261-264`](https://github.com/cloudflare/cloudflare-os/blob/ba4036b93660/run-dev-server.js#L261-L264). Those still need the same Windows-safe executable/argv resolution, but this helper currently wraps only `execFileSync` and `spawnSync`. In the same environment, a direct `spawnSync("pnpm", ["--version"])` on current main returns `ENOENT`.
2. `scripts/release/build-release.mjs` documents direct `node scripts/release/build-release.mjs ...` invocation. In that entry path `npm_execpath` is not guaranteed, so this revision's helper falls back to bare `pnpm`; routing that call site through `execPnpm()` does not by itself make the direct Windows path work. I would leave it out unless the release script gets an explicit pnpm-entry contract.

One process caveat: the contribution-policy automation landed after this PR and now closes non-exempt external PRs over 30 changed lines on `synchronize`; this PR is currently 60 additions / 16 deletions. I would wait for maintainer direction before spending time on a rebase.

AI tools assisted the investigation. The environment, HTTP response, and spawn result above are measurements from the runs described, not predicted outcomes.

Maximo-Guk:
Hi @Sainigurnoor511 , I've gone ahead and integrated your approach in #235. Thanks for your contribution!

Sainigurnoor511:
> Ho @Sainigurnoor511 , I've gone ahead and integrated your approach in #235. Thanks for your contribution!

Thanks for picking it up and pushing it over the line — appreciate it 🫱🏻‍🫲🏻

PR #112, opened 2026-08-10
Name three more reserved ranges in the MCP endpoint blocklist

RenKoya1:
`BLOCKED_HOST_PATTERNS` in `mcp-shared/src/endpoint.ts` is explicitly not the SSRF boundary — `global_fetch_strictly_public` is, and the file says so. What it is for is the other thing the comment names: giving a user typing a URL into the connect form a legible refusal instead of an opaque network error later. Three ranges of exactly the same class as the entries already listed were missing from it:

| range | note |
| --- | --- |
| `100.64.0.0/10` | carrier-grade NAT; some networks route it to infrastructure of their own |
| `fe80::/10` | IPv6 link-local — only `fc00::/7` (`f[cd]xx:`) was covered |
| `::` | the unspecified address — only `::1` was covered |

Nothing else changes: the patterns keep the existing bracket-tolerant style, and `normalizeHost` already feeds them the canonical spelling.

### Verification

Added the new hosts to the existing blocked-hosts case, plus `100.63.0.1`, `100.128.0.1` and `[2606:4700::1111]` to the accepted-hosts case so the new patterns are pinned on both sides of their boundaries. That test fails on `main` (`100.64.0.1: expected true to be false`) and passes here.

`pnpm exec tsc --noEmit` clean, full `mcp-shared` suite 230/230, oxlint reports nothing new.


RenKoya1:

PR #114, opened 2026-08-10
proposal: add entrypoint-based custom export api

maxwellpeterson:
This PR proposes a new file export API for Gadgets. Currently, Gadgets can only be exported to PDF by rendering their UI in a remote browser and printing the page:

image

We want Gadgets to be able to register their own custom export formats. For example, a document editor Gadget could support Markdown, HTML, and PDF exports, and a data analysis Gadget could support CSV and XLSX exports:

image

The API interface described in this PR allows Gadgets to customize their own export behavior, and has several desirable attributes:
- All Gadgets can be exported as HTML or PDF by default without implementing the export handler interface
- Gadgets can support custom file formats as well as multiple HTML or PDF variants (e.g. with different content) by implementing the export handler interface
- Gadgets can disable or reimplement the default HTML and PDF export formats for additional customization if needed
- The export handler interface implemented by the Gadget is only callable by the workshop backend and does not change the public RPC interface of the Gadget

This PR exists to illustrate the proposed API interface. It's meant for discussion, and not intended to be merged as-is.

### DIFF
--- a/packages/workshop-shared/src/export.ts
+++ b/packages/workshop-shared/src/export.ts
@@ -1,0 +1,116 @@
+import type { DurableObject, WorkerEntrypoint } from "cloudflare:workers";
+
+// Gadget-facing export API:
+
+/** Name under which a Gadget may export its optional export handler entrypoint. */
+export const GADGET_EXPORT_ENTRYPOINT = "ExportHandler";
+
+/**
+ * Describes an export format that a Gadget supports. Export formats are split
+ * into two modes:
+ *
+ * Browser mode: The Gadget UI is loaded in a remote browser and the rendered
+ * output is captured, either by printing the page to PDF or returning the
+ * underlying HTML. Browser mode exports are limited to a small set of formats
+ * that can be easily captured using the Browser Run binding.
+ *
+ * Custom mode: The server-side export handler entrypoint returns the file
+ * content directly. Export formats are unrestricted, and a Gadget may (on rare
+ * occasion) reimplement a browser-mode export format server-side if it wants
+ * total control over the exported file content.
+*/
+export type GadgetExportFormat =
+  | {
+    // A Gadget can support multiple HTML or PDF export variants by returning
+    // multiple browser-mode export entries with different ids. When the Gadget
+    // UI is loaded in a remote browser, globalThis.gadgetExportFormatId will be
+    // set to the selected format id so the Gadget UI knows which variant to
+    // render.
+    id: string;
+    // User-facing label for the export format. Usually "HTML" unless multiple
+    // HTML variants are supported
+    label: string;
+    mode: "browser";
+    contentType: "text/html";
+  }
+  | {
+    id: string;
+    // User-facing label for the export format. Usually "PDF" unless multiple
+    // PDF variants are supported
+    label: string;
+    mode: "browser";
+    contentType: "application/pdf";
+  }
+  | {
+    id: string;
+    label: string;
+    mode: "custom";
+    contentType: string;
+    fileExtension: string;
+  };
+
+/**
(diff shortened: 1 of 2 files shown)


packages/workshop-shared/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+  | {
+    /** Unique, non-empty identifier for this target. */
+    id: string;
+
+    /** User-facing label for the export target. */
+    label: string;
+
+    /** Selects server-side generation by the Gadget's export handler. */
+    mode: "custom";
```
kentonv: Maybe "server"? Or "worker"?

I imagine in the future the browser mode may be extended with some client-side API to give better control over the rendering, but then "custom" makes less sense as a mode since both modes involve customization...

packages/workshop-shared/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+    /** User-facing label, usually "HTML" unless multiple variants are supported. */
+    label: string;
+
+    /** Selects browser-based rendering. */
+    mode: "browser";
+
+    /** Media type produced by this browser-based target. */
+    contentType: "text/html";
+  }
+  | {
```
kentonv: Hmm, I'm not feeling the union here. I'd just collapse them all into one record type I think, with documentation saying that "browser" mode only supports "text/html" and "application/pdf". (Should we also support image formats as screenshots?)

packages/workshop-shared/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+ * output is captured, either by printing the page to PDF or returning the
+ * underlying HTML. Browser mode exports are limited to a small set of file
+ * formats that can be easily captured using the Browser Run binding.
+ *
+ * Custom mode: The server-side export handler entrypoint returns the file
+ * content directly. Export file formats are unrestricted, and a Gadget may (on
+ * rare occasion) reimplement a browser-mode file format server-side if it wants
+ * total control over the exported file content.
+*/
+export type GadgetExportTarget =
```
kentonv: Maybe `GadgetExportFormat`?

packages/workshop-shared/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+  export(gadget: Fetcher<Gadget>, id: string): Promise<ReadableStream<Uint8Array>>;
+}
+
+// Export API exposed to the Workshop UI:
+
+/**
+ * Describes an export target that a Gadget supports. Hides browser- vs.
+ * custom-mode and default target implementation details.
+ */
+export type GadgetClientExportTarget = {
```
kentonv: Hmm, is it really worth defining a separate type to hide the `mode`? I think I'd reuse the same type for simplicity.

packages/workshop-shared/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * File extension, including the leading dot. Inferred by the Workshop for
+   * browser-mode exports.
+   */
+  fileExtension: string;
+};
+
+/**
+ * Export methods that will be added directly to `GadgetClient`, rather than
```
kentonv, line 86: Are you writing this here for now just for illustration, but planning to move these methods to `GadgetClient` later, or is the idea that `GadgetClient` will inherit this?

I would prefer to keep it simple and declare them directly on `GadgetClient`.

kentonv (commented):
Good design, just some aesthetic comments.

packages/workshop-shared/src/export.ts
maxwellpeterson: This is just for illustration, these methods will be declared directly on `GadgetClient`

kentonv:
Superseded by #197.

PR #123, opened 2026-08-10
Add AI binding support for pi

Maximo-Guk:
Why?
We want to enable users to use AI gateway without needing to provide an API token ( which puts them at risk of credential expiry, rotation, etc ) and makes them need to visit the dashboard just to provision a token during the deploy flow.

How?

Workers AI Gateway can now be accessed over the AI binding!

```
AI.fetch(https://workers-binding.ai/ai-gateway/gateways/platform-gateway/*)
```

How was this tested?

You can test it with the updated deploy flow which is based off this branches head https://gadgets-deploy.gadgets-staging.workers.dev/deploy ( notice you don't need to provision an api token ), and also locally.


docs/ai-gateway-billing.md
```diff
@@ -54,19 +53,22 @@ CLOUDFLARE_OAUTH_CLIENT_SECRET=...
...
 
-# Required whenever CF_AI_GATEWAY is set (all inference goes over HTTPS with tokens):
+# Required whenever CF_AI_GATEWAY is set:
 CF_AI_GATEWAY_ACCOUNT_ID=...
+# Required unless the WORKERS_AI binding carries gateway traffic; always required for the
+# google provider:
 CF_AI_GATEWAY_API_TOKEN=...
-
-# To send Workers AI straight to its REST endpoint (no gateway, no cost logs):
-CF_AI_GATEWAY_WAI_DIRECT=true
```
Maximo-Guk, line 62: It's unclear to me why we introduced a `CF_AI_GATEWAY_WAI_DIRECT` env var in the first place, I don't see why we'd want workers-ai traffic to not go through the gateway. I think the `CF_AI_GATEWAY_USE_BINDING=false` makes a lot more sense, it opts out all the providers from flowing through the binding transport and with `CF_AI_GATEWAY_ACCOUNT_ID` allows you to use a cross account gateway for all providers

kentonv:
Looks like upstream PR was merged, shall we update to use that?

Maximo-Guk:
Just waiting on a PI release before we can use the upstreamed changes, I'll put up a PR to update once it gets released.

The [vendored transport changes](https://github.com/cloudflare/cloudflare-os/pull/123/changes#diff-0d5f7afbda36526cb07507a5a0d0d1f73daee6cce99ac3979deb4fbca3baea70) in this PR are identical to what got merged in to PI, I would like to get this out so we can start battle testing what we upstreamed in to PI.

Maximo-Guk:
Drafting this for the time being as I work through on updating the PI implementation to be smarter & more efficient

Maximo-Guk:
bonk break glass

docs/ai-gateway-billing.md
kentonv: I think this may have been an artifact of an early misunderstanding where I didn't realize AI Gateway was available via the Workers AI binding. Yeah I think we can remove.

scripts/preview/staging-config.ts
```diff
@@ -592,22 +597,38 @@ export function resolveAiGateway({
...
-      ...(accountId ? [] : ["CF_AI_GATEWAY_ACCOUNT_ID"]),
-      ...(apiToken ? [] : ["CF_AI_GATEWAY_API_TOKEN"]),
-    ];
-    throw new Error(`${missing.join(" and ")} must be set when CF_AI_GATEWAY is: inference goes ` +
-        "over HTTPS with a Run + Read token, so the backend refuses to start a chat without it");
+  if (!accountId) {
+    throw new Error("CF_AI_GATEWAY_ACCOUNT_ID must be set when CF_AI_GATEWAY is: the backend " +
+        "cannot discover its own account, and refuses to start a chat without it");
+  }
+  if (useBinding !== undefined && useBinding !== "true" && useBinding !== "false") {
```
ndisidore, line 607: nit: could normalize to the make it more foolproof
e.g. `useBinding.toLowercase().trim()`

ndisidore (approved):

packages/workshop-backend/src/ai-models.ts
```diff
@@ -420,48 +446,41 @@ function getModelViaGateway(
...
+  // is active (gwConfig.binding is unset when CF_AI_GATEWAY_USE_BINDING=false opts out) --
+  // even for Google inference, which itself rides HTTPS (see AiGatewayConfig.bindingFor).
+  const logRoute = (gateway: string): AiGatewayLogRoute => gwConfig.binding
+      ? { gateway }
+      : { gateway, accountId: gwConfig.accountId, apiToken: gwConfig.apiToken! };
+
+  if (config.provider === "google" && !gwConfig.apiToken) {
+    // Unreachable when google is an enabled provider (the config constructor throws), but a
+    // stored model config can still name google directly.
+    throw new Error("Google models require CF_AI_GATEWAY_API_TOKEN (the Workers AI binding " +
```
kentonv: Hmm why is this?

packages/workshop-backend/src/ai-gateway.ts
```diff
@@ -9,33 +9,66 @@ const QUICK_MODEL_ID = "@cf/meta/llama-3.3-70b-instruct-fp8-fast";
...
+   * reach gateways in the Worker's own account, so this is the platform gateway unless
+   * CF_AI_GATEWAY_USE_BINDING=false marks it cross-account.
+   */
+  readonly sameAccountGateway?: string;
   readonly accountId: string;
-  readonly apiToken: string;
+  readonly apiToken?: string;
+  /**
+   * Workers AI binding, used as the gateway transport whenever present unless
+   * CF_AI_GATEWAY_USE_BINDING=false opts out: binding requests are pre-authenticated in-account,
```
kentonv, line 37: What's the benefit of the opt-out flag vs. just unbinding WORKERS_AI?

scripts/preview/staging-config.ts
Maximo-Guk: Done!

packages/workshop-backend/src/ai-models.ts
Maximo-Guk: Added [this commit](https://github.com/cloudflare/cloudflare-os/pull/123/changes#diff-cd5b38b2c9d973789c5f130fcc9b8055bb4825ad0dc214ac58555a953065c2b0R11) which explains it better and provides justification 

Basically PI's google adapter doesn't support custom fetch https://github.com/earendil-works/pi/blob/v0.84.2/packages/ai/src/api/google-generative-ai.ts#L80 so it can't use the binding ( since the binding uses custom fetch )

packages/workshop-backend/src/ai-gateway.ts
Maximo-Guk: Basically webFetch's `.toMarkdown()` requires the binding. So maybe you want to opt out your regular traffic but still keep webFetch's on the binding. ( This is what we'll be using in our deployment too )

kentonv (approved):

Maximo-Guk:
bonk break glass

PR #124, opened 2026-08-10
Reopen the workspace after a Durable Object reset

ndisidore:
Sibling of #104 (the user-DO half) building on #110. This PR is the workspace half, plus a rework of the socket-reconnect path it leans on.

**The problem.** When a workspace's Overseer DO resets (storage timeouts, overload, ... clusters #56 measured), the long-lived `open()` call dies along with every capability pipelined on it, but the WebSocket underneath stays perfectly healthy. Our only break signal, `onRpcBroken`, is session-level (no way to signal "one capability died") so the client hears nothing. The workspace looks fine, but every chat send fails with "Failed to start conversation" until a full page reload. Reproduced on a pre-PR baseline. (The server already tries to convert DO-loss into a socket close (the `notifyClosed` hack in `server.ts`) but it demonstrably doesn't fire for these resets, and its own TODO asks for exactly this client-side fix.)

## Solve

```mermaid
flowchart TD
    X([something breaks]) --> D{what died?}

    D -- "the WebSocket —<br/>onRpcBroken fires" --> C["connection manager:<br/>jittered backoff, probe round-trip,<br/>proven before publish"]
    C --> N["new session: app rebuilds everything,<br/>one aggregated lost/restored report"]

    D -- "the workspace DO —<br/>socket healthy, no signal at all" --> Q{was a call in flight?}
    Q -- "yes: it rejects with a<br/>reset-classified error" --> L["logRpcFailure — quieting the error<br/>and scheduling recovery are<br/>the same code path"]
    Q -. "no (idle): silent until the next<br/>failing call — the accepted gap" .-> L
    L --> H["coalesced, jittered reopen:<br/>500ms → 5s → … 60s cap"]
    H --> O["fresh openGadget()<br/>over the same socket"]
    O -- "new overseer stub" --> Z["chat / workpieces / console logs /<br/>actions / gadget pane all resubscribe"]
    O -. "open hangs >15s: settlement guard →<br/>'Reconnecting…' pill, retry" .-> H
```

**Workspace reopen.** `logRpcFailure` now doubles as the recovery entry point: any error it classifies as a DO reset schedules a reopen of the open workspace over the still-healthy socket.ke automatically when the stub is replaced, so it heals with the workspace instead of sticking on an error banner.

**Connection manager.** The socket-reconnect logic moved out of `main.tsx` globals into a tested `connectionManager.ts`. Reconnects are proven via a real round-trip (`getServerConfig`, which touches no DO) before the new stub is published, so the app never sees a half-alive connection. A tab-wake probe catches sockets that died while the tab slept, and each outage logs one aggregated lost/restored summary instead of a line per retry.

## Accepted gap (deliberate)

**No noise when a worspace is idle** -  Because there's no ping/health, recovery waits until the next call. Because recovery rides `logRpcFailure`, any failing call in the workspace is now enough (send, approve, hook toggle, subscribe…), but the wait is real: in fault injection, a workspace stayed dormant for minutes after a sustained abort loop until the next user action, which then healed it. A liveness heartbeat would close this; we left it out on purpose to keep the moving parts down.

### DIFF
--- a/packages/workshop-frontend/src/ChatInterface.tsx
+++ b/packages/workshop-frontend/src/ChatInterface.tsx
@@ -5237,7 +5237,7 @@ function ChatInterface({
           forceUpdate();
         }
       } catch (err) {
-        if (!logRpcFailure("Failed to subscribe to chats:", err)) {
+        if (!logRpcFailure("Failed to subscribe to chats:", err, { reportSite: 'chat.subscribe' })) {
           reportIssue('chat.subscription-load', err)
           toasts.add({ title: "Unable to load conversations", variant: "error" });
         }
@@ -5246,9 +5246,13 @@ function ChatInterface({
 
     subscribe();
 
-    // Set up reconnection handling
-    overseer.onRpcBroken?.((error) => {
-      console.warn("RPC connection broken:", error);
+    // Socket-level teardown only: onRpcBroken never fires for a DO reset behind a healthy
+    // session (workerd probe — see useWorkspaceOpen), so the subscribe catch above owns that
+    // recovery path via logRpcFailure's workspace recovery hook. No logging here: the
+    // connection manager already reports the outage once, and callbacks from every overseer
+    // generation fire together at teardown (capnweb has no unregister API), so a log line
+    // here multiplies per reopen.
+    overseer.onRpcBroken?.(() => {
       setIsSubscribed(false);
       // Cache persists, component will get new overseer prop and resubscribe
     });
--- a/packages/workshop-frontend/src/Connections.tsx
+++ b/packages/workshop-frontend/src/Connections.tsx
@@ -21,6 +21,7 @@ import {
   loadBindingCardData,
 } from './components/BlueprintBindingCard'
 import { reportIssue } from './errorReporting'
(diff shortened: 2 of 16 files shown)


pol-cc:
We're hitting exactly this on a production deployment, with field data that may help: for a **collaborator** (non-owner), `open()` additionally awaits `ensureAmbientCapsules` + `ensureObserver` (per-gatekeeper `getVerifier` + facet `addObserver`), so a collaborator open on a gatekeeper-heavy workspace takes ~4 s when healthy — vastly widening exposure to DO wedges. In one captured episode, `open` (canceled at 16.9 s), `startGatekeeperSession` (canceled at 16.3 s) and a gatekeeper-session `listFiles` (ok at 16.3 s) all unblocked at the same instant, i.e. queued behind one wedge; after release the same calls run in ms. Net effect: owners never see it, collaborators intermittently get an eternal "Loading workspace…" (no client timeout in `useWorkspaceOpen`) or the terminal error page, and nothing retries. The 15 s settlement guard + coalesced reopen in this PR would cover it; a server-side deadline on the collaborator observer chain might be worth considering too.

PR #125, opened 2026-08-10
store composer drafts in session storage until sent

maxwellpeterson:
This PR stores prompt drafts in session storage until they are sent to the agent. This means page refreshes don't wipe composer state, which helps mitigate workspace and conversation creation failures. Prompts typed into the new workspace, new conversation, and existing conversation composers are all covered. Capsules are serialized to their corresponding resource URL for storage, and format specifiers (docs, slides, etc.) are fully preserved.

### DIFF
--- a/packages/workshop-frontend/src/ChatInterface.tsx
+++ b/packages/workshop-frontend/src/ChatInterface.tsx
@@ -81,6 +81,7 @@ import {
   WorkpieceId,
   BlueprintOutput,
   MessageFormatRef,
+  OutputIcon,
   OutputFormatOffer,
 } from "@gadgets/workshop-shared/api";
 import { ActionKind, ResourceDescription } from "@gadgets/workshop-shared/gatekeeper";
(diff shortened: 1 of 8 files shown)


ndisidore (commented):
Are you sure you don't want localStorage? Session storage goes away when the tab is closed; we did localstorage in CFOSv1 so that you could recover from that.
Possibly overkill: I like the session storage api much more

Maximo-Guk:
GPT Findings
```
1. Medium: Chat changes reset text but retain attachments, URL overlays, and pending capsule operations. This can apply chat A’s resources to chat B. ChatInterface.tsx:1864, ChatInterface.tsx:1885, ChatInterface.tsx:1993
2. Medium: Restored inline slash commands lose command semantics because drafts store only text/formats and send-time recovery handles only leading /commands. composerDraft.ts:18, ChatInterface.tsx:2500
3. Medium: After an ambiguous transient send, reload restores the possibly-sent message but loses main’s in-memory warning, risking duplicate messages or workspaces. ChatInterface.tsx:2596
```

maxwellpeterson:
@ndisidore I want to start simple with session storage here. I personally wasn't a big fan of the cross-tab persistence in v1, and found that it led to prompt drafts popping up when I didn't actually want them.

Maximo-Guk (approved):

PR #127, opened 2026-08-10
Add server-managed model providers (SERVER_MODELS)

RyanHecht:
## Problem

Every user has to supply their own model credentials before they can use the OS. That fits a deployment whose users each have provider accounts, but not one where the operator wants to fund inference centrally — an internal LLM proxy, a corporate gateway, or a self-hosted model server. Today the operator's only options are to have everyone paste the same key, or to patch the source.

AI Gateway mode already solves this for Cloudflare-hosted inference, but it's limited to `SUGGESTED_MODELS` reached through `gateway.ai.cloudflare.com`.

## What this adds

`SERVER_MODELS`: a JSON array of models the deployment offers to every user, pointing at any endpoint the existing providers can speak to, with credentials that stay in the environment. Users select them like any other model, never see the token, and can't edit or delete them.

```json
[{"id": "house-sonnet", "name": "Sonnet (internal)", "provider": "anthropic",
  "model": "claude-sonnet-5", "apiUrl": "https://llm.internal.example.com",
  "apiToken": "sk-...", "contextWindow": 1000000}]
```

Docs: `docs/server-managed-models.md`.

## Design notes

- **No new inference path.** Server models resolve to the same `UserAiModelRecord` that user-configured models produce, so `getModel()` and everything downstream is unchanged.
- **Exempt from Gateway routing.** Routing a server model through an AI Gateway would discard the `apiUrl`/`apiToken` it was configured with, and the endpoint often isn't one a Gateway can reach. A `serverManaged` flag on `AiModelConfig` carries that exemption; `addModel()` rejects it from user input so it can't be used to bypass Gateway mode.
- **Independent of AI Gateway mode.** Both can be enabled; server models simply aren't routed through the Gateway.
- **`contextWindow` / `outputLimit`** are optional additions to `AiModelConfig`. A model with no `SUGGESTED_MODELS` entry otherwise falls back to `DEFAULT_CONTEXT_WINDOW`, so a large-context model would be compacted far earlier than necessary.
- **Fails loudly.** Malformed config throws, naming the offending entry (`SERVER_MODELS[1]: "model" is required.`), rather than silently offering no models.

## Testing

18 unit tests in `packages/workshop-backend/__tests__/server-models.test.ts` covering parsing, defaults, resolution, ordering, token limits and every validation failure. `pnpm run lint` and `pnpm run types:check` are clean.

Also exercised on a self-hosted deployment running the whole stack on Miniflare/workerd behind Cloudflare Access, with models served through a GitHub Copilot proxy.

## Open questions

- Should `SERVER_MODELS` entries be able to *override* a `SUGGESTED_MODELS` id (currently the Gateway entry wins)?
- Would you prefer this validated eagerly at worker startup rather than on first use? There didn't seem to be an obvious hook.

Happy to adjust naming or shape.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,9 @@ While you can code a Gadget by hand if you want, the expectation is that AI writ
 
 You can choose your LLM. Cloudflare OS works with many major AI model providers and self-hosted models, with more providers being added all the time.
 
+By default each user supplies their own model credentials. A deployment can instead offer models to
+everyone with credentials held server-side — see [server-managed models](docs/server-managed-models.md).
+
 Because of the tightly-integrated and simplified nature of the platform, even when using the same underlying AI models, the Cloudflare OS coding agent often performs better and faster with fewer tokens than a general-purpose coding agent would.
 
 ### Collaborate with AI
--- a/docs/server-managed-models.md
+++ b/docs/server-managed-models.md
@@ -1,0 +1,85 @@
+# Server-managed models
+
+By default every user brings their own model credentials: each person adds a provider and API key
+in **Settings → Providers** before they can chat. That is the right default for a deployment whose
+users have their own accounts, but not for one where the operator wants to fund inference centrally
+— an internal LLM proxy, a corporate gateway, or a self-hosted model server.
+
+`SERVER_MODELS` lets a deployment offer models to everyone, with credentials that stay in the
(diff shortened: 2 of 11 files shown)


RyanHecht:
Sorry, an overeager agent with too many permissions opened this. It is by no means ready :)

PR #133, opened 2026-08-10
Absorb user-DO resets in the Worker

ndisidore:
Routine Durable Object resets — storage timeouts, overload aborts; ~3k over a recent 4-day window in production logs — were wedging sessions: a stub is permanently broken once its incarnation resets, so a cached stub turned one reset into every subsequent call failing until reload, surfacing as scary terminal errors.

terminal error screenshot
terminal error screenshot

## What changed

**Fresh stub per call.** The session re-resolves its user-DO stub on every access, so a post-reset call simply restarts the object

**Ordering is now a client-side contract.** Per-call stubs forfeit workerd's per-stub e-order, so cross-call delivery order is no longer guaranteeded arrival would silently invert the durable state — now guard in-flight (clicks are ignored until the RPC settles).

**Deliberately no retries.** With fresh stubs, the only thing a reset can still break is the call in flight at that moment; workerd's structured reset flags reach the client, which #110 already classifies and quiets. 

## Deliberately deferred / out of scope

- **Subscription self-heal**: the DO holds subscriber registrations in memory only, so a reset still silently kills them while the browser's WebSocket stays health
- **Workspace (Overseer) DO resets** keep their pre-existing session-abort handling;

## Testing

- **Integration** (`abortAllDurableObjects()`, the non-graceful teardown): the same session recovers transparently after a reset via fresh stubs, and the local-aborts-reject-flagless shape is pinned so the flag path can graduate from synthetic unit tests to real-reset integration tests if a future pool upgrade attaches the production flags. (The suite also caught a real wrapper subtlety: native RPC method handles are proxies, so the wrapper must invoke through the stub rather than `.apply` on an extracted handle.)
- **Unit**: the reset-flag predicate against production-shaped synthetic errors. 284 backend unit + integration + 135 frontend green.
- **Manual fault injection** (chrome-devtools against a local dev stack, temporary `debugAbort()` injector, composed with the #110 client): run on the pre-split branch — 72 aborts across idle, reload-race, and SPA-navigation-storm scenarios with every page functional and zero application console output; a chat send racing a workspace-DO abort showed the inline hint (#110's UI), preserved the draft, and did not double-send.


Maximo-Guk (approved):

Maximo-Guk:
GPT comments:
```
1. Medium: do-retry.ts:19 retries errors with both overloaded and reset flags. Cloudflare explicitly says overloaded errors must not be retried; this can amplify an overload. Check overloaded before retrying.
2. Medium: server.ts:125-152 allows queries to overtake earlier commands because each uses a fresh stub and queries bypass #commandChain. This regresses RPC e-order/pipelining. Queries should await a snapshot of prior commands.
3. Low: server.ts:811,861 retries authentication resets without emitting the new retry/surfaced telemetry, undercounting resets.
```

packages/workshop-backend/src/do-reset.ts
```diff
@@ -0,0 +1,19 @@
...
+// whenever the object's incarnation died — the production storage-timeout reset arrives as
+// `{remote, overloaded, durableObjectReset}`. The flags are attached natively in the calling
+// Worker, so no message matching is needed. Local vitest-pool-workers aborts reject FLAGLESS
+// (pinned by the "user-DO reset flags" integration test), so this predicate is unit-tested
+// with synthetic production shapes.
+
+/** True for rejections caused by a DO reset or lost connection. Used to classify surfaced
+ * errors for telemetry (user_do.reset.surfaced); the Worker deliberately does not retry them
+ * (see the chokepoint comment in server.ts). `overloaded` alone is excluded — that object is
+ * alive and shedding load. */
```
kentonv:
```suggestion
// True for rejections caused by a DO reset or lost connection. These are requests that could make
// sense to retry (although as of this writing, the code does not do so). `overloaded` is excluded
// because when the DO is overloaded, retrying would make the problem worse.
```

packages/workshop-backend/src/server.ts
```diff
@@ -87,8 +101,72 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
   private users: DurableObjectNamespace<UserDurableObject>;
 
+  #userId: DurableObjectId;
+
+  // A stub is permanently poisoned once its incarnation resets, so re-resolve per call instead
+  // of caching one for the session (stub creation is local — not a network call). The trade:
+  // e-order is per stub, so cross-call delivery ordering is gone — a call site that depends on
+  // a prior user-DO call must await it, and UI that can fire overlapping writes to the same
+  // state must guard in-flight (see the BlueprintList pin and providers quick-model toggles).
+  // `#`-private because RpcTarget members are runtime-visible and TS `private` is erased.
```
kentonv:
```suggestion
  // Get a stub pointing at the user DO. We create a new stub for every request so that we don't
  // have to worry about detecting when a stub has become broken.
```

packages/workshop-backend/src/do-reset.ts
kentonv, line 1: Creating a whole file just for this one utility method seems egregious. Can we put this in utils file of some sort?

Or perhaps if we move all the proxy stuff into a separate file as suggested later, then that makes more sense as a stand-alone file. But we shouldn't be creating a new file for a 4-line declaration.

packages/workshop-backend/src/server.ts
```diff
@@ -87,8 +101,72 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
+    }
+  }
+
+  /** Builds the #user sugar: `this.#user.listGadgets()` routes through #userCall
+   * with the DO method name as the telemetry operation (so the logged operation names the DO
+   * method that was hit, which for a handful of endpoints differs from the API method —
+   * `dismissSharedGadget` logs as `forgetSharedGadget`). Sites whose closure does more than a
+   * single same-args delegation (multi-call closures, helpers that take the stub) use the
+   * chokepoints directly. */
+  #doProxy(dispatch: (operation: string,
```
kentonv: This proxy stuff is a lot of complexity. Is it worth it?

If we really need this, I think we need to encapsulate it better. We could put all this stuff in a separate file that exports one method, like `wrapDoStubForTelemetry(stub)`. Then the `#userStub` getter (which can be renamed to just `#user`, I think) applies the wrapper. Note that calling the same wrapper multiple times would reuse the same stub, not recreate it -- that's fine, because we will call the `#user` getter once per logical operation, and that creates the stub on-demand.

Then I think all the changes below can just change `this.user` to `this.#user` and that's it.

packages/workshop-backend/src/server.ts
```diff
@@ -300,67 +382,73 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
   provisionAmbientAccount(vendorId: string): Promise<void> {
-    return this.user.provisionAmbientAccount(vendorId);
+    return this.#user.provisionAmbientAccount(vendorId);
   }
 
   subscribeConnectedAccounts(
       subscriber: RpcStub<ConnectedAccountsSubscriber>, filter?: ConnectedAccountsFilter)
       : Promise<RpcStub<{}>> {
-    return this.user.subscribeConnectedAccounts(subscriber, filter);
+    // TODO(deferred): the DO holds this registration in memory only, so a reset silently kills
```
kentonv: This problem should be solved on the client side by detecting when the subscription callback is disposed and re-subscribing. This comment shouldn't be here since there's nothing in particular we need to do here.

packages/workshop-frontend/src/components/BlueprintList.tsx
```diff
@@ -203,7 +203,13 @@ export default function BlueprintList() {
     }
   }, [authenticatedApi, load, toasts])
 
+  // Overlapping setBlueprintPinned calls have no ordering guarantee and could invert the
```
kentonv: This and the change in providers.tsx feel pendantic. I don't care that much since it's UI code but IMO the agent has gone overboard trying to solve a problem that doesn't matter.

packages/workshop-backend/src/server.ts
ndisidore: done :heavy_check_mark:

packages/workshop-backend/src/server.ts
ndisidore: comment removed

packages/workshop-backend/src/server.ts
```diff
@@ -286,11 +298,13 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
 
   async listGadgets(): Promise<GadgetMetadataWithTimestamps[]> {
-    return this.user.listGadgets();
+    return this.#user.listGadgets();
   }
 
   listOutputs(): Promise<ListOutputsResult> {
-    return this.user.listOutputs();
+    // The backfill inside is safe under concurrency: the DO sweeps one page per call and
+    // advances the cursor itself.
```
kentonv:
```suggestion
```

packages/workshop-backend/src/server.ts
```diff
@@ -107,56 +117,58 @@ class AuthenticatedApiImpl extends RpcTarget implements AuthenticatedApi {
...
   }
   completeOnboarding(): Promise<void> {
-    return this.user.completeOnboarding();
+    return this.#user.completeOnboarding();
   }
 
   getCloudflareUsage(): Promise<CloudflareUsageInfo> {
-    return getUsageInfo(this.env, this.user);
+    // Cache write-backs inside (account selection / credit snapshots) tolerate racing; a reset
+    // surfaces to the usage panel's fallback.
```
kentonv:
```suggestion
```

kentonv (approved):

PR #135, opened 2026-08-10
Fix code that aborts the WebSocket when an overseer DO dies.

kentonv:
b2a215e had switched the overseer death handling from a simple call to `ctx.abort()` (an experimental API) to a more elaborate mechanism that closes the WebSocket.

However, the code closed the wrong end of the WebSocket, and so it didn't work.

This change introduces a cleaner approach that actually shuts down the RPC session by disposing the main stub. However, this requried cloning some functions from Cap'n Web. Arguably, this feature should be folded back into Cap'n Web. However, at the same time, `ctx.abort()` will become generally available within a few days, and once it does we might as well switch back to it.

This PR also includes a couple of UI tweaks that make reconnecting more pleasant.


kentonv:

ndisidore (approved):
smart play patching in the `abortSignal` 👍

PR #140, opened 2026-08-11
Don't submit chat input on Enter during IME composition

chaspy:
## Problem

When typing with an IME (Japanese, Chinese, Korean, etc.), pressing Enter to confirm a text conversion sends the chat message instead of just committing the composition. This makes the chat input nearly unusable for CJK users.

## Fix

Check `KeyboardEvent.isComposing` before treating Enter as submit, in three places in `ChatInterface.tsx`:

- chat message send
- slash-command picker selection
- chat rename input

6 insertions, 4 deletions. Verified locally with a Japanese IME: Enter during conversion now only commits the composition; Enter after conversion sends as before.


chaspy:

maxwellpeterson:

PR #144, opened 2026-08-11
Add pull request template

dcartertwo:
## Summary

- surface the existing contribution policy before a pull request is submitted
- ask authors to explain the concrete fix and why it is trivially verifiable
- add checklist confirmations with stable hidden identifiers for future policy automation

## Verification

- `git diff --check`

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,23 +1,25 @@
 > [!IMPORTANT]
-> We are not generally accepting outside contributions. We only accept small,
-> trivially verified pull requests that fix a concrete problem. Please read our
-> [contribution guidelines](https://github.com/cloudflare/cloudflare-os/blob/main/CONTRIBUTING.md)
-> before submitting. For larger changes, please start a
-> [discussion](https://github.com/cloudflare/cloudflare-os/discussions) instead.
+> We are not generally accepting outside contributions. We may accept a small
+> change only when, in the maintainers' assessment, it is obviously correct and
+> trivially verifiable by reading the patch.
+>
+> If you found a bug, please [file an issue](https://github.com/cloudflare/cloudflare-os/issues/new)
+> first. For feature requests and other proposals, please
+> [open a discussion](https://github.com/cloudflare/cloudflare-os/discussions).
 
-## What does this fix?
+## What does this change?
 
-<!-- Describe the concrete problem and how this patch fixes it. -->
+<!-- Describe the concrete problem and how this patch addresses it. -->
 
-## Why is this trivially verifiable?
+## Why is this obviously correct and trivially verifiable?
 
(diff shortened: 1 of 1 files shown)


dcartertwo:

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
+> [!IMPORTANT]
+> We are not generally accepting outside contributions. We only accept small,
+> trivially verified pull requests that fix a concrete problem. Please read our
+> [contribution guidelines](https://github.com/cloudflare/cloudflare-os/blob/main/CONTRIBUTING.md)
+> before submitting. For larger changes, please start a
+> [discussion](https://github.com/cloudflare/cloudflare-os/discussions) instead.
```
kentonv: Let's separate:
* For bugs, file an issue.
* For feature requests, open a discussion.

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+
+<!-- Explain how a reviewer can establish correctness by reading the patch. -->
+
+## Checklist
+
+If you cannot check every item, please close this pull request and start a
+discussion instead. Pull requests that do not meet these criteria will be closed.
+
+- [ ] <!-- contribution-policy:concrete-fix --> This fixes a concrete problem; it is not a feature, refactor, or low-value cleanup such as a typo fix.
+- [ ] <!-- contribution-policy:small-patch --> This patch is about a dozen changed lines and can be fully verified by reading it.
```
kentonv: I wonder if we can also capture whether it is "obviously correct"... There have been some PRs that were only a few lines but the effects were subtle, e.g. disabling a feature for ollama so that the adapter worked with arbitrary OpenAI-compatible endpoints. It's one line but the effects are hard to judge.

Maybe we should actually remove this triviality condition entirely and just say we don't accept PRs, otherwise people might get annoyed when they open a one-liner and we decide it isn't sufficiently trivial.

dcartertwo: how about

> We are not generally accepting outside contributions. We may accept a small
> change only when, in the maintainers' assessment, it is obviously correct and
> trivially verifiable by reading the patch.

i think proactively sharing that it is in the "maintainers' assessment" is reasonable and not that uncommon.

dcartertwo: I kept the narrow exception, but clarified that acceptance depends on the maintainers' assessment that the change is both obviously correct and trivially verifiable. I also removed the line-count language so a small diff is not presented as sufficient.

.github/pull_request_template.md
dcartertwo: Updated the template to direct bug reports to Issues and feature requests or other proposals to Discussions.

kentonv (approved):

PR #145, opened 2026-08-11
Trace gatekeeper session resolution and stamp identity on loopback invocations

ndisidore:
Looking at production trace data, nearly every retained `GatekeeperLoopback.jsrpc` record was a "code had hung" cancellation, arriving in bursts tied to agent-initiated gatekeeper calls. The canceled invocations carry no vendor or caller identity, and the traces don't show which stage stalled, so the investigation had to infer both from sibling spans.

This adds three small instruments to workshop-backend so the next investigation can read the answer straight off the trace:

- Every loopback invocation stamps a `gatekeeper.loopback.identity` span with vendorId, target, callerFrom, and chatId. Canceled invocations become directly groupable by vendor.
- A `gatekeeper.loopback.resolve` span times session resolution. It uses a new `createPipelinedTracer` helper that returns the RpcPromise unchanged, since the existing `traced` helper would break capnweb pipelining here.
- A `gatekeeper.session.open` span (plus a failure-only log event) wraps `openSession` in the overseer DO. Comparing it against the loopback spans localizes a hang to overseer dispatch, connector session open, or the downstream connector call.

Connector-side instrumentation of the tool call itself is deferred to a follow-up.

### DIFF
--- a/packages/backend-utils/__tests__/tracing.test.ts
+++ b/packages/backend-utils/__tests__/tracing.test.ts
@@ -127,6 +127,9 @@ describe("tracedPipelined", () => {
 // the stamp becomes redundant.
 describe("unended span visibility in tail events", () => {
   it("streams the open event but not the attributes of a span that never ends", { timeout: 15_000 }, async () => {
+    // Start fresh so marker strings can be reused by future tests without flaking.
+    await probeEnv.SPAN_RECORDER.clear();
+
     // Control: an ended span reaches the recorder with its attributes.
     tracedPipelined("pin-ended-span", (span) => {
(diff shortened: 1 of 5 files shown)


kentonv:
I'm not necessarily against adding more trace info, but if we're specifically chasing the "Worker's code had hung and would never generate a response." errors, I think it's not worth it. I'm pretty confident this is a runtime bug that has no visible impact: the error is being reported spuriously at the time the event was about to end anyway. I would guess that this error is actually being produced commonly in local testing, even, but it doesn't get reported anywhere since there's no tail worker installed. We should try installing one and see what happens, then if we can reproduce, report it to the runtime team.

PR #148, opened 2026-08-11
fix(frontend): ignore stale blueprint loads after route changes

### LINKED ISSUE #146: Stale blueprint request can overwrite a newer route and target actions at the wrong blueprint
## Suggested severity

High (S2): an ordinary client-side route change can leave the page showing one blueprint while its actions operate on the blueprint ID in the current URL.

## Summary

`BlueprintLandingPage` does not associate an asynchronous `getBlueprint(id)` result with the route that started it. TanStack Router reuses the component when only the `$id` parameter changes, so an older request can settle after a newer request and replace the current blueprint state.

A controlled-promise test reproduces this deterministically on unmodified commit `8b08672050e6957987a87b5f9cab2801ba89b42c`.

## Deterministic reproduction on unpatched `main`

1. Render `/blueprint/blueprint-a` and keep `getBlueprint("blueprint-a")` pending.
2. Re-render the same route component for `/blueprint/blueprint-b`.
3. Resolve B with metadata titled `CURRENT BLUEPRINT`.
4. Resolve the older A request with metadata titled `STALE BLUEPRINT`.
5. Inspect the current route and rendered page.

### Actual result

The route remains `/blueprint/blueprint-b`, but A's title and metadata overwrite the page. A previously loaded blueprint can also appear during the first render of the next route. If B rejects after A has loaded, A remains visible instead of B's error.

### Expected result

Only the request associated with the current `id` and RPC stub should update the page. The previous blueprint should stop rendering as soon as the route changes.

## Impact

This mismatch is more than cosmetic. Event handlers use the current route `id`, while labels, confirmation text, binding metadata, and download filenames use the `blueprint` object in state. For example, a delete confirmation can name A while `deleteBlueprint(id)` targets B, and a download can use A's filename while requesting B's archive.

## Root cause

The metadata-loading effect has no cleanup or generation guard. Every `.then()`, `.catch()`, and `.finally()` callback can update shared state after its route is stale. Loaded state is also tied only to the reused component, rather than to the `(id, rpcStub)` load that produced it.

## Proposed fix

- Cancel state updates from an effect when its `id` or `rpcStub` is superseded.
- Track which `(id, rpcStub)` produced the loaded blueprint, preventing old content from rendering during the route-change commit before the next effect runs.
- Clear the previous blueprint and error when starting the next load.
- Cover out-of-order responses, the first render after navigation, not-found responses, and rejected requests.

ai-yang:
Fixes #146

## Summary

- Ignore `getBlueprint()` callbacks after their route ID or RPC stub has been superseded.
- Tie rendered blueprint data to the `(id, rpcStub)` that loaded it, so the old blueprint cannot flash during the first render of a new route.
- Clear loaded/error state at the start of each load.

Without these guards, an older request could settle after the current request and overwrite the page. The URL and action handlers would then target the new blueprint ID while the UI displayed metadata from the old blueprint.

## Regression coverage

The new component tests use controlled promises to verify that:

- an A response resolving after B cannot replace B;
- the first render after changing A to B shows loading rather than A;
- a not-found B response does not leave A visible; and
- a rejected B request shows B's error without leaving A visible.

## Test plan

- [x] `pnpm --filter @gadgets/workshop-frontend test` — 28 test files / 140 tests passed
- [x] `pnpm --filter @gadgets/workshop-frontend types:check`
- [x] `pnpm exec oxlint packages/workshop-frontend/src/BlueprintLandingPage.tsx packages/workshop-frontend/src/BlueprintLandingPage.test.tsx`
- [x] `git diff --check`

### DIFF
--- a/packages/workshop-frontend/src/BlueprintLandingPage.test.tsx
+++ b/packages/workshop-frontend/src/BlueprintLandingPage.test.tsx
@@ -1,0 +1,238 @@
+// @vitest-environment jsdom
+/* eslint-disable react/react-in-jsx-scope */
+
+import { act, type ComponentProps, type ReactElement, type ReactNode } from 'react'
+import { flushSync } from 'react-dom'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterAll, afterEach, describe, expect, it, vi } from 'vitest'
+import type { RpcStub } from 'capnweb'
+import type { BlueprintPublicInfo, PublicApi } from '@gadgets/workshop-shared/api'
+
+const route = vi.hoisted(() => ({ id: 'blueprint-a' }))
+
+vi.mock('@tanstack/react-router', () => ({
+  useParams: () => ({ id: route.id }),
+  useNavigate: () => vi.fn<() => void>(),
+  useRouter: () => ({ history: { canGoBack: () => false, back: vi.fn<() => void>() } }),
+}))
+
+vi.mock('./useAuth', () => ({
+  useAuth: () => ({
+    isAuthenticated: false,
+    authenticatedApi: null,
+    isLoading: false,
+    login: vi.fn<() => void>(),
+  }),
+}))
+
+vi.mock('@cloudflare/kumo', () => {
+  // oxlint-disable-next-line unicorn/consistent-function-scoping
+  const PassThrough = ({ children }: { children?: ReactNode }) => <>{children}</>
+  const Dialog = Object.assign(
+    ({ children }: { children?: ReactNode }) => <div>{children}</div>,
+    {
+      Root: ({ children, open }: { children?: ReactNode; open?: boolean }) =>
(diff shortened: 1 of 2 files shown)


ai-yang:

ai-yang:
Hi maintainers — this PR was opened before the new contribution-policy workflow was added.

The production change is 23 changed lines; the deterministic regression test accounts for the remaining 238 lines. I rechecked current `main`, and the stale blueprint request race is still present.

Would you prefer to keep the regression test under a `policy/override` label, or should I revise this into a production-only patch below the 30-line automatic limit? Once you confirm the preferred scope, I can rebase onto current `main`, update the PR description to the current template, and rerun validation.

If the current scope is acceptable, could you also approve the pending CI and Bonk PR Review workflows when convenient?

Thank you.

PR #149, opened 2026-08-11
fix(blueprints): preserve bindings on code-only updates

### LINKED ISSUE #147: Code-only blueprint updates unexpectedly rewrite binding metadata
## Suggested severity

High (S2): the dedicated code-only update path can silently change a published blueprint's connection contract, or fail for an unrelated binding-annotation error.

## Summary

The `Overseer.updateBlueprint()` API documents `updateCode` and `updateBindings` as independent options. However, calling `updateBlueprint(id, { updateCode: true })` currently calls `collectBindingMetadata()` and assigns its result to `record.metadata.bindings`.

The Workshop's **Update code** action invokes exactly this code-only option, so the normal UI workflow is affected.

## Deterministic reproduction on unpatched `main`

Using commit `8b08672050e6957987a87b5f9cab2801ba89b42c`:

1. Create a blueprint record whose stored binding metadata is `OLD_API`.
2. Make the source gadget's current binding metadata return `NEW_API`.
3. Call `updateBlueprint("blueprint-1", { updateCode: true })`.
4. Inspect `record.metadata.bindings` and the `collectBindingMetadata()` call count.

### Actual result

`collectBindingMetadata()` is called once and the stored metadata changes from `OLD_API` to `NEW_API`, even though `updateBindings` was not requested.

If collection throws because a current binding annotation is incomplete, the code-only update rejects before snapshotting the code. The controlled regression test failed on unpatched `main` with `promise rejected "Error: invalid current binding annotation" instead of resolving`.

### Expected result

A code-only update should update the code snapshot/version and preserve the blueprint's existing binding metadata. Binding metadata should be read and replaced only when `updateBindings: true` is present.

## Impact

- A user choosing **Update code** can unintentionally change the inputs/connections required by an already published blueprint.
- An invalid current binding annotation can block an otherwise valid code-only update.
- The behavior contradicts the public RPC contract, which assigns binding refresh to the independent `updateBindings` option.

## Root cause

`updateBlueprint()` enters one shared `if (options.updateCode || options.updateBindings)` block and unconditionally collects and assigns bindings before checking `options.updateCode`.

## Proposed fix

Keep the shared gadget-ID resolution, but guard collection and assignment with `if (options.updateBindings)`. Cover code-only, bindings-only, and combined updates.

ai-yang:
Fixes #147

## Summary

Make `Overseer.updateBlueprint()` refresh binding metadata only when callers explicitly pass `updateBindings: true`.

The API already treats `updateCode` and `updateBindings` as independent flags, and the Workshop's **Update code** action sends only `{ updateCode: true }`. Previously, that call still collected and replaced binding metadata, which could silently change the published blueprint contract or reject on an unrelated invalid annotation.

The production change is intentionally limited to guarding the existing binding collection. Code-only, bindings-only, and combined updates otherwise retain their existing propagation behavior.

## Regression coverage

The new tests verify that:

- a code-only update never calls `collectBindingMetadata()`, preserves existing bindings, and still succeeds when collection would throw;
- a bindings-only update refreshes bindings without taking a code snapshot; and
- a combined update refreshes both code and bindings in one propagation pass.

## Test plan

- [x] `pnpm --filter @gadgets/workshop-backend exec vitest run __tests__/blueprint-update.test.ts` — 1 test file / 3 tests passed
- [x] `pnpm --filter @gadgets/workshop-backend test` — 26 unit-test files / 282 tests passed; 4 environment-dependent integration tests skipped by the existing configuration
- [x] `pnpm --filter @gadgets/workshop-backend types:check`
- [x] `pnpm exec oxlint packages/workshop-backend/src/overseer.ts packages/workshop-backend/__tests__/blueprint-update.test.ts` — exit 0; pre-existing warnings in `overseer.ts` are unchanged
- [x] `git diff --check`

### DIFF
--- a/packages/workshop-backend/__tests__/blueprint-update.test.ts
+++ b/packages/workshop-backend/__tests__/blueprint-update.test.ts
@@ -1,0 +1,151 @@
+import { RpcStub as NativeRpcStub } from "cloudflare:workers";
+import { describe, expect, it, vi } from "vitest";
+import { OverseerDurableObject } from "../src/overseer.js";
+
+vi.mock("capnweb-validate", () => ({ validateRpc: () => () => undefined }));
+
+const originalBindings = {
+  OLD_API: {
+    type: "gatekeeper" as const,
+    title: "Original API",
+    gatekeeperName: "original",
+    typeUrlPattern: "https://old.example/*",
+  },
+};
+
+const currentBindings = {
+  NEW_API: {
+    type: "gatekeeper" as const,
+    title: "Current API",
+    gatekeeperName: "current",
+    typeUrlPattern: "https://new.example/*",
+  },
+};
+
+async function makeClient(options: { bindingError?: Error } = {}) {
+  let record = {
+    id: "blueprint-1",
+    gadgetId: 7,
+    codeVersion: 3,
+    metadata: {
+      title: "Demo",
+      description: "Demo blueprint",
+      author: { type: "user" as const, id: "owner", name: "Owner" },
+      created: new Date("2026-01-01"),
+      lastUpdated: new Date("2026-01-01"),
+      version: 1,
+      bindings: structuredClone(originalBindings),
+    },
+  };
+  let collectBindingMetadata = vi.fn(() => {
+    if (options.bindingError) throw options.bindingError;
+    return structuredClone(currentBindings);
+  });
+  let snapshot = new Uint8Array([1, 2, 3]);
+  let snapshotCode = vi.fn(async () => snapshot);
+  let propagateBlueprint = vi.fn(async () => {});
+  let owner = {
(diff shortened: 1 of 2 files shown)


ai-yang:

ai-yang:
Hi maintainers — this PR was opened before the new contribution-policy workflow was added.

The production change is 8 changed lines; the deterministic regression test accounts for the remaining 151 lines. I rechecked current `main`, and code-only blueprint updates still refresh binding metadata unconditionally.

Would you prefer to keep the regression test under a `policy/override` label, or should I revise this into a production-only patch below the 30-line automatic limit? Once you confirm the preferred scope, I can rebase onto current `main`, update the PR description to the current template, and rerun validation.

If the current scope is acceptable, could you also approve the pending CI and Bonk PR Review workflows when convenient?

Thank you.

PR #150, opened 2026-08-11
Add Bonk break-glass command

ndisidore:
Allow authorized collaborators to skip Bonk by commenting `bonk break glass`.

### DIFF
--- a/.github/workflows/bonk-pr.yml
+++ b/.github/workflows/bonk-pr.yml
@@ -7,6 +7,10 @@ on:
     types: [created]
 
 jobs:
+  # Break-glass invariants:
+  # - The main ruleset requires `bonk` from GitHub Actions app 15368.
+  # - Labeled PRs run a successful step so future head SHAs also pass `bonk`.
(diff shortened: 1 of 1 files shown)


ndisidore:
@ask-bonk
1. Invalid. The active ruleset requires exactly bonk from GitHub Actions app ID 15368. The normal job produces that exact name/app, and the synthetic check will too.
2. Invalid. Cancellation is followed by polling, then the matching bonk check is created. The canceled result cannot arrive afterward and supersede it.
3. Valid policy concern. Break-glass authority includes changes to this workflow itself. That is inherent to the feature.
4. Invalid. After a push, the bonk job runs but skips only the Bonk steps; the job concludes successfully. The direct check handles the current SHA, while the label handles future SHAs.
5. Invalid. Removing the label creates a newer real bonk check with the same app/name, which governs just like a normal workflow rerun.
6. Partly valid. Bot-triggered synchronization must produce a successful bypass check while the label remains. The review incorrectly references a labeled rerun; that trigger was removed.

Maximo-Guk (approved):

PR #161, opened 2026-08-11
Automatically reject clear contribution-policy violations

dcartertwo:
The contribution policy has both objective requirements and a subjective final judgment. This automates only the objective rejection cases: incomplete author confirmations and patches over 30 changed lines. It deliberately does not decide whether a remaining PR is obviously correct or trivially verifiable; that stays with maintainers.

Because closing fork PRs requires a write-capable workflow, it executes only repository-owned code and never checks out the contributor’s branch.

Depends on #144.

### DIFF
--- a/.github/workflows/contribution-policy.yml
+++ b/.github/workflows/contribution-policy.yml
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   enforce:
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request.state == 'open' && github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     timeout-minutes: 5
     permissions:
--- a/.github/workflows/contribution-policy.yml
+++ b/.github/workflows/contribution-policy.yml
@@ -1,0 +1,37 @@
+name: Contribution policy
+
+on:
+  pull_request_target:
+    types: [opened, reopened, edited, synchronize, ready_for_review, converted_to_draft, labeled, unlabeled]
+
+permissions: {}
+
+concurrency:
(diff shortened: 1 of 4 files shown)


Maximo-Guk (approved):

.github/workflows/contribution-policy.yml
```diff
@@ -0,0 +1,37 @@
...
+
+permissions: {}
+
+concurrency:
+  group: contribution-policy-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  enforce:
+    if: github.event.pull_request.draft == false
```
Maximo-Guk: Is it intentional that we won't enforce this for draft PR's?

.github/workflows/contribution-policy.yml
```diff
@@ -0,0 +1,37 @@
+name: Contribution policy
+
+on:
+  pull_request_target:
+    types: [opened, reopened, edited, synchronize, ready_for_review, converted_to_draft, labeled, unlabeled]
```
Maximo-Guk, line 5: From GPT:
```
.github/workflows/contribution-policy.yml:5: edited, labeled, and unlabeled also fire for closed or merged PRs. Without checking state === "open", the bot can post a false rejection comment on an already merged PR.
```

.github/workflows/contribution-policy.yml
dcartertwo: yes

.github/workflows/contribution-policy.yml
dcartertwo: fixed this

PR #162, opened 2026-08-11
Absorb user-DO resets in the Overseer's session capabilities

ndisidore:
When a user DO resets under an open workspace, the session stays broken until the browser reconnects. Chat sends fail, the model picker and gatekeeper setup wedge until a reload, and pin/title updates stop propagating. The cause is the same one #133 fixed at the Worker layer: the Overseer minted its user-DO stubs once at `open()` and reused them for the whole session, and a stub is permanently broken once its incarnation dies.

### Changes

The session capability classes now store the user ID and create a fresh stub for each call, through the same telemetry wrapper the Worker uses.

### Deliberately out of scope

No retries, matching #133. No call deadline either: a user DO wedged behind a stuck input gate hangs `newChat` silently for an unbounded-but-usually-finite window (i.e. the 30s DO timeout), then rejects (and now recovers). Will add as a followup if we think its necessary

### DIFF
--- a/packages/workshop-backend/src/overseer.ts
+++ b/packages/workshop-backend/src/overseer.ts
@@ -38,6 +38,7 @@ import { SharingManager, SharingCaller, CollaboratorRecord, ShareKeyRecord } fro
 import { AutoApprovalDrainer } from "./auto-approval";
 import { collectSlashCommands, invokeSlashCommand } from "./slash-commands";
 import { createWorkshopLogger, obsContext, traced } from "./observability";
+import { wrapDoStubForTelemetry } from "./do-telemetry";
 import type { ChatGatewayRpcTarget, SubmitExternalMessageResult } from "@gadgets/workshop-shared/external-message-gateway";
 import {
   assertChatAttachmentSupportedByProvider,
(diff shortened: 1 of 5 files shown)


packages/workshop-backend/src/overseer.ts
```diff
@@ -8195,18 +8208,18 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
   async newChat(initialMessage: string | SlashCommandRequest, chosenModelId: string | null,
                 capsules?: CapsuleSpecifier[], attachments?: ChatAttachmentHandle[],
                 formats?: MessageFormatRef[]): Promise<number> {
-    let userMeta = await this.clientUser.getChatContext(chosenModelId);
-    return this.impl.newChat(this.clientUser, userMeta, initialMessage, capsules, attachments,
+    let userMeta = await this.#clientUser.getChatContext(chosenModelId);
+    return this.impl.newChat(this.#clientUser, userMeta, initialMessage, capsules, attachments,
```
ndisidore, line 8213: Considered changing impl.newChat/sendChatMessage to take a `clientUserId: string` instead of the stub. Decided against it for this PR: the stub is essentially request scoped anyway so it adds more churn for not much value

kentonv (approved):

Maximo-Guk (approved):

PR #163, opened 2026-08-12
Drop the dead `dist` emit and the duplicated type check

Maximo-Guk:
This PR implements a few build hygiene improvements for CI + local dev:

- Cache the pnpm store, drop `types:check` from the `lint` job so the type check runs once. The two jobs stay parallel, so a lint failure and a test failure surface on the same run.
- Drop the dead `dist` emit - only `typed-storage` has `exports` resolving to `dist`. Everything else is `noEmit`.
- Codegen scripts compare content before writing, so repeat runs don't reset mtimes.
- Toolchain versions move into a `pnpm-workspace.yaml` catalog instead of being pinned in package.json files.
- Delete `.gitlab-ci.yml`: this repo is now on GitHub Actions


packages/backend-utils/package.json
```diff
@@ -23,17 +23,16 @@
...
   },
   "dependencies": {
     "@gadgets/error-reporting": "workspace:*"
   },
   "devDependencies": {
-    "@cloudflare/vitest-pool-workers": "^0.18.8",
+    "@cloudflare/vitest-pool-workers": "catalog:",
     "@cloudflare/workers-types": "^5.20260722.1",
-    "vitest": "^4.1.10"
+    "vitest": "catalog:"
```
Maximo-Guk, line 36: See https://pnpm.io/catalogs

"Catalogs" are a [workspace feature](https://pnpm.io/workspaces) for defining dependency version ranges as reusable constants. Constants defined in catalogs can later be referenced in package.json files.

See our pnpm-workspace.yaml https://github.com/cloudflare/cloudflare-os/pull/163/changes#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794cR10

packages/gatekeeper-context/tsconfig.json
```diff
@@ -6,10 +6,8 @@
     "module": "ESNext",
     "moduleResolution": "bundler",
     "types": ["./worker-configuration.d.ts", "node"],
-    "outDir": "./dist",
-    "declaration": true,
```
Maximo-Guk, line 10: - Gatekeeper-context package.json declares main: "./src/index.ts" - the only entry point is source. No exports map, and no types field pointing at dist.
  - No package in the repo depends on @gadgets/gatekeeper-context at all, and nothing anywhere imports gatekeeper-context/dist.
  - Its wrangler.jsonc has main: ".wrangler/validate/src/index.ts" - wrangler bundles from source, so the deploy path never touched dist either.

From my agent

```
declaration: true was minting .d.ts files no resolver could reach, and sourceMap: true was mapping JS nothing loaded. 

It was worse than idle: because rootDir was inferred rather than set, cross-package source imports pulled siblings into the program, 

so gatekeeper-context/dist actually contained {gatekeeper-context,typed-storage,workshop-shared} subtrees — it was emitting copies of its neighbours' output.

The rule was applied mechanically from each package.json, not by hand, which is why typed-storage is the one package that kept all of it:
```

.github/workflows/ci.yml
```diff
@@ -22,19 +25,21 @@ jobs:
         with:
           persist-credentials: false
 
+      # Ahead of setup-node, which shells out to `pnpm store path` to find the directory it caches.
```
ndisidore, line 26: nit: lots of extra comments in here that are not super necessary

packages/backend-utils/package.json
ndisidore: Love this! ❤️ 
Nice way to get consistency

packages/backend-utils/tsconfig.json
```diff
@@ -5,12 +5,8 @@
     "lib": ["ES2022"],
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "types": ["@cloudflare/workers-types/experimental", "./src/node-async-hooks.d.ts"],
-    "outDir": "./dist",
-    "declaration": true,
-    "sourceMap": true,
```
ndisidore, line 11: We may still want sourcemaps i.e. for observability upload
https://developers.cloudflare.com/workers/observability/source-maps/

packages/error-reporting/package.json
```diff
@@ -10,11 +10,11 @@
   },
   "scripts": {
     "build": "tsc",
-    "types:check": "tsc --noEmit",
```
ndisidore, line 13: is removing this intentional? it seems like build and typecheck may now be the same?

pnpm-workspace.yaml
```diff
@@ -1,9 +1,20 @@
 packages:
   - packages/*
 
+# The toolchain versions every package shares. Declared once here so a bump is one edit rather than
+# needing to hand edit a bunch of different package.json files.
+catalog:
```
ndisidore, line 6: :nice:

ndisidore (commented):
Awesome cleanup initiative!

packages/backend-utils/tsconfig.json
Maximo-Guk: The source maps that observability upload uses come from wrangler's own bundling at deploy time (via upload_source_maps = true), which bundles from TS source directly

.github/workflows/ci.yml
Maximo-Guk: Fixed!

packages/error-reporting/package.json
Maximo-Guk: The `types:check` in our root package.json is now an alias for pnpm build

Before this change we were actually running tsc twice per package (the lint job ran pnpm lint -> recursive types:check and the test job ran pnpm build). Now typechecking is done only in the build step of the "Build and test" job and the Lint job runs oxlint only

ndisidore (approved):

PR #164, opened 2026-08-12
Run oxlint through Vite+

Maximo-Guk:
Ports `.oxlintrc.json` into the `lint` block of our root `vite.config.ts`, and switches `build`/`test`/`clean` from `pnpm run --recursive` to `vp run -r`, which caches each package's task against its inputs.

The root `oxlint` devDependency goes away with the config file: nothing invokes oxlint directly once both scripts route through `vp lint`, and `vite-plus` already depends on it at an exact `=1.76.0`

`pnpm build`, on my machine, medians of repeated runs:

| | before | after |
|---|---|---|
| cold, no cache | 35s | **22s** |
| warm cache | 33s | 13s |

Look at those cache savings!
image

Note: No tsconfig sets `incremental`. This is deliberate. `tsc` reads its own `.tsbuildinfo` and writes it back, and Vite+ won't cache a task that reads a path it also writes. Measured with it on, the whole type check drops out of the task cache - 2% hits and 21.1s, vs 65% and 13.4s without `incremental` enabled.

Also drops `baseUrl` from all tsconfigs. It wasn't doing anything and was one of the blockers to ts7, every `paths` entry here is an explicit relative path, and `tsc` has resolved `paths` against the tsconfig's own directory since 4.1. Also three configs had it with no `paths` at all anyways. We're still blocked from migrating to ts7 until we resolve the capnweb recursive Stub/Stubify issue, but this at least sets us up better

Stacked on #163.

Thanks to #33 by @ubugeeei for the performance suggestions


vite.config.ts
```diff
@@ -0,0 +1,152 @@
...
+    options: {
+      // Note: type-aware linting is intentionally not enabled. The type-aware engine
+      // uses tsgo (TypeScript 7), and three packages do not type-check under it:
+      // workshop-frontend, integration-tests and workshop-backend blow past tsgo's
+      // instantiation depth on capnweb's recursive Stub/Stubify types (TS2321/TS2589).
+      // Full type safety is still enforced by `tsc` via the `build` script.
+      typeAware: false,
+    },
+    env: {
+      es2024: true,
```
ndisidore, line 27: just noting: i don't remember seeing es2024 in the codebase (seen es2022, es2023, esNext)
do we know where we pulled this value?

ndisidore (approved):

vite.config.ts
Maximo-Guk: This came over from our `.oxlintrc.json` https://github.com/cloudflare/cloudflare-os/pull/164/changes#diff-2113aef2ff3d50c021fca1fb6ecf4d4553a0e3f5f51ded7302ad499e43bc2b1eL77

PR #167, opened 2026-08-12
Add OpenAI-compatible model provider

ndisidore:
Adds beta support for user-configured OpenAI-compatible Chat Completions providers (`openai-compatible` in config, **OpenAI-compatible** in the UI).

This first version is intentionally minimal:

- Direct mode only; it bypasses user unified billing and is unavailable in platform AI Gateway mode.
- Requires a model ID, bearer token, HTTPS base URL, context window, and output limit.
- Rejects URL credentials, query strings, fragments, and invalid token limits. Pasted `/chat/completions` endpoints are normalized automatically.
- Supports text only. Historical images become omitted-attachment markers.
- Persists a `conservative` compatibility profile, using `system` and `max_tokens` while disabling optional features generic endpoints may not support. We can look at expanding this in the future

Should unlock the following services:
- OpenRouter
- Groq
- DeepInfra
- DeepSeek
- Mistral AI
- Moonshot AI / Kimi
- Hugging Face Inference Providers
- HTTPS-hosted vLLM, LiteLLM, llama.cpp, and similar self-hosted gateways
- And a bunch more

## Screenshots
Screenshot from 2026-08-12 08-56-28
Screenshot from 2026-08-12 09-05-30
Screenshot from 2026-08-12 09-05-44
Screenshot from 2026-08-12 09-06-17
Screenshot from 2026-08-12 09-07-37
Screenshot from 2026-08-12 09-08-27


packages/workshop-shared/src/api.ts
```diff
@@ -943,18 +943,19 @@ export type CloudflareAccountOption = {
...
 };
 
-// Supported AI providers.
-export type AiModelProvider = "openai" | "anthropic" | "google" | "cloudflare" | "ollama";
+/** Supported AI providers. "openai-compatible" names the Chat Completions protocol, not a vendor. */
+export type AiModelProvider =
+    "openai" | "anthropic" | "google" | "cloudflare" | "ollama" | "openai-compatible";
 
-// Information about the AI gateway configuration. Returned by `AuthenticatedApi.getAiConfig()`.
+/** Information about the AI gateway configuration. Returned by `AuthenticatedApi.getAiConfig()`. */
```
kentonv: We use `//` style in this codebase, not `/*`.

I usually don't comment because it doesn't really matter but seeing a diff that just changes `//` to `/**` is triggering me...

packages/workshop-shared/src/api.ts
```diff
@@ -973,15 +974,30 @@ export type AiModelConfig = {
...
+  outputLimit?: number;
+
+  // Request behavior for an OpenAI-compatible endpoint. Defaults to "conservative".
+  compatibilityProfile?: "conservative";
 };
 
-// Workers AI adds the response cap to the prompt and rejects a request whose total exceeds the
-// model's window, so every Cloudflare model reserves this much of it for the response.
+/** Largest custom context window accepted for compaction and inference. */
+export const MAX_CUSTOM_MODEL_CONTEXT_WINDOW = 2_000_000;
```
kentonv: Why do we need to enforce a maximum?

packages/workshop-shared/src/api.ts
kentonv: Many of the diffs in this file are changing the comment format -- please revert spurious diffs.

packages/workshop-backend/src/ai-model-config.ts
```diff
@@ -0,0 +1,62 @@
+import {MAX_CUSTOM_MODEL_CONTEXT_WINDOW, type AiModelConfig} from "@gadgets/workshop-shared/api";
+
+const CHAT_COMPLETIONS_PATH = "/chat/completions";
+
+// Validate and normalize fields owned by the OpenAI-compatible provider.
+export function normalizeAiModelConfig(config: AiModelConfig): AiModelConfig {
+  if (config.provider !== "openai-compatible") return config;
```
kentonv, line 13: Why do we need to validate openai-compatible configs but not other types of configs?

A lot of this function is redundantly validating things that capnweb-validate will already have checked for us. Can it be simplified? Or maybe it can be removed entirely? Is any of this validation actually important?

packages/workshop-frontend/src/AddModelModal.tsx
```diff
@@ -31,6 +32,7 @@ const API_TOKEN_PLACEHOLDERS: Record<AiModelProvider, string> = {
   google: 'AIza...',
   cloudflare: 'Cloudflare API token',
   ollama: '(optional)',
+  'openai-compatible': 'sk-...',
```
kentonv: Is the `sk-` prefix required for all openai-compatible endpoints? I would guess not, so maybe the placeholder text needs to be more generic.

packages/workshop-frontend/src/AddModelModal.tsx
```diff
@@ -39,6 +41,9 @@ const FALLBACK_EXAMPLE_MODEL = { modelId: 'gemma4:31b', name: 'Gemma 4 31B' }
 
 // Pick an example model to show in the custom-model placeholders for the given provider.
 function exampleModel(provider: AiModelProvider): { modelId: string, name: string } {
+  if (provider === 'openai-compatible') {
```
kentonv: This is sort of inconsistent with how ollama is handled. ollama also has no suggested model, which means it falls back to FALLBACK_EXAMPLE_MODEL, which is actually currently set specifically to be a good example for ollama.

Seems like we should extend FALLBACK_EXAMPLE_MODEL to be a full on map from provider name to an exmaple model?

packages/workshop-backend/src/ai-model-config.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (!config.apiUrl) throw new Error("OpenAI-compatible base URL is required.");
+
+  let apiUrl: URL;
+  try {
+    apiUrl = new URL(config.apiUrl);
+  } catch {
+    throw new Error("OpenAI-compatible base URL must be a valid HTTPS URL.");
+  }
+  if (apiUrl.protocol !== "https:") {
+    throw new Error("OpenAI-compatible base URL must be a valid HTTPS URL.");
```
kentonv: Why can't they be HTTP? It could be running on localhost or a trusted local network, in which case you don't necessarily need HTTPS.

packages/workshop-backend/src/ai-models.ts
```diff
@@ -154,6 +156,21 @@ function workersAiCompat(catalog: Model<Api> | undefined): OpenAICompletionsComp
   };
 }
 
+const OPENAI_COMPATIBLE_PROFILES = {
```
kentonv: Pi seems to have some of its own code to guess at capabilities based on known hostnames, although apparently it doesn't have great defaults for hostnames that aren't in the list (it defaults to `supportsDeveloperRole: true` which several people complained about). It seems like it would be nice to leverage Pi's list though, rather than always assume conservative settings. Can we figure out how to do that?

packages/workshop-backend/src/chat-attachment-validation.ts
```diff
@@ -28,18 +29,26 @@ const isTextOrImageMime = (mimeType: string) =>
 const isTextImageOrPdfMime = (mimeType: string) =>
   isTextOrImageMime(mimeType) || mimeType === PDF_MIME_TYPE;
 
-// pi-ai encodes only text and image content parts, so text + images are universal. PDFs ride an
-// image part and are bridged to a provider's native document input where one exists: Gemini takes
-// application/pdf inline data as-is, and Anthropic/OpenAI payloads are rewritten in flight (see
-// chat-attachment-pdf.ts). Workers AI and Ollama chat endpoints have no document input at all.
+// PDFs ride an image part and are bridged to a provider's native document input where one exists.
```
kentonv: The old comment was much better, please revert this change.

packages/workshop-backend/src/agent-compaction.ts
```diff
@@ -25,6 +26,14 @@ const DEFAULT_CONTEXT_WINDOW = 128_000;
...
     {inputBudget: number, maxOutputTokens?: number} {
+  if (config.provider === "openai-compatible") {
+    let contextWindow = Math.min(config.contextWindow!, MAX_CUSTOM_MODEL_CONTEXT_WINDOW);
+    let maxOutputTokens = Math.min(config.outputLimit!, contextWindow - 1);
+    return {
+      inputBudget: contextWindow - maxOutputTokens,
+      maxOutputTokens,
+    };
+  }
   let model = SUGGESTED_MODELS[config.provider][config.model];
```
kentonv: Ugh I'm realizing this whole function is kinda broken, predating this change. It doesn't work for models not in SUGGESTED_MODELS!

@AshishKumar4 Can you please figure out a better approach here?  Does pi-agent-core give us any new options for figuring out a model's token limits? If not maybe we could pull from models.dev?

packages/workshop-shared/src/api.ts
ndisidore: was meant as a compaction guardrail, but it shouldn’t reject a provider’s advertised limit. I’ll remove the public maximum and keep any necessary safety bound internal to compaction

packages/workshop-backend/src/ai-model-config.ts
ndisidore: There's a lot of user input here because there's not much we can infer nor do I think its worth the complexity to do so - to keep things simple I want to support but leave as an advanced escape hatch

Cap’n Web covers the wire shape, but not the semantic checks or URL normalization. We still need the conditional fields, URL checks, integer limits, and outputLimit < contextWindow. I’ll remove anything that only repeats the generated validator.

packages/workshop-backend/src/ai-model-config.ts
ndisidore: Good point for local development. Production can’t reach private-network endpoints anyway. I’d allow HTTP for loopback hosts only and keep HTTPS required for public endpoints since we send a bearer token.

packages/workshop-backend/src/ai-models.ts
ndisidore: We actually already are using this but our `conservative` profile pins seven fields that generic endpoints commonly reject (which really just helps models without host entries).
If we're okay with it, we can remove compat and trust Pi completely. Known hosts improve, but unknown hosts receive optimistic OpenAI defaults.

packages/workshop-backend/src/generated/openai-compatible-compat.ts
```diff
@@ -0,0 +1,616 @@
+// Generated by scripts/build-openai-compatible-models.mjs. Do not edit.
```
ndisidore, line 1: These don't have me particularly excited, but the alternative is pulling in the full pi catalog with is hundreds of thousands of entries long.
This is built as a smaller subset with only the necessary info directly from Pi at build time

kentonv:
I don't think #184 was intended to close this; silly github.

@ndisidore are you still working on this?

ndisidore:
Yeah still spending some cycles here. Attempting to use `pi` for some of this config seems smart, but I'm trying to reconcile what it make sense to bundle vs not (since the entire pi catalog is 1MM+ entries, I don't want to ship to either the frontend or the backend, but that means some possible build script to pull out the "important ones/important bits")
Still thinking through this

PR #169, opened 2026-08-12
Add bounded progressive MCP discovery

dcartertwo:
An MCP grant is broader than the bounded catalog shown to the agent: a server-wide grant also covers authorized tools that fall beyond the catalog’s size limit. This adds bounded search and exact-name lookup so those tools remain discoverable and callable without loading an unbounded `tools/list` response.

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/169" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/gatekeeper-mcp-portal/README.md
+++ b/packages/gatekeeper-mcp-portal/README.md
@@ -28,7 +28,7 @@ every system the organization has connected in one click.
 pinning different tools of one upstream server share both the name and the endpoint, so the scope is
 the only thing that distinguishes them.
 
-The session API — a typed method per tool, plus `callTool`, `getActionResult`, and `listTools` — is
+The session API — a typed method per described tool, plus `callTool`, `getActionResult`, and `listTools` — is
 the same as [`gatekeeper-mcp`](../gatekeeper-mcp/README.md#what-it-provides).
 
 Scoping to one server also shrinks what the agent reads: a 57-tool portal generates 57 tool
--- a/packages/gatekeeper-mcp-portal/src/portal.ts
+++ b/packages/gatekeeper-mcp-portal/src/portal.ts
@@ -26,6 +26,7 @@ import {
 import type { ToolCatalog } from "@gadgets/mcp-shared/client";
 import {
   classifyTool,
+  MAX_TOOLS_PER_SERVER,
   type ServerTrust,
 } from "@gadgets/mcp-shared/tools";
 import { bindingNameFragment, hostOf } from "@gadgets/mcp-shared/util";
@@ -397,7 +398,7 @@ class McpServerConfiguratorUI extends RpcTarget implements McpServerConfigurator
   // form treats both as having nothing to grant rather than telling them apart.
   async listServerOptions(): Promise<ConfiguratorUIOption[]> {
     const { tools, truncated } = await this.#tools();
-    if (!looksLikePortal(tools, truncated)) return [];
+    if (!looksLikePortal(tools, { truncated, cap: MAX_TOOLS_PER_SERVER })) return [];
 
     const servers = reconcilePortalServers(
       await this.#fetchPortalServers(), tools, truncated);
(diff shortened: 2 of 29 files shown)


Maximo-Guk:
GPT Findings:

- High: Invalidated actions only stop the current drain. Later drains skip them and auto-apply successors. packages/workshop-backend/src/auto-approval.ts:61-86
- High: The new successful applyAction() invalidation result fails open during staggered deployment: an older backend ignores it and records the undispatched action as approved. Use a backward-compatible thrown sentinel. packages/mcp-shared/src/facet.ts:320-344
- Medium: Empty-string MCP cursors are treated as pagination completion, although cursors are opaque strings. packages/mcp-shared/src/client.ts:605-637
- Medium: Catalog refresh and approval-time tool revalidation bypass the new discovery semaphore/deadline, allowing stalled requests and unbounded waiters. packages/mcp-shared/src/facet.ts:193-205,331-342
- Low: Duplicate search terms can amplify matching work across the 4 MiB scan budget. Deduplicate/cap terms. packages/mcp-shared/src/tool-search.ts:39-53
- Low: Activity history labels invalidated actions but omits their reason. packages/workshop-frontend/src/Activity.tsx:611-644

packages/workshop-backend/src/auto-approval.ts
```diff
@@ -84,7 +84,8 @@ export class AutoApprovalDrainer {
       try {
         // Attribute the auto-approval to the user who enabled the rule -- it runs under their
         // authority.
-        await this.applyPendingAction(fresh, rule.enabledBy, true);
+        const outcome = await this.applyPendingAction(fresh, rule.enabledBy, true);
```
Maximo-Guk: I'm kind of surprised to see auto approval / kernel changes in this PR

> An MCP grant is broader than the bounded catalog shown to the agent: a server-wide grant also covers authorized tools that fall beyond the catalog’s size limit. This adds bounded search and exact-name lookup so those tools remain discoverable and callable without loading an unbounded tools/list response.

> Tool policy and connection generation are checked again before dispatch. If either changes while an action awaits approval, the action becomes explicitly invalidated, is never sent, and stops later auto-approved actions from running past it.


It kind of sounds like we might be solving two problems in this PR, should the second paragraph be another PR?

dcartertwo: The revalidation/invalidation path is intentionally part of this PR because progressive discovery allows an action to be staged from a tool definition that was fetched outside the bounded catalog. The connection generation, grant, and policy must be checked again at approval time before dispatch; splitting that out would leave an intermediate mergeable state where progressively discovered actions rely on stale approval metadata. I kept the Workshop/shared API portion limited to the terminal invalidated outcome and its in-order auto-approval barrier.

dcartertwo:
Update after splitting the kernel work from this PR: #169 is now discovery-only at `06870ec9`. It retains the opaque empty-cursor fix, bounded/deduplicated search, and bounded catalog discovery paths. Approval-time revalidation, rolling-deploy signaling, persistent auto-approval barriers, and invalidation UI moved to #198, stacked after #170.

Discovery-only verification: 300 MCP tests, both MCP connector suites/builds, and full `pnpm lint`/monorepo build.

packages/workshop-backend/src/auto-approval.ts
Maximo-Guk: Can't we just split the kernel/protocol changes out in to a separate PR and land that first?

dcartertwo: kernel changes are split out

Maximo-Guk:
Couple more 

```
1. Medium: packages/mcp-shared/src/client.ts:620-624 rejects an entire page before inspecting it when the page crosses 5,000 tools. findTool() therefore fails even when the requested tool is the first entry of a 5,001-tool page.
2. Low: Generated overloads at packages/mcp-shared/src/schema-to-ts.ts:364-367 allow a variable containing both search and name, despite runtime rejection. Add the opposite property as optional never to each specialized overload.
```

Maximo-Guk (approved):

packages/workshop-backend/src/auto-approval.ts
dcartertwo: Agreed and split. #169 is now discovery-only; the approval-time snapshot/revalidation, invalidation handling, kernel barrier, and UI are in #198, stacked after #170 as a separately reviewable commit.

dcartertwo:
> Couple more
> 
> ```
> 1. Medium: packages/mcp-shared/src/client.ts:620-624 rejects an entire page before inspecting it when the page crosses 5,000 tools. findTool() therefore fails even when the requested tool is the first entry of a 5,001-tool page.
> 2. Low: Generated overloads at packages/mcp-shared/src/schema-to-ts.ts:364-367 allow a variable containing both search and name, despite runtime rejection. Add the opposite property as optional never to each specialized overload.
> ```

updated

ndisidore (approved):
Reviewed and looks good! Big changeset, but on the whole seems sound

PR #172, opened 2026-08-12
Frontend reconnect hardening: tab wake & reconnect probe

ndisidore:
Handles 2 cases the server cannot automagically help with recovery on:
- Reconnecting flicker: probe each candidate with `ping()` and only publish a stub that has round-tripped. This stops false reporting a reconnected state just to go back to disconnected 
- Probe the connection on tab wake. `visibilitychange`/`online` triggers a probe of the current stub 

Screenshot from 2026-08-12 11-48-33
_note reconnecting chip_

###  Testing

Types, unit tests, lint. Manually against the local dev stack:

- Outage (backend down ~3 min with a workspace open): chip appears once, no flicker; exactly one RPC connection lost / one RPC connection restored; measured attempts at +0.2s, +1.3s, +3.2s, +6.8s, +14s, then ~10s apart (correct jittered doubling backoff); chat works after recovery.
- Zombie (kill -STOP on workerd, then a wake event): probe times out at ~10s, disposes the stub, and the app recovers on its own once the process thaws (no user action needed).
- Tab-switch noise (counted probe frames by hooking WebSocket.send): a burst of 8 wake/online events yields exactly 1 probe past the 15s idle window, 0 within it.

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/172" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/workshop-backend/src/server.ts
+++ b/packages/workshop-backend/src/server.ts
@@ -643,6 +643,8 @@ class PublicApiImpl extends RpcTarget implements PublicApi {
     this.users = this.ctx.exports.UserDurableObject;
   }
 
+  async ping(): Promise<void> {}
+
   async getServerConfig(): Promise<ServerConfig> {
     return getServerConfig(this.env);
   }
--- a/packages/workshop-frontend/src/GadgetEditor.tsx
+++ b/packages/workshop-frontend/src/GadgetEditor.tsx
@@ -16,6 +16,7 @@ import {
 } from '@phosphor-icons/react'
 import { RpcStub, RpcTarget } from 'capnweb'
 import { useAuthenticatedApi } from './AuthContext'
+import { useConnectionLost } from './RpcContext'
 import UserMenu from './components/UserMenu'
 import SiteLogo from './components/SiteLogo'
 
@@ -449,6 +450,7 @@ export default function GadgetEditor() {
   isEditingTitleRef.current = isEditingTitle
   const [titleInput, setTitleInput] = useState('')
 
+  const rpcConnectionLost = useConnectionLost()
   const {
     overseer,
     metadata,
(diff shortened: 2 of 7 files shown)


kentonv:
* I think there's a subtle change in behavior here. Previously, when the stub was detected broken, it was immediately replaced by a stub representing the _eventual_ new connection, so new RPCs made while the reconnection handshake was in flight would be queued and delivered once the connection was up. After this change, I believe we leave the old, broken stub in place until _after_ we've reconnected and probed the new connection to prove that it is good. This means in the meantime a bunch of RPCs might fail, but it would probably be better to have those RPCs wait for the new connection. I believe you should be able to accomplish this by setting the top-level stub to an `RpcStub` constructed from a `Promise<RpcStub>`, where you later resolve the promise to the final stub once the probe succeeds.

* `getServerConfig()` is not exactly free on the server side. It does a KV read and some other stuff, which is all wasted in the case that it's being used as a probe. I think we should instead add a `ping()` method that just returns nothing.

kentonv:
Hmm, it looks like Cap'n Web actually might not currently expose a way to construct an RpcPromise/RpcStub from a Promise<RpcStub> (or generally a Promise<T>). We should probably add that. In raw Cap'n Proto we use it all the time.

Either `RpcPromise` should have a constructor from `Promise` or we should have some sort of `makePipelined(p: Promise<T>): RpcPromise<T>` function...

ndisidore:
> Hmm, it looks like Cap'n Web actually might not currently expose a way to construct an RpcPromise/RpcStub from a Promise (or generally a Promise). We should probably add that. In raw Cap'n Proto we use it all the time.
> 
> Either `RpcPromise` should have a constructor from `Promise` or we should have some sort of `makePipelined(p: Promise<T>): RpcPromise<T>` function...

I actually have a branch that does exactly this: let me see if I can get it over the line

kentonv (approved):

PR #175, opened 2026-08-12
Bump capnweb to 0.11.1

Maximo-Guk:
Bumps capnweb `^0.10.0` → `^0.11.1` across the workspace. 0.11.0's shared `RpcPromise` alias (cloudflare/capnweb#237) is what unblocks TypeScript 7 type-checking

Second commit moves capnweb and capnweb-validate into the pnpm catalog (one edit per bump instead of 20 manifests), which also takes capnweb-validate `0.2.1` → `0.2.4` (cloudflare/capnweb#240: it ships its own capped typescript dependency for the `@validateRpc` transform)


Kieran-Hulsman (approved):
🔥

ndisidore (approved):

PR #177, opened 2026-08-13
Allow hooks to target any gadget in workspace.

kentonv:
The way that agents register hooks was originally designed when a workspace could only contain one gadget. When we extended workspaces to multiple gadgets, hooks were given a restriction that they could only target the first gadget in the workspace.

This change redesigns the API exposed to agents. See the changes to the prompt in `agent.ts` (written by me) to see how the API has changed. The changes in `overseer.ts` (written by Fable, reviewed by me) implement this change.

### DIFF
--- a/packages/workshop-backend/src/overseer.ts
+++ b/packages/workshop-backend/src/overseer.ts
@@ -51,12 +51,12 @@ const logger = createWorkshopLogger("workshop.overseer");
 export const AGENT_RUNNING_ERROR_MESSAGE = "Agent is running, wait for it to finish.";
 
 let CODE_MODE_HARNESS =
-`import { WorkerEntrypoint, restore, RpcStub, RpcTarget } from "cloudflare:workers";
+`import { WorkerEntrypoint, restore } from "cloudflare:workers";
 import agent from "agent.js";
 
 export default class extends WorkerEntrypoint {
   verify() {}
-  async run(self, callbackResolvers) {
+  async run(self, callbackResolvers, restoreForger) {
     let env = this.env;
     if (callbackResolvers) {
       for (let [index, {resolve, reject}] of Object.entries(callbackResolvers)) {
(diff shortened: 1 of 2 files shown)


packages/workshop-backend/src/overseer.ts
```diff
@@ -6259,15 +6324,73 @@ class OverseerImpl implements AgentHooks {
...
+    let codeId = crypto.randomUUID();
+    let forger: Fetcher<RestoreForgerEntrypoint>;
+    try {
+      this.#codeIdMap.set(codeId, RESTORE_FORGER_WORKER);
+      forger = await this.ctx.restore({type: "gadget", gadgetId, codeId});
+    } finally {
+      this.#codeIdMap.delete(codeId);
+    }
+
+    let stub = await forger.forge(params);
```
ndisidore, line 6376: Do we care try and and dispose here? i.e.
```ts
let stub: unknown;
try {
  stub = await forger.forge(params);
} finally {
  forger[Symbol.dispose]();
}
```

ndisidore (commented):
We have a test package for hooks here: `packages/workshop-backend/__tests__/overseer-hooks.test.ts`
If you could add tests for scenarios include multiple gadgets, non-default targets, persistence/reload/invocation, and non-gadget rejection I think that would go a long way for confidence. If testing at this layer of complexity is not feasible, fair.

packages/workshop-backend/src/overseer.ts
```diff
@@ -2927,11 +3003,19 @@ class OverseerImpl implements AgentHooks {
...
+    // forged the callback via `env.<GADGET>[restore]` during the currently-running executeCode
+    // invocation, so when exactly one gadget had a stub forged there, attribute the hook to it;
+    // otherwise (or for other callers) fall back to the workspace's first gadget.
+    // TODO: Replace this heuristic with introspection of the callback stub's actual restore
+    //   target once the runtime offers an API for that.
+    let gadgetId: WorkpieceId | undefined;
+    if (caller.from === "gadget" && caller.gadgetId !== undefined) {
+      gadgetId = caller.gadgetId;
+    } else {
+      gadgetId = (caller.from === "agent" ? this.#soleForgedRestoreTarget(caller.chatId) : undefined)
```
ndisidore, line 3028: Are we confident this works in practice? Maybe this isn't possible in practice but if n execution forges 2+ gadget stubs, it looks like the sole-forged-target heuristic gives up and every hook falls back to the workspace's first gadget.
I believe this is pre-existing, but this would ripple through a `removeGadget()` that uses that gadgetId, so removing the wrong gadget either kills healthy hooks or orphans live ones firing into a deleted gadget

packages/workshop-backend/src/overseer.ts
```diff
@@ -106,13 +138,60 @@ class PlaceholderRpcTarget extends RpcTarget {
...
+
 interface CodeModeEntrypoint extends WorkerEntrypoint {
   verify(): void;
   run(self?: unknown,
       callbackResolvers?: Record<string, {
         resolve: NativeRpcStub<(v: unknown) => void>,
         reject: NativeRpcStub<(e: unknown) => void>
-      }>): Promise<void>;
+      }>,
+      restoreForger?: RestoreForgerImpl): Promise<void>;
```
ndisidore: nit: for the type passed here do we want the concrete class type or should this be `NativeRpcStub<RestoreForgerImpl>`?

packages/workshop-backend/src/overseer.ts
kentonv: `forger` is a service stub (`Fetcher`), which doesn't require disposal since it isn't holding on to any live object (it is essentially just a worker ID + entrypoint name + props).

packages/workshop-backend/src/overseer.ts
kentonv: Right, this would break down if a single `executeCode` invocation registered multiple hooks pointing at different gadgets. That's fairly unusual but possible. Unfortunately, this is the best we can do right now. Per the TODO comment above, I'd like to add a runtime API to allow unwrapping the persistent stub so that we can find out which gadget it actually points at.

packages/workshop-backend/src/overseer.ts
kentonv: Good catch, it should indeed be NativeRpcStub.

ndisidore (approved):

kentonv:
> We have a test package for hooks here: `packages/workshop-backend/__tests__/overseer-hooks.test.ts`
> If you could add tests for scenarios include multiple gadgets, non-default targets, persistence/reload/invocation, and non-gadget rejection I think that would go a long way for confidence. If testing at this layer of complexity is not feasible, fair.

I asked the agent to write a test but I wasn't very happy with it -- I feel these unit tests are over-fitted and don't really tell you anything useful about whether the system works end-to-end.

I asked the agent to look at the integration test harness instead, but it found that the harness is too limited right now, and I didn't want to get distracted on improving the harness in this PR. E.g. the harness currently doesn't allow loading any dynamic code, for some reason, and there isn't a way to mock the agent to make the necessary `executeCode` calls.

I did, though, manually test this end-to-end by creating a two-gadget workspace and creating scheduled tasks hooked to each gadget. It all worked.

PR #178, opened 2026-08-13
Qualify Slack account identity with its workspace

bjesus:
The Workshop dedupes connected accounts by `(vendorId, uniqueName)`, so `uniqueName` has to be globally unique per vendor. A Slack handle is only unique *within* a workspace, so the unqualified handle we were reporting collapsed two different workspaces the same person connected into a single account.

Qualify it with the workspace host from `auth.test` (`handle@workspace-host`), falling back to the team ID and then to the bare handle. `displayName` is unchanged, so the UI still shows the person's real name.

Follow-up to the Slack gatekeeper; no new OAuth scope is required. Existing connected accounts keep their legacy bare `uniqueName` until the user reconnects, which is cosmetic.

### DIFF
--- a/packages/gatekeeper-slack/src/slack-api.ts
+++ b/packages/gatekeeper-slack/src/slack-api.ts
@@ -403,13 +403,20 @@ export class SlackApi {
     return user;
   }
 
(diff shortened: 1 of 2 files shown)


bjesus:

PR #181, opened 2026-08-13
Guide agents to build responsive gadgets

dcartertwo:
## Summary
- tell coding agents to make Gadget UIs responsive and phone-friendly by default

## Testing
- not run (prompt-only change)

### DIFF
--- a/packages/workshop-backend/src/agent.ts
+++ b/packages/workshop-backend/src/agent.ts
@@ -416,6 +416,8 @@ document.body.appendChild(document.createTextNode(greeting));
 
 Note that there is no index.html. Instead, client.js must build the entire UI using JavaScript code.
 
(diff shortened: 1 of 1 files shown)


packages/workshop-backend/src/agent.ts
```diff
@@ -416,6 +416,8 @@ document.body.appendChild(document.createTextNode(greeting));
 
 Note that there is no index.html. Instead, client.js must build the entire UI using JavaScript code.
 
+Make Gadget UIs responsive and usable on phones by default.
```
kentonv: Maybe "both desktop and phones" so it doesn't think it's only targeting phones.

dcartertwo: good catch

PR #197, opened 2026-08-13
add entrypoint-based custom export api

maxwellpeterson:
Implements the Gadget file export API design proposed in #114, which I would read first (it's small). The API interface is unchanged in this PR.

The implementation is an extension of the existing PDF export infrastructure. The same remote browser setup is extended to support HTML, PNG, and JPG exports in addition to PDF. [DOMPurify](https://github.com/cure53/dompurify) is added as a dependency to assist with HTML sanitization on export. This is slightly unfortunate since newer browsers support the [HTML Sanitizer API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API) which we could use instead, but the older version of Chromium used by Cloudflare's remote browsers does not support this API. I also refactored functions passed to `page.evaluate()` out into a new `packages/workshop-backend/browser/browser-export-page.ts` file so they can be properly typed for the browser environment that they run in.

This PR doesn't attempt to detect or prevent Gadget code changes during the export process. It's possible that Gadget  code will change between the user selecting an export format and the exported file being produced. This came up repeatedly in local AI code review, but I'm not sure it's a big problem in practice. Exports may fail or produce mismatched file types if Gadget code changes at specific points in the process. We could detect these changes and abort the export, but this feels harsh since most code changes are unlikely to be problematic. Gadget export code (if any exists) should be infrequently updated and cover a small fraction of all total Gadget code. More sophisticated change detection and revision pinning across export operations both add complexity that I wasn't sure we needed. If we did want to revisit this in the future, we could add stronger guardrails without changing the current Gadget API interface.

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/197" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/workshop-backend/src/agent.ts
```diff
@@ -490,7 +490,11 @@ Note that there is no index.html. Instead, client.js must build the entire UI us
 
 Make Gadget UIs responsive and usable on both desktop and phones by default.
 
-Every Gadget UI can be exported to PDF using platform-owned controls outside the Gadget. Never add print or export UI to a Gadget and never call \`window.print()\`. When asked to support or improve PDF export, only add standard print CSS such as \`@media print\`, \`@page\`, and CSS fragmentation properties so the PDF remains readable.
+Every Gadget UI can be exported to HTML or PDF using platform-owned controls outside the Gadget. Never add print or export UI to a Gadget and never call \`window.print()\`. Browser-mode PDF exports render using print media; HTML, PNG, and JPEG exports render using screen media. When asked to support or improve PDF export, use standard print CSS such as \`@media print\`, \`@page\`, and CSS fragmentation properties so the output remains readable.
```
kentonv: At this point we should probably split all the export stuff into its own section with a heading.

packages/workshop-backend/src/agent.ts
```diff
@@ -490,7 +490,11 @@ Note that there is no index.html. Instead, client.js must build the entire UI us
...
+
+During a browser-mode export, the client context is initialized with another special global variable named \`gadgetExportFormatId\`. This variable is only defined during export; during normal interactive rendering, referencing it directly throws a \`ReferenceError\`. Guard access with \`typeof gadgetExportFormatId !== "undefined"\` or read \`globalThis.gadgetExportFormatId\`. Use \`gadgetExportFormatId\` when the Gadget supports multiple HTML, PDF, PNG, or JPEG export variants. Do not declare or import \`gadgetExportFormatId\` in client.js.
```
kentonv: Does the gadget need to signal somehow when it's ready to render?

Could we say: Use top-level await to wait until the the UI is fully rendered.

I like this since it feels natural and doesn't require defining any further API surfaces.

packages/workshop-backend/src/agent.ts
```diff
@@ -490,7 +490,11 @@ Note that there is no index.html. Instead, client.js must build the entire UI us
...
+
+To add, replace, or disable export formats, server.js may export a class named \`ExportHandler\`, which must extend \`WorkerEntrypoint\`. Its \`getExportFormats(gadget)\` method returns the complete list of formats, and its \`export(gadget, id)\` method returns a \`ReadableStream<Uint8Array>\` for formats whose mode is \`"server"\`. Read any needed Gadget state before \`export()\` returns; do not capture the borrowed \`gadget\` parameter in the returned stream. If \`getExportFormats(gadget)\` returns only browser-mode formats, do not implement \`export(gadget, id)\`. \`export\` is valid as a JavaScript class method name; write it directly as \`async export(gadget, id)\`, without quoting it or using a computed property. Browser mode supports \`text/html\`, \`application/pdf\`, \`image/png\`, and \`image/jpeg\`; server mode supports any media type. Each format must contain a unique non-empty \`id\`, a \`label\`, a \`mode\`, a \`contentType\`, and a \`fileExtension\` beginning with a dot. Returning an empty list disables export. The Workshop supplies default HTML and PDF formats only when server.js does not export \`ExportHandler\` at all.
```
kentonv: We should provide example code.

packages/workshop-backend/src/overseer.ts
```diff
@@ -2521,13 +2529,79 @@ class OverseerImpl implements AgentHooks {
...
+    let browser = this.env.BROWSER;
+    if (!browser) throw new Error("Gadget export is not configured for this deployment.");
+    let bundle = this.getGadgetUiBundle(gadgetId, chatId);
+    if (!bundle) throw new Error("This Gadget does not have a UI to export.");
+    let browserGadget = await this.getGadgetFacet(gadgetId, chatId);
+    let title = this.getGadgetRecord(gadgetId).title;
+    return renderGadgetInBrowser(browser, bundle.jsCode, title, browserGadget, format);
+  }
+
+  prepareGadgetChat(chatId?: number): void {
```
kentonv: Name seems kinda vague.

Maybe: `checkChatExistsAndMaterializeDrafts`

packages/workshop-backend/src/overseer.ts
```diff
@@ -2521,13 +2529,79 @@ class OverseerImpl implements AgentHooks {
...
+      : Promise<ReadableStream<Uint8Array>> {
+    this.prepareGadgetChat(chatId);
+    let {formats, handler} = await this.#resolveGadgetExportFormats(gadgetId, chatId);
+    let format = formats.find(candidate => candidate.id === formatId);
+    if (!format) throw new Error(`This Gadget does not support export format: ${formatId}`);
+
+    if (format.mode === "server") {
+      if (!handler) throw new Error("The Gadget export handler is unavailable.");
+      using gadget = await this.getGadgetFacet(gadgetId, chatId);
+      // Cap'n Web and native RPC stubs interoperate at runtime, but their types do not yet.
```
kentonv: Hmm, `getGadgetFacet()` actually returns a `NativeRpcStub` that has been coerced to Cap'n Web `RpcStub`, so this is just coercing it back. That's going to need some cleanup at some point but maybe not in this PR.

packages/workshop-backend/src/overseer.ts
```diff
@@ -2521,13 +2529,79 @@ class OverseerImpl implements AgentHooks {
...
+  async exportGadget(gadgetId: WorkpieceId, formatId: string, chatId?: number)
+      : Promise<ReadableStream<Uint8Array>> {
+    this.prepareGadgetChat(chatId);
+    let {formats, handler} = await this.#resolveGadgetExportFormats(gadgetId, chatId);
+    let format = formats.find(candidate => candidate.id === formatId);
+    if (!format) throw new Error(`This Gadget does not support export format: ${formatId}`);
+
+    if (format.mode === "server") {
+      if (!handler) throw new Error("The Gadget export handler is unavailable.");
+      using gadget = await this.getGadgetFacet(gadgetId, chatId);
```
kentonv: `#resolveGadgetExportFormats` also calls this internally, maybe it should just return the stub rather than acquire a new one here?

packages/workshop-backend/src/overseer.ts
```diff
@@ -2521,13 +2529,79 @@ class OverseerImpl implements AgentHooks {
...
+      // Cap'n Web and native RPC stubs interoperate at runtime, but their types do not yet.
+      return await exportServerFormat(() =>
+        handler.export(gadget as unknown as NativeRpcStub<any>, format.id));
+    }
+
+    let browser = this.env.BROWSER;
+    if (!browser) throw new Error("Gadget export is not configured for this deployment.");
+    let bundle = this.getGadgetUiBundle(gadgetId, chatId);
+    if (!bundle) throw new Error("This Gadget does not have a UI to export.");
+    let browserGadget = await this.getGadgetFacet(gadgetId, chatId);
```
kentonv: Here too we could potentially reuse the stub from `#resolveGadgetExportFormats`.

packages/workshop-backend/src/overseer.ts
```diff
@@ -2521,13 +2529,79 @@ class OverseerImpl implements AgentHooks {
...
+    let format = formats.find(candidate => candidate.id === formatId);
+    if (!format) throw new Error(`This Gadget does not support export format: ${formatId}`);
+
+    if (format.mode === "server") {
+      if (!handler) throw new Error("The Gadget export handler is unavailable.");
+      using gadget = await this.getGadgetFacet(gadgetId, chatId);
+      // Cap'n Web and native RPC stubs interoperate at runtime, but their types do not yet.
+      return await exportServerFormat(() =>
+        handler.export(gadget as unknown as NativeRpcStub<any>, format.id));
+    }
```
kentonv, line 2580: Nitpick but I feel like `} else {` here wrapping the rest of the function would be clearer.

packages/workshop-backend/src/agent.ts
maxwellpeterson: Moved to an "Exporting files from Gadgets" subsection under "Writing Gadgets"

packages/workshop-backend/src/overseer.ts
maxwellpeterson: Yeah, I would prefer to leave this for a separate PR

packages/workshop-backend/src/agent.ts
maxwellpeterson: I like this too, previously we were waiting for the DOM to "settle" (no updates for 250ms) but top-level await is simpler mechanically and gives client.js more direct control over when export starts

kentonv (approved):

ndisidore:
bonk break glass

ndisidore (approved):

PR #170, opened 2026-08-12
Support large MCP portal catalogs

dcartertwo:
Depends on #169.

An MCP Portal flattens many upstream servers into one tool catalog, which may be too large to retain safely. The configurator trusts the portal’s bounded server list first, falls back to a compact global index only when that list is incomplete, and fetches one bounded summary list for the selected server. Every grant names exactly one upstream server, optionally narrowed to specific tools.

Portal-native management tools are never grantable. Truncated or malformed discovery fails closed, local scope errors are rejected before remote scans, and all catalog work is bounded.

The endpoint must expose direct tools. Deployments where Code Mode is default-on can opt this client out with `?codemode=off`; enforced Code Mode is unsupported.

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/170" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #199, opened 2026-08-13
Harden MCP portal authentication changes

dcartertwo:
Portal authentication mode and deployment-owned tokens can change while an account or facet remains live. This makes auth-mode transitions clear stale OAuth state, refuses fallback/callback paths no longer allowed by current configuration, and invalidates captured connection generations when token authority changes.

OAuth redirects record their configuration revision before handing off to the callback, preventing an existing facet from invalidating the attempt while the user is authorizing.

Stacked on #170. Verified with the MCP and portal suites plus full `pnpm lint`.

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/199" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/gatekeeper-mcp-portal/__tests__/account-revision.test.ts
+++ b/packages/gatekeeper-mcp-portal/__tests__/account-revision.test.ts
@@ -1,0 +1,54 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import {
+  McpAccountBase,
+  type ConnectedServer,
+  type ConnectOutcome,
+} from "@gadgets/mcp-shared/account";
+import { McpAccount } from "../src/portal.js";
+
+afterEach(() => vi.restoreAllMocks());
+
+describe("McpAccount configuration revision", () => {
+  it("records the current revision before handing an OAuth attempt to its callback", async () => {
+    const values = new Map<string, unknown>([["portalConfigRevision", "old"]]);
+    const ctx = {
+      id: { toString: () => "account" },
+      storage: {
+        kv: {
+          get: (key: string) => values.get(key),
+          put: (key: string, value: unknown) => values.set(key, value),
+          delete: (key: string) => values.delete(key),
+        },
+      },
+      exports: {},
+    };
+    const env = {
+      MCP_PORTAL_URL: "https://portal.example.com/mcp",
+      MCP_PORTAL_AUTH: "oauth",
+    };
+    const server: ConnectedServer = {
+      endpoint: env.MCP_PORTAL_URL,
+      serverId: "portal",
+      serverName: "Portal",
+      provenance: "deployment",
+      auth: "oauth",
+    };
+    const base = McpAccountBase.prototype as unknown as {
+      beginConnect(nonce: string, target: ConnectedServer | null): Promise<ConnectOutcome>;
+    };
+    vi.spyOn(base, "beginConnect")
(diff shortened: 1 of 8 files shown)


packages/gatekeeper-mcp-portal/vitest.config.ts
```diff
@@ -0,0 +1,13 @@
+import { fileURLToPath } from "node:url";
+import capnwebValidate from "capnweb-validate/vite";
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  plugins: [capnwebValidate()],
+  test: {
```
ndisidore, line 7: per sol
> C1 — HIGH — vitest.config.ts:5-13 omits include, so the suite runs a stale shadow copy that asserts the opposite of this commit's fix
All three flagged this. Two independently produced runtime evidence. I confirmed it directly:
packages/gatekeeper-mcp-portal/vitest.config.ts   → no `include`, no `environment`
packages/mcp-shared/vitest.config.ts (its source) → include: ["__tests__/*.test.ts"], environment: "node"
>
> .wrangler/validate/__tests__/config.test.ts       Aug 13 16:15  (pre-branch, collectible)
> .wrangler/validate/src/config.ts:131              return (connected === "token") !== (configured === "token");
pnpm test:run reports 5 files / 67 tests where only 3 files exist; a clean checkout gives 3 / 38. The consequence is not merely noise — the run simultaneously asserts both answers for the same input:
> - new __tests__/config.test.ts:150: portalAuthRequiresReconnect("oauth","none") → true
> - stale .wrangler/.../config.test.ts:144: same call → false
So the package's own green suite proves nothing about the function this commit exists to change, and a future revert would be confirmed by a passing test. Compounding it: .wrangler/** is deliberately excluded from the test task's fingerprint (scripts/vitest-task-vite-config.ts), so a cached run can replay green over a snapshot nothing invalidates, and pnpm clean (rm -rf dist src/generated) never removes it.
> 
> Fix is one line: include: ["__tests__/*.test.ts"] (plus environment: "node" for parity). Adding capnwebValidate() is correct and necessary — only the omissions are the problem.

packages/gatekeeper-mcp-portal/__tests__/account-revision.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const internals = account as unknown as {
+      awaitingSelection(nonce: string): boolean;
+      server(): ConnectedServer | undefined;
+    };
+    vi.spyOn(internals, "awaitingSelection").mockReturnValue(true);
+    vi.spyOn(internals, "server").mockReturnValue(server);
+
+    await expect(account.beginConnect("nonce", server)).resolves.toMatchObject({ kind: "redirect" });
+
+    expect(values.get("portalConfigRevision")).not.toBe("old");
```
ndisidore, line 78: this test is very light and doesn't really exercise the riskiest parts of this

> Untested branches introduced by the diff: all four guards at portal.ts:328-329; both sides of the legacy previous === undefined branch; the rotation branch at 343-346; the #connectingRevisions refcount; allowsOAuthCallback (both outcomes, in either package); the portal's own allowsOAuthFallback (mcp-shared tests a hand-written fake instead); invalidateConnectionState() (no test invokes it at all); and — noted by agent 2 — portalAuthRequiresReconnect("token","token"), the single most common deployment configuration, asserted nowhere. A regression making it true bricks every token deployment silently.

ndisidore (approved):

PR #203, opened 2026-08-14
Honor rejectAction restarts and dependent-action rejections

ndisidore:
Rejecting an action had two gaps:

1. `Gatekeeper.rejectAction()` can return `{restart: true}`, and its docs promise the overseer will restart the Gadget but the overseer ignored the return value entirely, so the Gadget kept running against simulated state that no longer existed.
2. When rejecting a page creation, confluence and notion also delete the pending actions that depended on it (edits, child pages)  _but only locally_. The overseer's approval records for those actions stayed pending forever, and resolving them later failed with "unknown action".

This PR makes the overseer honor `restart` (it aborts the calling gadget's facet, only after all writes commit) and adds a feedback channel for the cascade: `rejectAction()` can now return `rejectActions`, the gatekeeper's own action IDs that the rejection took down with it. The overseer marks those records rejected too, tagged with `cascadedFrom` so the activity log shows why. Confluence and notion adopt it here; github and linear are follow-ups.

Also fixes a small latent bug: rejecting a manual gate now drains auto-approvable actions queued behind it, matching what approve already did.

Backward compatible: existing gatekeepers returning `void` or `{restart}` are unchanged. New unit tests cover the cascade, restart ordering, and race guards.


packages/workshop-backend/src/action-rejection.ts
ndisidore, line 1: I know there's aversion to adding new files. If its an issue I'll move it into overseer

Maximo-Guk:
GPT Findings:

1. High: In-flight approval can overwrite a cascade rejection. applyPendingAction() persists its stale pre-await record after applyAction() returns, restoring state to approved after rejection. packages/workshop-backend/src/overseer.ts:2506, packages/workshop-backend/src/action-rejection.ts:102
2. High: {restart:true} does not cancel an existing auto-approval drain. The drain can resume its snapshot and apply later actions despite the restart guard only skipping a new drain. packages/workshop-backend/src/auto-approval.ts:63, packages/workshop-backend/src/overseer.ts:7803
3. Medium: Only the primary caller’s gadget restarts. Cascaded actions from other gadgets can leave those gadgets running against invalidated simulated state. packages/workshop-backend/src/action-rejection.ts:41

packages/workshop-backend/src/action-rejection.ts
Maximo-Guk: Personally in favor of breaking down this monolith so love this, and we already kind of did this with the `auto-approval.ts`

ndisidore:
Known follow-ups deliberately left out of this PR (from review discussion):

**Cascade adoption in remaining gatekeepers** — github (`#rejectActionsForResource`, `#rejectReplyDependencyChain`), linear (`#cascadeRejectProvisional`), and spotify (`playlistCreate` reject) all cascade-reject dependents locally without reporting them, leaving orphaned pending overseer records (pre-existing; recoverable via reject, which no-ops gatekeeper-side on unknown actions). The overseer core already supports cascade-without-restart, so adoption is mechanical.

**Restart per cascaded caller** — restart currently targets only the primary record's calling gadget. In a multi-gadget workspace sharing a gatekeeper, cascaded records staged by *other* gadgets leave those gadgets running against invalidated simulated state (documented as a limitation in `action-rejection.ts`). Every cascaded record already carries its own `caller`, so the fix is to hand the cascaded records to the restart hook and abort each distinct gadget facet once. Deferred because it generalizes the `restart` flag's meaning from "restart my caller" to "restart everyone whose staged work I invalidated" — worth its own review. Note the residual gap either way: gadgets that merely *observed* simulated state without staging anything aren't reachable by this mechanism.

**Pending-state guard in gatekeeper rejects (R1)** — confluence/notion `rejectStoredAction` deletes unconditionally without checking `record.state`, so an approve/reject race can destroy an applied record (and its revert info) gatekeeper-side. One-line guard (`if (record.state !== "pending") return;`) in each; the overseer side already re-gets and warns.

**Notion/confluence restart policy for lone creates** — restart is derived from cascade-emptiness, so rejecting a `createPage`/`createContent` with no dependents requests no restart even though the gadget holds a dangling provisional ID. Pre-existing policy, newly consequential now that restart is honored; decide alongside cascade adoption.

**Staged-vs-submitted race** — `stageAction` persists the store record before awaiting `submitAction` (input gate open across the RPC), so a cascade can report a key the overseer hasn't seen yet, and the late-landing submit creates an orphan pending record (recoverable via reject; surfaced by the new unmatched-key warn). Real fix wants github-style staged/submitted states — natural to fold into the shared action store work.

**Frontend `cascadedFrom` badge** — the field flows through `ActionLogEntry` but nothing renders it yet; cascaded rejections currently display like manual denials. Small Activity/chat-card badge ("skipped: depends on #N").

Maximo-Guk (approved):

packages/workshop-backend/src/action-rejection.ts
kentonv: I'm in favor of breaking down the monolith but against every new function and type declaration being its own file -- the opposite end of the spectrum is worse IMO. We need some sort of sane organization. I've been meaning to think about the right way to break up overseer.ts...

packages/workshop-backend/src/action-rejection.ts
```diff
@@ -0,0 +1,151 @@
...
+        (rec): rec is ActionRecord & { type: "action" } =>
+            rec.gatekeeperId === record.gatekeeperId && rec.type === "action" &&
+            rec.state === "pending" && rec.id !== record.id && keys.has(rec.action));
+
+    for (let candidate of cascade) {
+      keys.delete(candidate.action);
+      // Defensive re-check. The loop is fully synchronous today (no awaits since the snapshot),
+      // so this can never fire -- it stays so the mutation remains locally safe if an await is
+      // ever introduced above.
+      let fresh = storage.actions.get(candidate.id);
```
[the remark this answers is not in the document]
ndisidore: Good catch on the mechanics, but the composed scenario can't happen: the cascade's contents are decided at the **gatekeeper**, which serializes apply against reject.

- If the apply is processed first, the record flips to `applied` gatekeeper-side, and the purge loops only iterate pending records so it can never appear in `rejectActions`, and the cascade never touches it.
- If the reject is processed first, the record is deleted gatekeeper-side, so the parked `applyAction()` throws "Unknown action"  `applyPendingAction` throws before its write.

So "apply succeeds AND the same action is cascade-reported" requires a gatekeeper that reports keys it hasn't actually invalidated, which the `rejectActions` contract forbids.

On the suggested fix: skipping the write after a successful apply would make this worse at that point the action is real in the world, so a record left saying "rejected" is a bigger lie than the overwrite. World wins.

The fair point underneath is that this rests on gatekeeper conformance and a violation would be silent. Added in ed9fcf36: `applyPendingAction` now re-gets after the await and logs an error if the record raced to non-pending (then still writes approved), so a nonconforming cascade adopter is observable.

packages/workshop-backend/src/overseer.ts
```diff
@@ -7770,21 +7793,48 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
...
     // sibling approval from resuming this turn.
+
+    // Rejecting a manual gate (or cascading over several) may unblock later auto-eligible pending
+    // actions on the same gatekeeper, so drain once the rejection is fully applied (one drain
+    // suffices: the cascade is same-gatekeeper by construction) -- UNLESS the gatekeeper requested
+    // a restart. Restart signals that remaining pending actions may have been staged on top of
+    // now-invalid simulated state (e.g. edits stacked on a rejected edit of the same page), so
+    // don't rush to auto-apply them; a later trigger will drain once the queue head is decided.
+    if (!result?.restart) {
+      this.impl.ctx.waitUntil(this.impl.drainAutoApprovals(action.gatekeeperId));
```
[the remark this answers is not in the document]
ndisidore: Confirmed — the trace is exactly right: `#cancelled` was only cleared when the cancelled drain fully unwound, so a resume that coalesced into the still-parked drain ran its rerun pass against the still-set flag and did nothing. There was a second hole too: `cancel()` no-oped when no drain was in flight, so a drain triggered *during* the rejection (e.g. by a concurrent submit) ran unimpeded through the protected window.

Fixed in 25b0d5a2 by replacing cancel with what the comments were already describing — a **pause**: refcounted `pause()`/`resume()` on the drainer, drain requests made while paused coalesce or park instead of being dropped, and the reject path pauses after validation, resumes in a `finally`, and re-triggers a drain only on success without restart. Your exact ordering (resume + re-trigger while the old drain is still parked mid-apply) is now a regression test.

ndisidore:
Simplification pass: the auto-approval pause/cancel machinery is removed and the branch was squashed to 3 commits for review (prior threads' anchors may show "outdated").

Why removed: the pause was a racy best-effort delay, not an invariant — it had its own deterministic leak (when the gatekeeper serializes reject before apply, the `finally`-resume runs in the reject response's synchronous continuation, so a parked drain always observes "unpaused" and proceeds), and even when it worked it only delayed an outcome that recurs at the next trigger anyway: restart-surviving auto-eligible actions auto-apply on the next submit/approve/rule-enable regardless.

Correctness never depended on it. The validity authority is the gatekeeper's cascade: a cascade-rejected action is deleted on the gatekeeper side, so applying it throws "Unknown action" and the drainer stops; the drainer also re-gets each record before applying. `auto-approval.ts` and its tests are now byte-identical to main.

What remains from the concurrency rounds:
- `if (!result?.restart)` around the reject-path drain trigger — a stateless timing choice (don't eagerly re-drain right after a restart signal), documented as such, not a correctness gate.
- The `applyPendingAction` raced-record detection log (a conforming gatekeeper can't cascade-reject an action it applied, so a raced record fingers a nonconforming gatekeeper).

If truly *blocking* restart-suspect survivors is ever wanted, it needs storage-backed per-record state consulted at the drainer's re-get (e.g. clearing `autoApprovable`), not an in-memory pause — noted as possible future work, not built here.

packages/workshop-backend/src/overseer.ts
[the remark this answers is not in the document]
ndisidore: Superseded — rather than fixing the pause, the machinery is deleted: it only re-timed an inevitable outcome (survivors drain at the next trigger regardless) and correctness is already enforced by the gatekeeper's cascade + apply-throws + the drainer's re-gets. The swallowed-resume regression test went with the flags it guarded.

packages/workshop-backend/src/action-rejection.ts
ndisidore: Yeah that's fair: an `actions.ts` is probably better then approve and reject being their own files.
But for now I'll just bring it in to overseer

ndisidore:
**Concurrency review notes — what's fixed here vs. follow-up work.**

While reviewing the reject path we found that a concurrent approve + reject of the same action can interleave (the overseer DO's input gate is open across the profile fetch and the gatekeeper RPC). c76a3a6f hardens the worst consequence in this PR: gatekeeper-side `rejectStoredAction` now no-ops unless the record is still pending, so a reject that loses the race can no longer delete the *applied* record (which orphaned revert) or cascade-delete still-valid pending children of an applied creation. The requirement is now stated in the `Gatekeeper.rejectAction()` contract so future adopters (github, linear) inherit it.

Remaining follow-ups, deliberately out of scope for this PR:

**1. Serialize approve/reject per action in the overseer.** Both callers can still return success when only the approval took effect (final record state is truthful — approved — and both sides log, but the rejecting client isn't told it lost). Proposed shape: an in-memory per-action-id single-flight lock in `OverseerImpl` acquired in `applyPendingAction` and around `rejectPendingAction`, with a re-get + pending re-validation after acquiring. That also makes the lost reject surface naturally as the existing `Action is not pending` error instead of a silent success. Related: the "nonconforming gatekeeper" diagnosis in `applyPendingAction`'s raced-record log is misleading — a plain human approve-vs-reject race produces the same state.

**2. Cascaded rejections don't restart other affected gadgets.** Simulated state is per *gatekeeper* (one overlay store per gatekeeper DO, reads unfiltered by caller), so a cascade can invalidate what a *different* gadget — or a gadget, when the primary was agent-called — already observed, and only the primary caller's gadget is restarted today. This is the documented tradeoff in `rejectPendingAction`'s JSDoc ("rare; re-observes on next run"), and it's strictly better than main (which honored neither restart nor cascade), but if we want to close it: have `rejectPendingAction` return the matched cascade records and let `rejectAction` restart every distinct gadget caller among primary + cascade when `restart` is set. Exposure is narrow (shared gatekeeper + other gadget mid-execution), so this needs a judgment call on whether it's worth it.

**3. (Deeper, gatekeeper-internal)** A reject of action Y that cascade-deletes X while X's apply HTTP call is mid-flight can have the apply continuation resurrect X's deleted record via `putAction`. Pre-existing on main, not addressable from the overseer side; noting it for completeness.

Issue #205, opened 2026-08-14
Long agent turns can lose completion when their Worker execution ends


andrskr:
## Problem

I ran into a lifecycle problem while testing longer agent tasks. A turn starts in the workspace Durable Object, but the path that returns its result still depends on the Worker execution that started it.

Short turns finish before this matters. A longer turn can outlive that execution. When this happens, the user can lose the final result even though the workspace still knows that the turn was active.

## Why this matters

The duration of a turn depends on the model, tools, and work being done. A correct task should not fail only because it needed more time.

The lifecycle also needs to remain consistent after a restart. We should not run the same turn twice, leave stale active state, or require the user to retry without knowing whether the original work is still running.

## Expected behavior

A long turn should have a durable completion path. It should either return its result or fail cleanly. A restart should not duplicate the work, and a finished turn should not leave active state behind.

## Related changes

The implementation is split into two stacked pull requests:

1. #206 adds the release support needed to deploy Workflow definitions.
2. #207 adds the durable agent-turn lifecycle and depends on #206.

PR #206, opened 2026-08-14
Support Workflow definitions in release manifests

andrskr:
Part of #205.

## What does this change?

This adds Workflow bindings and same-Worker Workflow definitions to release manifest version 2. Workflow names use the deployed Worker name, so each installation receives the correct binding and definition.

The change also adds a focused manifest test. It does not enable a Workflow in any Worker.

A Workflow binding does not create a same-Worker Workflow during a fresh installation. The deploy service needs the definition after it uploads the Worker. The manifest now carries that definition next to the binding, and the test checks the complete output.

## Checklist


andrskr:

andrskr:
recheck

PR #207, opened 2026-08-14
Make long-running agent turns durable

### LINKED ISSUE #205: Long agent turns can lose completion when their Worker execution ends

andrskr:
Closes #205.

This pull request is stacked on #206. It will contain the release-support commit until #206 merges.

## What does this change?

Agent turns still start through the current path. If a turn remains active for 60 seconds, the workspace promotes it to a Cloudflare Workflow with a stable ID. The Workflow adopts the existing turn and waits for its completion. It never starts a second model run.

The Durable Object alarm handles only promotion, bounded retries, completion delivery, and cleanup. A durable outbox keeps the completion notification until it is delivered. Short turns finish without creating a Workflow.

The change also registers the Workflow and adds tests for promotion, restart recovery, duplicate creation, parallel turns, completion delivery, and cleanup.

A long turn can outlive the Worker execution that started it. If that execution ends, the final result no longer has a durable delivery path. Increasing a timeout only moves the same failure.

The Workflow becomes the durable owner without changing the existing agent execution. Every state transition uses the stored chat ID and Workflow ID. Duplicate creation adopts the same Workflow, retries are bounded, and terminal paths remove the stored state. The tests exercise each transition directly.

## Checklist

- [ ]  This is a small, concrete change; it is not a feature, refactor, or low-value cleanup.


andrskr:

andrskr:
recheck

PR #208, opened 2026-08-14
Stop dropping Context collections from the agent catalog

AshishKumar4:
## What does this change?

With hundreds of skills the agent saw a truncated Context Library, and collections went missing from it. Collections are how the agent finds anything at all; individual skills it can enumerate through the session's `list()`/`search()`.

The returned catalog was always being bounded to a low value of 25, and at both, before returning the values in RPC and an outer bound, which made most of the skills and collections be truncated.
This PR fixes that by increasing the outer bound threshold, and separately bounding the skills catalog to a value of 150.
This PR also simplifies the `getAgentCatalog` API by removing the request field.
Also, `normalizeAgentCatalog` clamps from the tail and sorts the survivors, so the gatekeeper's order decides what lives

The behavioural change is one moved `.toSorted()` and one reordered concatenation. The catalog policy moved into `buildContextCatalog()`, so it reads as a single pure function instead of being spread through a Durable Object method.

## Checklist

### DIFF
--- a/packages/gatekeeper-context/src/library-gatekeeper.ts
+++ b/packages/gatekeeper-context/src/library-gatekeeper.ts
@@ -5,9 +5,8 @@
 import { WorkerEntrypoint, DurableObject, RpcStub as NativeRpcStub, RpcTarget as NativeRpcTarget } from "cloudflare:workers";
 import { RpcStub } from "capnweb";
 import { validateRpc, skipRpcValidation } from "capnweb-validate";
-import { boundAgentCatalog } from "@gadgets/workshop-shared/gatekeeper";
 import type {
-  VendorDescription, AccountDescription, AgentCatalog, AgentCatalogRequest,
+  VendorDescription, AccountDescription, AgentCatalog,
   AppUiContext, GatekeeperUser, GatekeeperUiFrame, ApprovalQueue, ObservationAuthorizer,
   GatekeeperConnectCallback, GatekeeperConnectOptions, SupportedResource,
   Gatekeeper, GatekeeperUserVerifier, ResourceDescription, ActionKind,
@@ -18,8 +17,7 @@ import { ContextApiImpl, loadEnabledContextCollections } from "./context-api.js"
 import { ContextObserverTracker } from "./context-observers.js";
 import type { ContextVerifierApi } from "./context-observers.js";
 import {
-  buildAgentSkillCatalogEntries, buildAgentSkillCommands, buildAgentSkillMessage,
-  parseSkillManifest,
+  buildAgentSkillCommands, buildAgentSkillMessage, buildContextCatalog, parseSkillManifest,
   type CollectionSkills,
 } from "./agent-skill.js";
 import type { EnabledCollectionInfo } from "./context-types.js";
@@ -318,25 +316,12 @@ export class ContextGatekeeper
   }
 
   async getAgentCatalog(
-      request: AgentCatalogRequest,
       authorizer: NativeRpcStub<ObservationAuthorizer>): Promise<AgentCatalog> {
     let domain = this.ctx.props.sharingDomain;
     let userLibrary = this.#userLibraries().get(
       this.#userLibraries().idFromName(domainName(domain, this.ctx.props.accountId)));
     let collections = await loadEnabledContextCollections(this.env, domain, userLibrary);
-    let loaded = await this.#loadSkills(collections);
-    let skillEntries = buildAgentSkillCatalogEntries(loaded);
-    let collectionEntries = collections
-        .map(collection => ({
-          id: collection.id,
-          title: collection.title,
-          description: collection.description,
-        }))
-        .toSorted((left, right) =>
-          left.title.localeCompare(right.title) || left.id.localeCompare(right.id));
-    let entries = [...skillEntries, ...collectionEntries].toSorted((left, right) =>
-      left.title.localeCompare(right.title) || left.id.localeCompare(right.id));
-    let catalog = boundAgentCatalog(entries, request);
+    let catalog = buildContextCatalog(collections, await this.#loadSkills(collections));
     if (catalog.entries.length > 0) {
       let collectionIds = [...new Set(catalog.entries.map(entry => {
         let slash = entry.id.indexOf("/");
--- a/packages/workshop-shared/src/gatekeeper.ts
+++ b/packages/workshop-shared/src/gatekeeper.ts
@@ -103,44 +103,44 @@ export type AgentCatalogEntry = {
 
 /** The discovery metadata returned for one gatekeeper session. */
 export type AgentCatalog = {
-  /** The discoverable items, already truncated to the requested/maximum count. */
+  /**
+   * The discoverable items, in the gatekeeper's priority order: the Workshop clamps the list by
+   * dropping from the tail, so entries that must survive belong first.
+   */
   entries: AgentCatalogEntry[];
-  /** True if entries were dropped to fit the limit, so the agent knows the list is partial. */
+  /** True if entries were dropped to fit the caps, so the agent knows the list is partial. */
   truncated?: boolean;
-};
-
-/** Parameters the Workshop passes when requesting a catalog. */
-export type AgentCatalogRequest = {
-  /** Maximum number of entries to return. The gatekeeper must also enforce AGENT_CATALOG_MAX_ENTRIES. */
-  limit: number;
 };
 
 /**
  * Hard caps the Workshop enforces on any catalog, regardless of what the gatekeeper returns, since
  * the catalog is injected into the agent's context as untrusted data and must stay bounded.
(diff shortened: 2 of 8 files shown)


AshishKumar4:

packages/gatekeeper-context/src/library-gatekeeper.ts
```diff
@@ -334,9 +354,11 @@ export class ContextGatekeeper
         }))
         .toSorted((left, right) =>
           left.title.localeCompare(right.title) || left.id.localeCompare(right.id));
-    let entries = [...skillEntries, ...collectionEntries].toSorted((left, right) =>
+    
+    let entries = [...boundedSkillsEntries.entries, ...collectionEntries].toSorted((left, right) =>
       left.title.localeCompare(right.title) || left.id.localeCompare(right.id));
-    let catalog = boundAgentCatalog(entries, request);
+
+    let catalog = { entries: entries, truncated: boundedSkillsEntries.truncated }
```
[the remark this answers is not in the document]
AshishKumar4: I am still working on the PR bro

packages/gatekeeper-context/src/library-gatekeeper.ts
```diff
@@ -323,9 +329,23 @@ export class ContextGatekeeper
...
+
     let collections = await loadEnabledContextCollections(this.env, domain, userLibrary);
+
     let loaded = await this.#loadSkills(collections);
-    let skillEntries = buildAgentSkillCatalogEntries(loaded);
+    let skillsEntries = buildAgentSkillCatalogEntries(loaded)
+
+    // Let's only apply catalog limits to skills entries and not the whole catalog. 
+    // The number of collections would be far lower than the total number of skills which would be the problem.
+    const boundedSkillsEntries = boundAgentCatalog(skillsEntries, request);
```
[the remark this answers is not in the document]
AshishKumar4: I haven't pushed the changes bro

packages/gatekeeper-context/src/library-gatekeeper.ts
```diff
@@ -323,9 +329,23 @@ export class ContextGatekeeper
...
+    if (boundedSkillsEntries.truncated) {
+        logger.warn('Skills were truncated to fit in the limit', {
```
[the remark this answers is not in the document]
AshishKumar4: Alright

packages/gatekeeper-context/src/library-gatekeeper.ts
```diff
@@ -334,9 +331,11 @@ export class ContextGatekeeper
         }))
         .toSorted((left, right) =>
           left.title.localeCompare(right.title) || left.id.localeCompare(right.id));
-    let entries = [...skillEntries, ...collectionEntries].toSorted((left, right) =>
-      left.title.localeCompare(right.title) || left.id.localeCompare(right.id));
-    let catalog = boundAgentCatalog(entries, request);
```
Maximo-Guk, line 339: I believe if we're no longer bounding the catalog before RPC, a gatekeeper who has a large library can make `getAgentCatalog()` exceed RPC limits, before the backend clamp runs, resulting in no catalog.

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -103,47 +103,27 @@ export type AgentCatalogEntry = {
...
-  /** Maximum number of entries to return. The gatekeeper must also enforce AGENT_CATALOG_MAX_ENTRIES. */
-  limit: number;
-};
-
 /**
  * Hard caps the Workshop enforces on any catalog, regardless of what the gatekeeper returns, since
- * the catalog is injected into the agent's context as untrusted data and must stay bounded.
+ * the catalog is injected into the agent's context as untrusted data and must stay bounded. The
+ * entry count is what the context cost scales with: the catalog is inlined in the system prompt on
+ * every turn and compaction never reaches it, so at these field caps 200 entries is already ~80 KB.
```
Maximo-Guk: Are we sure these numbers are accurate? From my agent:

Medium: The prompt-size estimate understates the maximum by roughly 2x. gatekeeper.ts:115-122  
At the documented field caps, 200 ASCII entries serialize to about 159 KB, not 80 KB, roughly 40K tokens before other system context. The cap should be based on a tested serialized/token budget.

packages/gatekeeper-context/src/library-gatekeeper.ts
```diff
@@ -334,9 +331,11 @@ export class ContextGatekeeper
...
+    // Collections first: they are the agent's entry points into the library, and the Workshop
+    // clamps by dropping from the tail, so hundreds of skills can't push them out (skills past
+    // the cap stay reachable via the session's list()/search()). The merged list is left
+    // unsorted: the Workshop sorts the survivors, so sorting here would only pick the losers.
+    let catalog = {entries: [...collectionEntries, ...buildAgentSkillCatalogEntries(loaded)]};
```
Maximo-Guk: From GPT: 

Medium: Hidden collections are still authorized as observed. library-gatekeeper.ts:338-350  
With over 200 collections, Context authorizes and commits all collections before the backend drops the tail. A collaborator lacking a dropped private collection can block the whole catalog or later be denied access despite that collection never reaching the agent. Authorization should cover only returned entries.

packages/gatekeeper-context/src/library-gatekeeper.ts
AshishKumar4: Technically true but the RPC limits are for 32mb of data iirc. And we can apply soft limits on our curated gatekeepers but it won't really be much of a guaranteed limit. 
If it's needed, we can apply a generous cap on the data returned - maybe of 16mb worth? Based on length of combined string ~ number of bytes?

Maximo-Guk: Aren't we already capping on the other end though, couldn't we apply the same cap here?

packages/workshop-shared/src/gatekeeper.ts
AshishKumar4: Well, 200x(100+400) = 100kb and although sure, that maybe true, these are worst case numbers. 
We do know that we do need to ensure all the context collections are visible to the agent, and then the remaining skills can go in there, which we currently have 97+ of.

packages/gatekeeper-context/src/library-gatekeeper.ts
AshishKumar4: Done in the PR - we now bound the internal response too with the same utility - boundAgentCatalog, after bounding the number of skills first to a lower value

packages/gatekeeper-context/src/library-gatekeeper.ts
AshishKumar4: Fixed by bounding the skills catalog separately to a lower value, and then bounding the total to a higher threshold of 1000

kentonv (approved):

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -103,44 +103,44 @@ export type AgentCatalogEntry = {
...
  * the catalog is injected into the agent's context as untrusted data and must stay bounded.
+ *
+ * The entry count is a ceiling, not a budget. At the field caps below one entry serializes to about
+ * 793 ASCII bytes, so 1000 entries is ~775 KiB of prompt, and the catalog sits in the system prompt
+ * on every turn where compaction never reaches it. A gatekeeper is expected to return far fewer than
+ * the ceiling and to bound whichever of its item classes can grow without limit (the Context Library
+ * caps its skills), leaving this as the backstop against one that doesn't.
  */
-export const AGENT_CATALOG_MAX_ENTRIES = 25;
+export const AGENT_CATALOG_MAX_ENTRIES = 1000;
```
Maximo-Guk, line 125: Another one from GPT:

packages/workshop-shared/src/gatekeeper.ts:119-125 permits roughly 775 KiB ASCII or 2.3 MiB UTF-8 in the permanent system prompt. This can exceed the 128k-token fallback window (agent-compaction.ts:19-21), and compaction cannot remove it. A large Context library or untrusted gatekeeper could make every agent turn fail. Enforce a safe serialized/token budget rather than 1,000 maximum-sized entries

Maximo-Guk (approved):

packages/workshop-shared/src/gatekeeper.ts
AshishKumar4: That is an upper cap; We are capping the skills catalog to a much lower value which would be the ones that tend to have big descriptions. Thus most of the 1k would be catalogs and small descriptions/titles. 
We can try to narrow down the thresholds in the future as we have more information, but premature optimizations can bite us just like it did here with the original threshold of 25.
And over the long term, the ideal solution would be a heirarchial catalog or discovery IF the number of collections truly become that big

AshishKumar4:
bonk break glass

PR #213, opened 2026-08-14
Name the document an invoked Agent Skill came from

AshishKumar4:
## What does this change?

Currently, for a skill invoked via slash commands, the agent never sees the actual document ID/path, and thus any relative files/resources referred in the skill also don't make sense to the agent. This PR fixes that.

Just simply adds the docId to the prompt

## Checklist

### DIFF
--- a/packages/gatekeeper-context/src/agent-skill.ts
+++ b/packages/gatekeeper-context/src/agent-skill.ts
@@ -2,7 +2,7 @@ import { parse as parseYaml } from "yaml";
 import { z } from "zod";
 import type { SlashCommandDescriptor } from "@gadgets/workshop-shared/gatekeeper";
 import type { EnabledCollectionInfo } from "./context-types.js";
-import { encodeDocId } from "./context-types.js";
+import { docIdRoot, encodeDocId } from "./context-types.js";
 
 const AGENT_SKILL_NAME_MAX_LENGTH = 64;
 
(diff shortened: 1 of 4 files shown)


packages/gatekeeper-context/src/agent-skill.ts
```diff
@@ -63,11 +63,18 @@ export function buildAgentSkillCatalogEntries(
...
+ * document the skill came from, because otherwise the paths the skill cites resolve to nothing. The
+ * name is prose rather than an element attribute since a document path may contain quotes.
  */
-export function buildAgentSkillMessage(content: string, args: string): string {
+export function buildAgentSkillMessage(docId: string, content: string, args: string): string {
   let usesArgument = /\$ARGUMENT(?![A-Za-z0-9_[])/.test(content);
   let expanded = content.replace(/\$ARGUMENT(?![A-Za-z0-9_[])/g, () => args);
-  let message = `<agent_skill>\n${expanded}\n</agent_skill>`;
+  let message = `<agent_skill>\n${expanded}\n</agent_skill>\n\n` +
+    `skill root: ${docId.slice(0, docId.lastIndexOf("/") + 1)} — read the documents it references ` +
```
Maximo-Guk: I think it would be worth pulling this out in to a helper something like a docIdRoot(docId) next to the existing helpers https://github.com/cloudflare/cloudflare-os/blob/9ba6a72db0e30906f37be76c592b2c6d3b14d17b/packages/gatekeeper-context/src/context-types.ts#L76

Maximo-Guk (approved):

packages/gatekeeper-context/src/agent-skill.ts
AshishKumar4: Done — added `docIdRoot()` beside `encodeDocId`/`decodeDocId` in `context-types.ts` and used it here.

Agreed on the reasoning: that module owns the `<collectionId>/<path>` grammar, so a hand-rolled `lastIndexOf("/")` in `agent-skill.ts` was a second place encoding the same format, free to drift from `encodeDocId`.

No dedicated test — the four `buildAgentSkillMessage` tests assert the complete output string including the root, so a regression in the helper fails them immediately. Happy to add a direct one if you would rather have it pinned separately.

AshishKumar4:
bonk break glass

PR #221, opened 2026-08-15
Fix oxlint plugin test on Windows

### LINKED ISSUE #220: Oxlint plugin test fails on Windows before running
## Summary

On Windows, the oxlint plugin test exits before any `RuleTester` cases run because its dynamic `import()` receives an absolute filesystem path.

## Reproduction

1. Install dependencies with pnpm.
2. Run `node --test scripts/oxlint-plugin.test.js` on Windows.

The process fails with:

```text
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
```

I reproduced this with Node 22.22.3 and pnpm 11.17.0.

## Cause

`vitePlusRequire.resolve("oxlint/plugins-dev")` returns an absolute filesystem path. On Windows, passing that path directly to `import()` makes the drive letter look like a URL scheme.

Converting the resolved path with Node's `pathToFileURL()` before importing allows all 32 existing test cases to run and pass, while remaining valid on POSIX platforms.

AI tools assisted the investigation and drafting. The failure and passing test result above were verified in the stated Windows environment.

snowyukitty:
## What does this change?

Fixes #220.

Convert the absolute path returned by `require.resolve()` to a `file:` URL before dynamically importing Vite+'s pinned oxlint plugin test helper. This lets the test file start on Windows, where a raw drive-letter path is not a valid ESM specifier.

`pathToFileURL()` is Node's standard conversion from a filesystem path to an importable file URL. The resolved module and all test behavior remain unchanged; only the representation passed to `import()` changes. The patch adds one standard-library import and wraps the existing resolved path.

Verification on Windows with Node 22.22.3 and pnpm 11.17.0:

- `node --test scripts/oxlint-plugin.test.js` — 32 passed
- `pnpm lint:check` — passed (existing warnings only)
- `pnpm lint` — lint passed, but the recursive type/build phase hit the existing `spawnSync pnpm ENOENT` failure in unchanged `packages/gatekeeper-scheduler/build-app.mjs` tracked by #19 and #107

AI tools assisted the investigation, patch drafting, and review. The failure and verification results above were reproduced locally.

## Checklist

### DIFF
--- a/scripts/oxlint-plugin.test.js
+++ b/scripts/oxlint-plugin.test.js
@@ -1,11 +1,14 @@
 import { describe, it } from "node:test";
 import { createRequire } from "node:module";
+import { pathToFileURL } from "node:url";
 import { preferJsdoc } from "./oxlint-plugin.mjs";
 
 const require = createRequire(import.meta.url);
 const vitePlusRequire = createRequire(require.resolve("vite-plus/package.json"));
 // Test against Vite+'s pinned oxlint without adding a second direct dependency that can drift.
-const { RuleTester } = await import(vitePlusRequire.resolve("oxlint/plugins-dev"));
+const { RuleTester } = await import(
+  pathToFileURL(vitePlusRequire.resolve("oxlint/plugins-dev")),
+);
 
 RuleTester.describe = describe;
 RuleTester.it = it;


Maximo-Guk:
Thank you! https://github.com/cloudflare/cloudflare-os/pull/235 should address the remaining Windows gaps

PR #224, opened 2026-08-15
Add Cloudflare Workers observability to Cloudflare gatekeeper

ndisidore:
Adds account- and Worker-scoped observability sessions with `listKeys()` and `listValues()` for discovery, `listEvents()`, `listInvocations()`, `listTraces()`, and `getTrace()` for telemetry inspection, and `calculate()` for aggregate metrics.

Screenshot from 2026-08-15 11-20-51
Screenshot from 2026-08-15 11-22-08
Screenshot from 2026-08-15 10-45-49
Screenshot from 2026-08-15 10-46-02


Maximo-Guk:
Small nit: 

It might be useful to document the whole Grant type + Token Authentication method etc in the README.md of the Cloudflare Gatekeeper ( preexisting prior to your change, but probably still a good idea )
image

Maximo-Guk:
Few GPT comments:

```
1. Medium: observability-api.ts:396 replaces the provider’s total match count with the current page length for Worker bindings. A query with 250 matches and limit: 100 incorrectly reports count: 100.
2. Medium: cloudflare-configurators.ts:34 uses the non-paginated /accounts helper. Cloudflare returns 20 accounts by default, so later accounts cannot be selected or searched.
3. Medium: observability-api.ts:397,427 derives continuation cursors after filtering foreign events. If a provider page contains only foreign events, pagination ends even when valid later results exist.
4. Medium: observability-api.ts:223-230,299-301 logs provider error messages verbatim. Cloudflare may echo caller-supplied filter or search values, potentially placing sensitive telemetry queries in server logs.
```

Maximo-Guk:
Can't leave comments on diff ( due to github status ):

But don't see why we're changing this ( maybe rebase got wrong ? )

image

Also looks like we're lowering dependency versions in a lot of places 

image

ndisidore:
bonk break glass

AGENTS.md
```diff
@@ -30,6 +30,14 @@ The project structure is:
...
 * packages/gatekeeper-scheduler: Scheduled Tasks — an auto-provisioned gatekeeper whose account provides an ambient singleton for registering persistent workspace callbacks plus a read-only management UI. One account-scoped `ScheduleDriver` Durable Object stores enabled schedules and delivers them from a shared alarm; hook enablement remains in the Workshop Connections UI.
+* packages/gatekeeper-cloudflare: Cloudflare OAuth, serving three unrelated purposes from one connected account — sign-in (`AUTH_GATEKEEPERS`), AI Gateway billing, and **Workers Observability** read-only telemetry resources. Two resource granularities (whole account, or one Worker) both map to the single indivisible `workers-observability.read` scope, so the capability boundary is the *binding*, not the grant.
```
Maximo-Guk, line 33: I know we've been dumping a lot of stuff in AGENTS.md, but I wonder if maybe we should prune/clean this up/make sure we're not putting a bunch of stuff in here that agent's can't just figure out on their own to avoid filling up context for everyone

Maximo-Guk (approved):

Maximo-Guk:
Few more GPT comments:

1 High: Cloudflare sign-in accounts never publish later observability grants. cloudflare.ts:304-307 calls the original login callback, whose credentialsRestored() is a no-op (auth/login-flow.ts:146-154). The persisted account description remains billing-only, so UIs repeatedly request authorization instead of enabling the resource.
2. High: Worker-scoped aggregate queries are not defensively scoped. observability-api.ts:598-605 returns calculations unchanged, while only event arrays are re-filtered (:373-381). If Cloudflare ignores the injected service filter, calculate() and provider counts/statistics disclose account-wide telemetry.
3. Medium: Changing accounts retains the previously selected Worker. cloudflare-worker-configurator-ui.tsx:27-30 calls clearFields("workerName"), which only clears autocomplete query state, not values.workerName. This can create an invalid account/Worker binding.
4. Medium: Calculation responses remain blindly cast. observability-parse.ts:185-193 accepts { calculation: "count" } despite required aggregates and series. The workerd test does exactly this, so malformed provider responses escape the advertised trust boundary.

ndisidore:
bonk break glass

PR #233, opened 2026-08-17
Convert all scripts/ to typescript

Maximo-Guk:
[Running TypeScript Natively](https://nodejs.org/learn/typescript/run-natively#running-typescript-natively)

Rather than having a bunch of untyped scripts, we can go ahead and convert all of them to typescript to align with the rest of the codebase. We don't even need a build step, simply by adding [erasableSyntaxOnly](https://github.com/cloudflare/cloudflare-os/pull/231/changes#diff-bb0ef161f0c803b8b038faab449fbbec5024e7e85760cf4e6a6d9179f7e4c0eaR17) to our tsconfig, and using node >= 22.18.0

PR #235, opened 2026-08-17
Fix nested pnpm spawns on Windows

### LINKED ISSUE #19: pnpm run-local cannot start on Windows: spawnSync pnpm ENOENT

### LINKED ISSUE #120: Windows: pnpm run-local fails with spawnSync pnpm ENOENT
I am attempting to run Cloudflare OS locally from the official repository:

`cloudflare/cloudflare-os`

The repository clones successfully, but the documented local startup command fails when `scripts/run-local.mjs` attempts to execute `pnpm install`.

## Environment

* **OS:** Windows 10/11
* **Shell:** PowerShell
* **Repository:** `cloudflare/cloudflare-os`
* **Branch:** `main`
* **Node.js:** `v26.4.0`
* **pnpm:** installed and available from PowerShell
* **Working directory:**

```text
C:\Users\user\Downloads\cloudflare-os
```

## Steps to reproduce

```powershell
cd C:\Users\user\Downloads\cloudflare-os
pnpm run-local
```

The command successfully resolves `pnpm run-local` and starts:

```text
$ node scripts/run-local.mjs
```

The script then attempts to run:

```text
pnpm install
```

but fails with:

```text
node:internal/child_process:1144
result.error = new ErrnoException(result.error, 'spawnSync ' + options.file);
^

Error: spawnSync pnpm ENOENT
    at Object.spawnSync (node:internal/child_process:1144:20)
    at spawnSync (node:child_process:912:24)
    at execFileSync (node:child_process:955:15)
    at run (file:///C:/Users/user/Downloads/cloudflare-os/scripts/run-local.mjs:118:3)
    at file:///C:/Users/user/Downloads/cloudflare-os/scripts/run-local.mjs:122:3 {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawnSync pnpm',
  path: 'pnpm',
  spawnargs: [ 'install' ],
  status: null,
  signal: null,
  output: null,
  pid: 0,
  stdout: undefined,
  stderr: undefined
}

Node.js v26.4.0
[ELIFECYCLE] Command failed with exit code 1.
```

## Observed behavior

`pnpm` is available directly from PowerShell, but the Node.js child process created by `scripts/run-local.mjs` cannot resolve/spawn `pnpm`.

In other words:

```text
PowerShell → pnpm → works
Node child_process.spawnSync("pnpm") → ENOENT
```

This appears to be related to Windows executable resolution/PATH handling, potentially involving `pnpm.cmd`.

## Expected behavior

Running:

```powershell
pnpm run-local
```

should execute the repository's local setup process, including:

```text
pnpm install
```

and subsequently start the local Cloudflare OS development environment.

## Diagnostics

The pnpm-related PATH entries on the system include:

```text
C:\Users\user\AppData\Local\pnpm\bin
C:\Users\user\AppData\Roaming\npm
```

I would like to confirm whether the current `scripts/run-local.mjs` implementation is expected to work on Windows with this pnpm installation.

## Questions

Could the maintainers clarify:

1. **Supported Node.js versions**

   * Is Node.js `v26.4.0` currently supported?
   * Is Node.js 22 LTS the recommended version for the current `main` branch?

2. **Supported pnpm version**

   * Which pnpm version is currently expected/recommended?

3. **Windows support**

   * Is `pnpm run-local` currently expected to work on Windows?
   * Are there any documented Windows-specific prerequisites?

4. **Executable resolution**

   * Should `scripts/run-local.mjs` invoke `pnpm.cmd` on Windows instead of `pnpm`?
   * Is there a recommended PATH configuration for pnpm on Windows?

5. **Alternative local setup**

   * If Windows is not currently supported for the local workflow, is WSL2 the recommended approach?

## Potential workaround

I have not modified the repository yet because I wanted to confirm the officially supported environment before changing the startup script or switching Node.js versions.

If Windows + Node 22 LTS + pnpm 10 is the recommended configuration, I can test that combination and provide the resulting output.

Thanks for maintaining Cloudflare OS and for any guidance on the expected Windows development environment.

Maximo-Guk:
Fixes #19, fixes #120. Supersedes #107 by @Sainigurnoor511, whose `npm_execpath` approach this builds on. Stacked on #233.

`pnpm run-local` and `pnpm dev-server` cannot start on Windows: `pnpm` on PATH is a `.cmd` shim, which Node cannot spawn without a shell

Changed it so that the `pnpm exec` sites now resolve their binary directly, only falling back to pnpm if the binary is not found. We were actually already using this approach for the `run-dev-server` file to reduce overhead from invoking pnpm ( ~0.33 s of overhead compared to invoking node binary directly ).

Verified on my Windows 11 VM.


ndisidore (approved):

Maximo-Guk:
bonk break glass

PR #236, opened 2026-08-17
Tell the agent how to use catalogs and where outputs go

AshishKumar4:
## What does this change?

Two things the agent is given but never told how to use.

**Outputs.** `writeFile` takes a `workpiece`, so every write lands inside one specific Gadget as one
of its source files. There is no free-standing file path, and nothing in the prompt says so. When an
instruction says "save it to `notes.md`" the agent has no defined behaviour: sometimes it answers in
chat, sometimes it writes a `.md` into a Gadget's source tree where it renders as code rather than a
document, sometimes it creates the right output. This states the rule as its own paragraph in the
section that already explains the deployment's formats, directly above the list it refers to.

**Catalogs.** A resource's discovery catalog is inlined into the system prompt, and `truncated` is
emitted as a bare JSON key with nothing explaining it. The agent is told to `describeBinding` a
resource when one is relevant, but never to consult the entries first, and never that a partial list
has a remainder reachable through the binding's own API. So a listed item goes unused unless the user
names it, and a truncated list looks complete.

Prompt text only: 13 added lines across two functions, no code path changes. The complete effect is
the wording, read in place.

## Checklist

### DIFF
--- a/packages/workshop-backend/__tests__/agent-catalog.test.ts
+++ b/packages/workshop-backend/__tests__/agent-catalog.test.ts
@@ -80,6 +80,17 @@ describe("normalizeAgentCatalog", () => {
     // A resource with no catalog gets just its env line, nothing inlined.
     expect(message).toContain("- Empty: `env.EMPTY`");
     expect(message).not.toContain("- Empty: `env.EMPTY`\n{");
+    expect(message).toContain("Check the entries listed above before starting a task");
+    expect(message).toContain('A catalog marked `"truncated":true` is partial');
+  });
+
+  it("describes the catalogs only when a resource published entries", () => {
+    let message = formatAlwaysAvailableResourcesPrompt([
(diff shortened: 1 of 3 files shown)


AshishKumar4:
Closing — #208 removed the basis for both halves of this.

The catalog legend explained `"truncated":true`, but #208 landed the cap at 1000. The library is 116 skills + 1 collection = 117 entries, so that flag can no longer fire for this deployment. Documenting a field that cannot appear isn't worth prompt space.

The output-path paragraph doesn't clear the bar either, as discussed: the section above it already tells the agent to instantiate a format rather than build an equivalent, the motivation came from an external skill corpus that doesn't live in this repo, and there's no eval here to show the agent actually gets it wrong. Arguing from a plausible failure mode isn't enough for the most contended text in the product.

PR #244, opened 2026-08-17
Honour break glass on re-runs

ndisidore:
`bonk break glass` posts a passing check and labels the PR. The label was read from the event payload, and a re-run replays the payload captured before the label existed — so a bypassed PR got reviewed anyway and went red again. It now reads the label from the API, which is what makes the bypass survive a re-run.

Waiting for in-flight reviews is now a re-listing loop against one deadline, so the passing check is never posted while a run is still live and able to bury it.

Not attempted: patching the failing check run. A workflow token lost that ability on [2025-03-31](https://github.blog/changelog/2025-02-12-notice-of-upcoming-deprecations-and-breaking-changes-for-github-actions/#changes-to-check-run-status-modification).

Also adds a concurrency group, attributes the bypass to the commenter, and rejects re-runs of the break-glass job, which keep the original actor's privileges but re-resolve the head SHA.

### DIFF
--- a/.github/workflows/bonk-pr.yml
+++ b/.github/workflows/bonk-pr.yml
@@ -9,7 +9,11 @@ on:
 jobs:
   # Break-glass invariants:
   # - The main ruleset requires `bonk` from GitHub Actions app 15368.
-  # - Labeled PRs run a successful step so future head SHAs also pass `bonk`.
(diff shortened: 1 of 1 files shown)


ndisidore:
bonk break glass

PR #252, opened 2026-08-18
Fix frontend build cleanup on Windows

### LINKED ISSUE #250: workshop-frontend build cannot run its clean task on Windows
## Summary

On Windows, the `workshop-frontend` Vite+ build task cannot get past its `clean:dist` dependency because that task asks Vite+ to execute the POSIX `rm` binary directly.

This also prevents the repository's required `pnpm lint` command from completing on Windows.

## Reproduction

At `3562627` on Windows 11, Node 24.19.0, and pnpm 11.17.0:

```text
> pnpm exec vp run -F @gadgets/workshop-frontend --no-cache build

error: Failed to find executable rm under cwd <checkout>\packages\workshop-frontend
* cannot find binary path
```

The same failure occurs when the frontend task is reached through `pnpm lint`.

## Cause

#212 added this Vite+ task:

```ts
'clean:dist': { command: 'rm -rf dist', cache: false },
```

Vite+ resolves the task command as an executable, and a normal Windows environment has no `rm` binary. The older `clean` package scripts are different: package-manager scripts go through their platform shell, while this task is launched by Vite+ directly.

## Verified fix

Using Node's built-in filesystem API preserves the recursive, missing-path-tolerant cleanup without requiring a platform-specific executable:

```ts
command: `node -e "require('node:fs').rmSync('dist', { recursive: true, force: true })"`,
```

With that change on the same checkout:

- the targeted frontend build completed, including both TypeScript checks and the production Vite bundle;
- a stale sentinel placed under `dist/` before the run was removed by `clean:dist`;
- the full `pnpm lint` command completed successfully (existing warnings only).

AI tools assisted the investigation and drafting. The failure and verification results above were reproduced locally.

snowyukitty:
## What does this change?

Fixes #250.

Replace the POSIX-only `rm` executable in the `workshop-frontend` Vite+ `clean:dist` task with Node's built-in recursive filesystem removal. This lets the frontend build, and therefore `pnpm lint`, run on Windows without changing the task's cleanup behavior.

`fs.rmSync(..., { recursive: true, force: true })` has the two properties this task needs from `rm -rf`: it removes the complete `dist` tree and does not fail when the path is absent. Node is already the runtime for every repository task, so the replacement adds no dependency or platform-specific executable. The task remains uncached and remains the `build` dependency; only its command changes.

Verification on Windows 11 with Node 24.19.0 and pnpm 11.17.0:

- before: `pnpm exec vp run -F @gadgets/workshop-frontend --no-cache build` failed with `Failed to find executable rm`;
- after: the same command completed both TypeScript checks and the production Vite build;
- a stale sentinel placed under `dist/` before the run was removed;
- `pnpm lint` completed successfully (existing warnings only).

AI tools assisted the investigation, patch drafting, and review. The failure and verification results above were reproduced locally.

## Checklist

### DIFF
--- a/packages/workshop-frontend/vite.config.ts
+++ b/packages/workshop-frontend/vite.config.ts
@@ -14,7 +14,10 @@ const ownDist = { pattern: '!dist/**', base: 'package' } as const
 const runConfig = {
   run: {
     tasks: {
-      'clean:dist': { command: 'rm -rf dist', cache: false },
+      'clean:dist': {
+        command: `node -e "require('node:fs').rmSync('dist', { recursive: true, force: true })"`,
+        cache: false,
+      },
       /**
        * `build` is a task rather than a package.json script so `env` can declare the `VITE_*` flags
        * it reads: a cached `vp` run executes scripts in a clean environment, and the values would be


snowyukitty:
Superseded by #373: `main` now uses `fs.rmSync` with `recursive: true` and `force: true` for `clean:dist`, covering the Windows cleanup failure addressed here. This change is already present upstream; the original issue #250 remains open.

PR #256, opened 2026-08-18
fix: file export follow-ups

maxwellpeterson:
Validating https://github.com/cloudflare/cloudflare-os/pull/197 before release turned up a few issues, which this PR fixes:
- exportGadget incorrectly used .move() instead of .dup()
- restore 250ms DOM settlement window to avoid breaking PDF export for existing gadgets that don't implement top-level await
  - new gadgets should use top-level await by default, but existing gadgets rely on DOM settlement logic for PDF export

### DIFF
--- a/packages/workshop-backend/browser/browser-export-page.ts
+++ b/packages/workshop-backend/browser/browser-export-page.ts
@@ -39,6 +39,28 @@ export async function waitForClientModule(): Promise<void> {
 
 // Functions that run in the isolated realm:
 
+/** Waits until the rendered DOM has remained unchanged for the requested interval. */
+export function waitForDomSettled(quietMs: number): Promise<void> {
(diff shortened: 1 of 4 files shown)


packages/workshop-backend/browser/browser-export-page.ts
```diff
@@ -39,6 +39,28 @@ export async function waitForClientModule(): Promise<void> {
 
 // Functions that run in the isolated realm:
 
+/** Waits until the rendered DOM has remained unchanged for the requested interval. */
```
maxwellpeterson, line 42: This function was originally included in #147 and then removed in favor of just top-level await. We actually want both.

ndisidore (approved):

maxwellpeterson:
bonk break glass

PR #258, opened 2026-08-18
ci: temporarily disable Bonk workflows

ndisidore:
Temporarily disables Bonk workflows until they are confirmed working.


Maximo-Guk (approved):

ndisidore:
bonk break glass

PR #260, opened 2026-08-18
Carry the route and signed-in user on frontend error reports

ndisidore:
Frontend error reports now carry `pageLocation` and `reportedUserId`, so triage doesn't start from a bare stack trace.

Neither is trusted. `POST /api/client-errors` has no credential, so the identity is whatever the client claimed — hence `reportedUserId` rather than `userId`: it's a report, not a finding, and nothing may read it to make a decision. AGENTS.md and REVIEW.md previously said reports "never convey identity or authority"; both are updated.

`pageLocation` is origin and pathname only, and `normalizeFrontendErrorReport` strips query and fragment at the boundary rather than trusting producers. That's the bit worth a close read: a share link's fragment is a bearer capability, so `href` instead of `pathname` would leak one into a third-party issue tracker.

On the frontend, identity is an injected getter rather than module state read inside the reporter, and it deliberately survives the stub swap a reconnect causes — clearing there would blank the identity for the whole `whoami` round trip.

Three commits so `error-reporting`/`workshop-backend` review apart from the UI; the first stands alone.


packages/workshop-frontend/src/errorReporting.ts
```diff
@@ -36,6 +38,27 @@ type BrowserReporter = Readonly<{
...
+ * A share link carries a bearer capability in its fragment, so a report must never carry one out
+ * of the tab. A failure resolves to '' rather than throwing, because the caller's catch would
+ * otherwise discard an entire report over a single missing field.
+ */
+function getPageLocation(): string {
+  try {
+    const url = new URL(window.location.href)
+    url.search = ''
+    url.hash = ''
+    return url.href
```
Maximo-Guk: I don't think this is that big of an issue, but might as well reconstruct it probably:


High: errorReporting.ts:56 preserves URL credentials. https://user:secret@host/path reaches the external Reporter because the backend normalizer at index.ts:158 only removes query/fragment. Reconstruct from parsed origin + pathname and reject non-HTTP(S) URLs.

packages/workshop-frontend/src/AuthContext.tsx
```diff
@@ -26,11 +27,18 @@ export function AuthProvider({ children, authenticatedApi, onLogout }: AuthProvi
   useEffect(() => {
     let cancelled = false
     authenticatedApi.whoami().then((info) => {
-      if (!cancelled) setCurrentUser(info)
+      if (cancelled) return
+      setCurrentUser(info)
+      // Only a real user account names a person: for a gadget author `id` is its owner's id.
+      if (info.type === 'user') setErrorReportingUserId(info.id)
```
Maximo-Guk: Medium: AuthContext.tsx:33 only sets identity inside AuthProvider. Inline login on /blueprint/:id uses a separate useAuth instance and never mounts that provider, so subsequent reports omit userId. Identity tracking should occur in shared authentication state or every successful auth path.

Maximo-Guk (approved):

packages/workshop-frontend/src/AuthContext.tsx
ndisidore: Validated and fixed in 42a3bf53 — good catch, this was a real hole.

Confirmed the mechanism end to end. `__root.tsx:40` computes `standalone = isSignup || (isBlueprint && !isAuthenticated)` and the branch at :94-104 renders the `Outlet` bare, so a signed-out visitor to `/blueprint/:id` gets no provider. `BlueprintLandingPage` holds its own `useAuth` instance and logs in inline at :487-493, and `useAuth` is per-instance `useState` with no shared store — so the root's instance never learns of it, `standalone` stays true, and no provider ever mounts. The gap runs until `handleCreate` does `window.location.href` (:570), i.e. across binding configuration and gadget create, which is where the interesting failures are.

Took your first option: the claim moved into `useAuth`, which is the one point all authentication converges on — both the token and CF Access paths mint their stub there. `AuthContext.tsx` is now byte-identical to base, untouched by this branch.

One thing that fell out of the move, worth flagging since it cuts against the previous design: clearing is now on **logout only, never on unmount**. Concurrent instances are reachable via `useAuth` in a way they weren't via the provider — the blueprint page nests its own inside the root's when a *signed-in* user views that page — so clearing on unmount would let navigating away blank an identity the root still holds.

Cost is one extra pipelined `whoami`, since `AuthProvider` still makes its own for `currentUser`. Deduping means threading `currentUser` through the provider's props, which is an auth refactor I'd rather not fold into an observability change.

`AuthContext.test.tsx` became `useAuth.test.tsx` (7 tests). Both mutations are covered: dropping the claim fails the inline-login test, and adding an unmount clear fails the concurrent-instance test.

PR #263, opened 2026-08-18
Retry pure user-DO reads once across a reset

ndisidore:
#133 and #162 made user-DO resets survivable by minting a fresh stub per call, but deliberately left retries out until telemetry showed they were worth adding. The `user_do.reset.surfaced` volume has now made that case: the one call in flight at the moment the DO resets still fails all the way to the browser.

This change retries that call once. A reset can't tell the caller whether the failed call was applied or not, so only calls that are safe to replay may retry. It requires an explicit `retryOnDoReset(() => this.#user.x())` wrapper around strictly pure reads. The thunk re-evaluates the fresh-stub getter, so the second attempt reaches the new incarnation; errors are never re-wrapped, so the workerd flags the browser's classifier reads stay intact.

Successful retries log `user_do.reset.recovered`, which correlated against `user_do.reset.surfaced` gives the retry win rate and tells us whether more sites are worth wrapping.

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/263" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/workshop-backend/src/do-retry.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * True for rejections caused by a DO reset or lost connection. This is the telemetry
+ * classification, deliberately not the retry policy, which is narrower (see
+ * `shouldRetryAfterReset`): a bare `overloaded` is excluded here only because a live object
+ * shedding load is not a reset.
+ */
+export function isDoResetError(e: unknown): boolean {
+  if (typeof e !== "object" || e === null) return false;
+  const flags = e as { durableObjectReset?: unknown; retryable?: unknown };
+  return flags.durableObjectReset === true || flags.retryable === true;
```
Maximo-Guk, line 30: Where does durableObjectReset come from, how is it different from retryable?

Maybe also worth linking to from this comment
https://developers.cloudflare.com/durable-objects/best-practices/error-handling/

ndisidore: i added a comment about it but they serve different purposes
`retryable` is about _the call_ (e.g. kj::DISCONNECTED), `durableObjectReset` is about _the object_ (its incarnation died, so every stub to it is dead)

not sure why its not documented but its def in workerd
https://github.com/cloudflare/workerd/blob/6cc5b04a8bb0c4e9fa54eea2e258298d94e4d643/src/workerd/jsg/util.c%2B%2B#L224-L232

Maximo-Guk: Maybe worth bringing up with the team to see why it’s not documented, IMO it’s pretty weird to not have that covered by retryable property

Maximo-Guk (approved):

kentonv (approved):

PR #264, opened 2026-08-18
harden missing file handling during gadget export

maxwellpeterson:
- skip server-side discovery if server.js does not exist
- don't advertise default browser-mode export formats if client.js does not exist

### DIFF
--- a/packages/workshop-backend/src/overseer.ts
+++ b/packages/workshop-backend/src/overseer.ts
@@ -2554,7 +2554,7 @@ class OverseerImpl implements AgentHooks {
       : Promise<GadgetExportFormat[]> {
     this.checkChatExistsAndMaterializeDrafts(chatId);
     let resolved = await this.#resolveGadgetExportFormats(gadgetId, chatId);
-    resolved.gadget[Symbol.dispose]();
+    resolved.gadget?.[Symbol.dispose]();
     return resolved.formats;
   }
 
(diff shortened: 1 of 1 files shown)


packages/workshop-backend/src/overseer.ts
```diff
@@ -2590,8 +2591,17 @@ class OverseerImpl implements AgentHooks {
...
+    gadget: NativeRpcStub<any> | null;
   }> {
+    let {ydoc} = this.buildYDoc("current");
+    if (chatId !== undefined) {
+      this.getProposedChanges(chatId).forEach(({update}) => {
+        if (update !== undefined) Y.applyUpdateV2(ydoc, update);
+      });
+    }
+    let files = ydoc.getMap<Y.Text>(this.gadgetRootName(gadgetId));
+    if (!files.has("server.js")) return {formats: [], handler: null, gadget: null};
```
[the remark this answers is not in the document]
maxwellpeterson: This is intentional. A server stub is required to render the client in a remote browser. Client-only Gadgets shouldn't exist.

packages/workshop-backend/src/overseer.ts
```diff
@@ -2590,8 +2591,17 @@ class OverseerImpl implements AgentHooks {
   async #resolveGadgetExportFormats(gadgetId: WorkpieceId, chatId?: number): Promise<{
     formats: GadgetExportFormat[];
     handler: Fetcher<GadgetExportEntrypoint> | null;
-    gadget: NativeRpcStub<any>;
+    gadget: NativeRpcStub<any> | null;
   }> {
+    let {ydoc} = this.buildYDoc("current");
```
Maximo-Guk, line 2596: Would it be worth turning this logic in some helper like `getGadgetFiles(gadgetId, chatId)` which gets reused in getGadgetsUIBundle and here?

Maximo-Guk (approved):

packages/workshop-backend/src/overseer.ts
maxwellpeterson: I think Kenton's git-based gadget storage PR is going to clobber this in a couple days, so going to leave as-is

PR #265, opened 2026-08-19
unpack built-in blueprints

maxwellpeterson:
- store blueprint source files in source control instead of .gadget archive
- this makes it much easier to develop and review changes to built-in blueprints


packages/workshop-backend/format-blueprints/workspace-docs/files/client.js
```diff
@@ -0,0 +1,1778 @@
...
+    a.appendChild(node);
+    node = a;
+  }
+  return node;
+}
+
+const BLOCK_TAGS = ["P", "H1", "H2", "H3", "H4", "H5", "H6", "BLOCKQUOTE", "PRE", "UL", "OL", "LI", "DIV"];
+
+function sanitizePastedHtml(html) {
+  const parsed = new DOMParser().parseFromString(html, "text/html");
```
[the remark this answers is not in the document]
maxwellpeterson: GPT disagrees:
> The parsed clipboard HTML is never inserted into the live document. It is parsed into an inert document, traversed, and rebuilt using an explicit element allowlist and DOM APIs. Script/style elements are discarded, text uses text nodes, links are normalized, and images accept only validated embedded data URLs. CodeQL does not recognize this custom sanitizer.

Regardless, this issue predates this format conversion PR and I don't want to spend time on it right now.

packages/workshop-backend/scripts/format-blueprint-files.ts
```diff
@@ -0,0 +1,215 @@
+import { lstat, readdir, readFile } from "node:fs/promises";
+import type { Dirent } from "node:fs";
+import { gzipSync, gunzipSync } from "node:zlib";
+import { join } from "node:path";
+import * as Y from "yjs";
```
Maximo-Guk, line 8: It's interesting to me that we're importing YJS here. I thought gadgets are no longer backed by yjs

scripts/tsconfig.json
```diff
@@ -19,5 +19,5 @@
     // The root sets allowJs; oxlint-plugin.mjs stays untyped.
     "checkJs": false
   },
-  "include": ["**/*.ts"]
+  "include": ["**/*.ts", "../packages/workshop-backend/scripts/*.ts"]
```
Maximo-Guk, line 24: Hmm, should this be in this tsconfig?

Maximo-Guk:
> 1. **P1** `scripts/import-format-blueprint.ts:203`: Extracted files matching Git ignore rules, such as `dist/`, `.env`, or rules from an imported `.gitignore`, are bundled locally but omitted from commits. Deployments then silently receive incomplete blueprint code. `.git` path components are similarly untrackable. Reject or detect untracked extracted paths.
> 2. **P2** `scripts/format-blueprint-files.ts:169`: The default `TextDecoder` strips a leading UTF-8 BOM. Importing a file beginning with U+FEFF therefore changes its content during the immediate rebuild. Use `ignoreBOM: true`.
> 3. **P2** `scripts/format-blueprint-files.ts:206`: Path validation permits Windows-invalid names including `CON`, `AUX.js`, `a:b.js`, trailing dots/spaces, and reserved characters. Importing on Windows can fail or create an NTFS alternate stream, while Linux can commit files Windows cannot check out. Reject non-portable segments.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/33547561500)

Some GPT comments to dedupe against

1. P1 Manifest JSON can escape sourceDir  
packages/workshop-backend/scripts/import-format-blueprint.ts:76-78  
...JSON.parse(source) overrides the trusted directory-derived name. A manifest containing "name":"../victim" causes the importer to replace and delete directories outside FORMAT_BLUEPRINTS_DIR.
2. P1 Invalid imports destroy the previous source before validation  
packages/workshop-backend/scripts/import-format-blueprint.ts:182-193,208-210,253  
Malformed metadata is written, the backup deleted, and only then the builder validates it. An archive with {} metadata exits unsuccessfully but leaves the old files replaced and version: null.
3. P2 Empty blueprints pass build but fail at runtime  
packages/workshop-backend/scripts/build-format-blueprints.ts:217-228  
An empty files/ directory builds successfully, while initializeFromBlueprint() rejects it at overseer.ts:8542-8543.
4. P2 Case-insensitive directory aliases corrupt paths  
packages/workshop-backend/scripts/format-blueprint-files.ts:185-191  
Only complete paths are compared. Foo/a.js and foo/b.js pass validation, but on macOS import as Foo/a.js and Foo/b.js, silently changing the archive.
5. P2 Extracted .gitignore files can hide imported source  
packages/workshop-backend/scripts/format-blueprint-files.ts:206-211  
An archive containing files/.gitignore with * hides itself and newly extracted files from Git. The manifest bump remains visible, so CI can rebuild and ship an incomplete update.

Maximo-Guk (approved):

packages/workshop-backend/scripts/format-blueprint-files.ts
maxwellpeterson: This is true but blueprints still use Y.js for backwards compatibility. Blueprints created before the git storage changes were encoded using Y.js, and we didn't want to break those or introduce a second blueprint encoding format.

scripts/tsconfig.json
maxwellpeterson: I think we could copy this tsconfig to the backend scripts folder or keep the reference here or bump this up to a top-level tsconfig.scripts.json or similar. It's a bit odd here but I think it's fine. Logically our scripts should be covered by the same tsconfig file.

PR #268, opened 2026-08-19
Add agent evaluations and Gadget durability tests

AshishKumar4:
Adds an evaluation suite that measures what the Workshop agent delivers, and integration tests that
pin the platform behaviour the suite depends on.

`docs/testing-strategy.md` describes the whole design. This description gives the summary.

## What this adds

- `packages/workshop-evals` — 8 tasks. Each task gives the agent a prompt, then calls the RPC of the
  Gadget that the agent built. A task scores the result, and never the method.
- `packages/integration-tests` — two new capabilities on `AgentSession`. `seedGadget()` writes
  hand-authored source, so a test can use a known implementation. `restartGadgets()` restarts every
  Gadget server.
- `packages/integration-tests/__tests__/gadget-durability.test.ts` — 4 tests. They run in `pnpm test`,
  they need no model, and they finish in 5 seconds.
- `packages/workshop-agent-cli` — runs one prompt locally, without a browser.
- `.github/workflows/workshop-evals.yml` — runs the evaluations each night and on demand.

## Scoring

The score is the fraction of checks that passed. Nothing inspects the tools that the agent called.

One task asks for an appointment desk that never sells more places than it has. The agent passed it
with one synchronous SQL sequence and a database trigger. A check for a mutation queue would have
failed that correct answer.

The tool calls, the errors, the turns, the tokens, the time, and the cost are diagnostics. They explain
a result, and they never decide it.

A trial is invalid when we cannot attribute its result to the agent. The summary excludes an invalid
trial from every rate, and counts it in a separate column. A rate limit therefore reads as missing
data, and not as a regression.

## Gates

| | Gate 1 | Gate 2 |
|---|---|---|
| Trigger | every pull request | each night, on demand, or the `run-evals` label |
| Command | `pnpm build`, `pnpm test`, `pnpm lint` | `pnpm eval:required`, `pnpm eval:frontier` |
| Blocks a merge | yes | only with the label |

A **required** task fails Gate 2 when its score is below 1. A **frontier** task records its score and
never fails. We set the state from measurement. A new task starts as frontier.

## Verified

- `pnpm build`, `pnpm test`, and `pnpm lint` pass. 76 unit tests, plus the 4 durability tests.
- A naive booking implementation oversells. Therefore the concurrency check can fail.
- A restart keeps storage and drops memory. It holds across 5 restarts, and across a restart that
  interrupts a write.
- 5 tasks passed a live run against `@cf/zai-org/glm-5.2`. One task passed one trial in three.
- 33 task constants match a reference implementation. This found one wrong assertion in a task.

## Not verified

- Two tasks have no live measurement. Repeated attempts hit the Workers AI rate limit. Both tasks
  carry the frontier state, so they cannot fail the job.
- I verified `restartGadgets()` on the mainline branch only. On a chat branch, `stock-ledger` proves
  the restart before it trusts it, so a silent failure reports itself.

One run against an account with fresh quota closes both:
`WORKSHOP_EVAL_TRIALS=3 pnpm eval:frontier`.

## Draft

I opened this as a draft. It needs a baseline run with 10 trials and 2 models. That run decides the
state of each task.


AshishKumar4:
Replaced by a stack, so the toolkit can be reviewed apart from the suite:

- #270 — the agent session driver and the Gadget durability tests. Deterministic, no model, runs in `pnpm test`.
- #271 — the evaluation suite, stacked on #270.

I also dropped the agent CLI from the change. Nothing in the suite used it.

PR #270, opened 2026-08-19
Add an agent session driver and Gadget durability tests

AshishKumar4:
Adds the reusable out-of-process driver for Workshop integration tests and live agent evals.

Everything in this PR is deterministic and runs in `pnpm test`. It does not call a model.

## AgentSession

`AgentSession` drives the same public Cap'n Web API as the browser. It:

- opens a new workspace and keeps one chat across turns
- waits for the production agent lifecycle to settle
- reads complete canonical history and workpieces
- connects to generated Gadget RPCs
- accepts the chat's proposed changes through the current git-backed merge API
- reports chat token and cost metadata when available
- disposes every subscription and RPC capability

Local sessions create an isolated password account. A deployed preview can attach an Access application JWT to the WebSocket handshake and call `authenticateFromCfAccess()`.

## Shared harness

`startHarness()` can retain the Worker Loader when an agent eval needs to execute generated Gadgets. Existing deterministic integration tests continue to use the same multi-Worker workerd and gatekeeper setup.

## Verification

- `pnpm --filter @gadgets/integration-tests build`
- `pnpm --filter @gadgets/integration-tests test:run`
- 27 tests pass

The live eval package that consumes this driver is stacked in #271.


packages/integration-tests/fixtures/seeded-gadgets.ts
```diff
@@ -0,0 +1,92 @@
...
+ * agent produced. Kept as strings because that is exactly what `AgentSession.seedGadget()` writes
+ * into the workspace's Yjs document.
+ */
+
+/**
+ * Stores notes in SQL and deliberately also keeps state in memory, so a test can tell a restart
+ * apart from a no-op: `rows` must survive one and `instance`/`callsSinceStart` must not.
+ */
+export const DURABLE_NOTES_SERVER = `
+import { DurableObject } from "cloudflare:workers";
```
Maximo-Guk: Is there a way we can maybe still type this?

AshishKumar4: Agreed. The seeded programs now live as ordinary .js modules with @ts-check and JSDoc contracts, so TypeScript checks the source that workerd executes. The fixture loader reads those checked modules as strings only at the Yjs seeding boundary.

AshishKumar4:
The reduced eval implementation and its shared session driver now belong to one review unit. #271 contains this branch and is being retargeted to main.

PR #271, opened 2026-08-19
Add an evaluation suite for the Workshop agent

AshishKumar4:
Adds real end-to-end evaluations for the Workshop agent. Stacked on #270, which supplies the shared workerd and Cap'n Web driver.

## Execution

Each trial gives the production agent a natural-language task in a real Workshop. The agent builds a Gadget, and the verifier calls that Gadget's real RPC. The same scenario definitions run against:

- local workerd, for iteration and repeated experiments
- an Access-authenticated deployed preview, for functional end-to-end validation

Run locally with `pnpm evals`. Set `WORKSHOP_EVAL_TARGET` to a preview router URL and provide `CF_ACCESS_TOKEN` to target a deployment.

Live cases use `evals/**/*.eval.ts`. The package's cached `test` task includes only `src/**/*.test.ts`, so `pnpm test` never invokes a model. The root eval command invokes the live package directly after the build, preserving credentials outside Vite+ task caching.

## Framework ownership

The package uses `vitest-evals` for normalized transcripts, tool calls, errors, usage, timings, scores, the local report UI, and combined GitHub reporting. Repository code is limited to the Workshop target bridge, Gadget RPC verifier, and task content. There is no custom statistics package, report format, trace UI, or shard reducer.

## Initial scenarios

- appointment capacity under concurrent booking
- a multi-turn household ledger with exact arithmetic and preserved earlier behavior
- creation and editing of a real standard Doc

Checks score observable behavior, not a prescribed implementation technique. Tool failures, agent errors, turns, token usage, cost, and wall time remain diagnostics.

## Automation

The `Workshop evals` workflow has only `workflow_dispatch`. It never runs on pull requests, pushes, merges, or a schedule. It uses existing AI Gateway secrets, native Vitest sharding, and the pinned `getsentry/vitest-evals` action.

## Live proof on current main

- Appointment desk: GLM 5.2 passed every check; score 1.00, 68,194 tokens, 27 tools, 805 seconds.
- Expense ledger: the current git-storage path completed both turns and the turn-two check proved turn-one state survived. The agent then failed to recover from one malformed `editFile` call, so the new budget methods were absent: 7/9 checks, score 0.78, 29,354 tokens, 13 tools. This is an agent failure recorded by the suite, not hidden as a harness pass.

## Verification

- `pnpm build`
- `pnpm test`
- `pnpm lint`
- 19 eval-package unit tests pass

### DIFF
--- a/.github/workflows/workshop-evals.yml
+++ b/.github/workflows/workshop-evals.yml
@@ -1,0 +1,95 @@
+name: Workshop evals
+
+on:
+  workflow_dispatch:
+    inputs:
+      models:
+        description: Comma-separated Workshop model IDs
+        default: "@cf/zai-org/glm-5.2,@cf/moonshotai/kimi-k2.7-code"
+        required: true
+      trials:
+        description: Repetitions per task and model
+        default: "3"
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  evals:
+    name: shard ${{ matrix.shard }}/3
+    runs-on: ubuntu-latest
+    timeout-minutes: 240
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3]
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Set up Vite+, Node.js and dependencies
+        uses: voidzero-dev/setup-vp@313600b80b104eadebb9111787d37a2e83e014ca # v1.17.0
+        with:
+          node-version: "24.19.0"
+          cache: true
+          run-install: true
+
+      - name: Build
+        run: pnpm build
+
+      - name: Run eval shard
+        id: run
+        continue-on-error: true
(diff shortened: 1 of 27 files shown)


.github/workflows/workshop-evals.yml
```diff
@@ -0,0 +1,157 @@
+name: Workshop evals
+
+# Live agent evals: every run boots a workerd Workshop and drives a real model, so this is kept out
+# of CI. It runs nightly, on demand, and on a pull request only when someone adds the `run-evals`
+# label. The `main` push trigger is deliberately absent: a merge that regresses the agent is caught
+# by the following night's run, and running on every merge would serialise the queue behind them.
+on:
+  schedule:
+    - cron: "0 5 * * *"
```
Maximo-Guk: Maybe we don't need cron right away, maybe later once we have an ingestion pipeline

AshishKumar4: Agreed. I removed the cron trigger. The workflow now runs only by manual dispatch or the run-evals PR label; recurring runs wait until results have durable experiment storage instead of expiring artifacts.

AshishKumar4:
Closing the stack so the reduced toolkit and eval suite can be reviewed together in one PR against main.

PR #272, opened 2026-08-19
Report malformed percent-escapes in Google resource URLs

### LINKED ISSUE #55: Gmail URLs containing a bare % throw an unhandled URIError when binding a resource

snowyukitty:
## What does this change?

Binding a Google resource from a URL containing a bare `%` throws an unhandled `URIError: URI malformed` instead of the explanatory errors every sibling validation path in `getGatekeeperClassFor()` raises. `https://mail.google.com/mail/u/0/#search/50%off` is the reproduction reported in #55.

This adds a `decodeResourceUrlPart()` helper — `decodeURIComponent()` in a `try`, rethrowing as `Invalid <part>: it cannot be percent-decoded. Encode a literal "%" as "%25".` — and uses it at the four bare decode sites in `getGatekeeperClassFor()`: the Gmail `#search/` query, the Gmail `#label/` name, the Google Calendar path segment, and the BigQuery path segments.

#55 notes the same bare-decode pattern in other gatekeeper packages; this PR deliberately stays on the reported backend path to remain a small, single-concern change.

Fixes #55

The success path returns exactly what `decodeURIComponent()` returned, unchanged. The only behavioral difference is that a thrown `URIError` becomes an `Error` whose message follows the wording style of the sibling validation messages in the same function. There are no control-flow, dependency, or API changes.

Verified on Windows 11 (Node 22.22.3, pnpm 11.17.0) against `4288713`: `tsc -p packages/gatekeeper-google` and full `pnpm lint` pass.

AI tools authored this patch under human direction, and an independent AI review approved it before submission; the verification commands above were run locally.

## Checklist

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-os/pull/272" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/gatekeeper-google/src/google.ts
+++ b/packages/gatekeeper-google/src/google.ts
@@ -160,6 +160,18 @@ function validateGmailQueryForGrouping(query: string): void {
   if (quote || stack.length > 0) throw new Error("Gmail query has unterminated grouping or quotes.");
 }
 
+// decodeURIComponent(), but an undecodable percent-sequence names the offending part of the
+// resource URL instead of leaking a bare URIError. A literal `%` in a Gmail search or label
+// ("50% off") hits this whenever it is not escaped as `%25`.
+function decodeResourceUrlPart(encoded: string, what: string): string {
+  try {
+    return decodeURIComponent(encoded);
+  } catch {
+    throw new Error(
+        `Invalid ${what}: it cannot be percent-decoded. Encode a literal "%" as "%25".`);
+  }
+}
+
 function getBaseUrl(env: Env) {
   return stripTrailingSlashes(env.BASE_URL || "http://localhost:8787/gatekeeper/google");
 }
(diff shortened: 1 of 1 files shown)


snowyukitty:
Closing this myself: CONTRIBUTING asks contributors to keep PRs to "a dozen or so lines" and this one is +16/-4, which is over that line. #221 was the right size and I should have kept to it here.

Leaving the report behind in case it is useful: in `packages/gatekeeper-google/src/google.ts`, `decodeURIComponent` is called on parts of a Google resource URL, so an unescaped literal `%` — a Gmail search or label like "50% off" — throws a bare `URIError` that does not say which part of the URL failed to decode.

I am not reopening this as a smaller PR. Splitting a change to fit the size guideline would be evading it.

PR #274, opened 2026-08-19
store context files as byte arrays instead of JS strings

maxwellpeterson:
This fixes a bug where file length was validated based on UTF-8 byte length, but then files were serialized as UTF-16 in DO storage. With the UTF-8 file length limit set to 1.4MB and DO row size limited to 2MB, a 1.1MB UTF-8 file would pass the length check but fail when written to DO storage.

The short term fix here is to store all files as byte arrays instead of strings. For text files, we control the encoding directly and use UTF-8 instead of UTF-16. For binary files, we avoid the double overhead of base64 and UTF-16 encoding. This lets us raise the supported file size limit to 1.8MB, and closes the gap that 1.1MB files previously snuck through.

This is not a long term solution though, since we will want to support files larger than 1.8MB in the future. We'll need to switch to a DO-based filesystem implementation that supports chunking files larger than 1.8MB over multiple rows, or stop using DO storage for files.

### DIFF
--- a/packages/gatekeeper-context/__tests__/artifact-sync-fetch.test.ts
+++ b/packages/gatekeeper-context/__tests__/artifact-sync-fetch.test.ts
@@ -1,0 +1,61 @@
+import { promises as fsp } from "node:fs";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+const git = vi.hoisted(() => ({
+  fetch: vi.fn(),
+  listFiles: vi.fn(),
+  listServerRefs: vi.fn(),
+  readBlob: vi.fn(),
+}));
+
+vi.mock("isomorphic-git", () => ({
+  clone: vi.fn(),
+  fetch: git.fetch,
+  listFiles: git.listFiles,
+  listServerRefs: git.listServerRefs,
+  readBlob: git.readBlob,
+  resolveRef: vi.fn(),
+}));
+
+vi.mock("isomorphic-git/http/web", () => ({ request: vi.fn() }));
+
+import { readArtifactRepoDocuments } from "../src/artifact-sync.js";
+
+const dir = "/tmp/artifacts/fetch-test";
+
+afterEach(async () => {
+  vi.clearAllMocks();
+  await fsp.rm(dir, { recursive: true, force: true });
+});
+
+describe("artifact repository refresh", () => {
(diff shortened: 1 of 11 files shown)


Maximo-Guk:
GPT comments:

- Medium: context-collection.ts:377-385 validates only body size. Unbounded caller-controlled contentType can still push the serialized record beyond the 2 MiB storage limit. Validate the complete record or bound all metadata.
- Medium: context-storage.ts:23-26 silently accepts malformed base64. For example, "AQID!" is stored and returned as "AQID". Invalid binary bodies should be rejected rather than corrupted.
- Low: context-storage.ts:4,18-20 uses a default TextDecoder, which strips a leading UTF-8 BOM. Web-written text therefore no longer round-trips exactly. Use { ignoreBOM: true }.

Maximo-Guk (approved):

packages/gatekeeper-context/__tests__/artifact-sync-fetch.test.ts
```diff
@@ -0,0 +1,97 @@
...
+const dir = "/tmp/artifacts/fetch-test";
+
+afterEach(async () => {
+  vi.resetAllMocks();
+  await fsp.rm(dir, { recursive: true, force: true });
+});
+
+describe("artifact repository refresh", () => {
+  it("reads the fetched commit instead of stale local HEAD", async () => {
+    await fsp.mkdir(`${dir}/.git`, { recursive: true });
```
Maximo-Guk, line 35: nit: Should this setup be in a beforeEach?

packages/gatekeeper-context/vitest.config.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default defineConfig({
+  plugins: [capnwebValidate(), cloudflareTest({
+    miniflare: {
+      compatibilityDate: "2026-02-02",
+      compatibilityFlags: ["nodejs_compat", "allow_irrevocable_stub_storage"],
+    },
+  })],
+  test: {
+    fileParallelism: false,
```
Maximo-Guk, line 13: Is there a reason why fileParallelism was set to false?

Maybe each test can just use a unique temp dir (e.g. via fs.mkdtemp), which would let files run in parallel again

maxwellpeterson:
Will address comments in follow-up PR, want to get this fix out now to fix prod

PR #275, opened 2026-08-19
Convert backing storage to git, change sync to OT, editor widget to CodeMirror

kentonv:
**LARGE CHANGE ALERT**

This change does three things:
* Accepted Gadget code is now stored in git format, rather than Yjs. Each workspace has a (hidden) git repository, and each gadget's code is a commit in this repository. (Each gadget is essentially a branch, co-existing within the same repo, even if the change histories are unrelated.)
* Pending changes (changes in a chat that haven't been accepted yet) are now represented as operational transforms rather than CRDT. This allows the changes to apply directly on top of the underlying git storage, without the need to copy the contents of git into a CRDT upfront before editing can start. (I had initially tried to continue using Yjs, but the need to "seed" the Yjs doc from git whet starting changes was inefficient and convoluted.) The OT library used is `@codemirror/state`.
* Since we switched to CodeMirror's OT library, I looked into CodeMirror in general as an alternative to Monaco, and concluded that it's actually a much better fit for us. So I went ahead and ripped out Monaco in favor of CodeMirror entirely. Otherwise, it would have been necessary to develop a hand-rolled adapter between Monaco and CodeMirror's OT format -- wasted work if we later decided to switch to CodeMirror anyway, which seemed likely.

[The plan is documented in `plans/git-storage.md`.](https://github.com/cloudflare/cloudflare-os/blob/d067d55cc47128f3197cbd30c7222675da00018f/plans/git-storage.md) Note that the plan is long and composed into three parts. Each part begins with a human-readable overview and then goes into increasing detail intended for an agent to follow.

The code was almost entirely written by Fable with Sol acting as code reviewer at each step. I am pretty happy with the code I've reviewed but need to spend a bit more time looking before un-drafting this.


kentonv:
Bugs reported by testers:

* ~Diff highlighter is too aggressive with granularity, trying to diff character-by-character. Should be limited to whole words.~
* ~"Draft changes in progress" appears randomly when the user hasn't made any changes -- maybe triggered (incorrectly) by agent changes.~
* ~Streaming of code changes is broken.~
* ~Clicking "accept changes" sometimes sends me out of the chat thread.~
* ~When the connection has non-negligible latency, in the code editor, the cursor seems to disappear 1RTT after each keypress. Somehow this doesn't seem to happen when running locally?~

Maximo-Guk (approved):

ndisidore (approved):

PR #286, opened 2026-08-20
Lay the foundation for a Google Drive gatekeeper

ndisidore:
Groundwork for a Drive gatekeeper, in two commits, with no new resource type yet.

The first gives `gatekeeper-google` a vitest project — it had none — and moves the Gmail validators and the five resource declarations into `gmail-validate.ts` and `resources.ts`, so adding a resource type stops meaning five edits scattered through a 3,800-line file. `getGatekeeperClassFor` becomes a switch over `parseResourceUrl`, fixing an unrecognised resource URL that fell through to Gmail rather than throwing.

The second extracts three engines: `observers.ts`, the strategy C tracker BigQuery and Calendar each kept a near-identical copy of, now shared with persisted key encodings unchanged and two faults fixed (an unbounded `Promise.all` over observers × tracked sets, and no cap on tracked sets at all — now 1,000, failing closed without disturbing existing observers); `cursor.ts`, Gmail's thread paging, generalised so a page emptied by a scope filter is walked past rather than mistaken for the end of the results; and `drive-api.ts`, which moves the configurator's Drive search off a raw `fetch` onto `fetchWithAuthRetry` so a stale token refreshes instead of 401ing.

### DIFF
--- a/packages/gatekeeper-google/__tests__/configurator-url.test.ts
+++ b/packages/gatekeeper-google/__tests__/configurator-url.test.ts
@@ -1,0 +1,82 @@
+// A configurator's `resourceUrl` mints the string that `parseResourceUrl` then turns into a
+// capability, and its `initialValuesFromResourceUrl` reads a URL the user pasted back into form
+// values. Neither can import `resources.ts`: `scripts/build-gatekeeper-configurator.ts` transpiles
+// each configurator module on its own, stripping only `@gadgets/configurator-ui` and type-only
+// imports, so a runtime import would not resolve inside the sandboxed frame. The duplication is
+// deliberate; this test is what keeps the copies honest. Drift on either side -- a lost
+// `encodeURIComponent`, a normalization one side does and the other does not -- shows up here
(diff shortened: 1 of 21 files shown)


maxwellpeterson:
A couple findings from local review:
```
1. Medium: observers.ts:194-200 enforces the 1,000-item cap during addObserver(), not when tracking items. Once crossed, every collaborator is denied on every reopen, including existing observers, because the overseer re-verifies each open (overseer.ts:6184-6188). Reject observations that would exceed the cap before persisting them.
2. Medium: resources.ts:202-214,254,259 includes complete resource URLs in errors. Credentials, query parameters, or fragments could consequently reach frontend error reporting. Use generic or sanitized host/path messages.
3. Low: cursor.ts:89-97 only detects consecutive repeated page tokens. A cycle such as 1 → 2 → 1 → 2 returns duplicate pages indefinitely; cursor.test.ts:137-147 explicitly permits this. Track all encountered tokens.
```

maxwellpeterson (approved):

PR #292, opened 2026-08-21
Add Google Drive metadata search

ndisidore:
This adds read-only Google Drive metadata search for a connected account, one organization-owned shared drive, or one file. Agents can find files and folders, inspect their metadata, and page through results without receiving file contents or changing anything in Drive. Each binding stays within its selected scope, and shared workspaces recheck access to every file they have seen. The change also adds resource pickers, OAuth scope mapping, the public TypeScript API, and coverage for queries, pagination, scopes, and sharing.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/292" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


Maximo-Guk:
Few GPT findings:

```
Findings
1. High: OAuth flows can record resources from a different consent attempt. packages/gatekeeper-google/src/google.ts:464-510  
While acceptAuthCode() awaits token exchange, prepareReconnect() can overwrite requestedResources. The earlier callback then records and deletes the newer flow’s resources, potentially granting Drive access never presented in that completed flow. Bind scopes/resources to the OAuth nonce.
2. High: Bulk observer admission has a disclosure race. packages/gatekeeper-google/src/observers.ts:261-276  
A concurrent observation can add an inaccessible file after listTracked() but before admission completes. Although the staged observer is excluded, the overseer ignores that ID until it is persisted at overseer.ts:7945-7947; admission then succeeds without checking the new file. Re-list newly tracked sets before finalizing admission.
3. High: Failed re-verification removes an existing observer. packages/gatekeeper-google/src/observers.ts:261-280  
Reopening overwrites the existing verifier before validation, then deletes it on failure. Previously opened collaborators remain authorized in the sharing graph but disappear from future excludeObservers checks. Preserve and restore the previous observer record on failure.
4. Medium: Shared-drive bindings expose a cross-scope access oracle. packages/gatekeeper-google/src/drive-session.ts:241-244, packages/gatekeeper-google/src/drive-session.ts:300-303  
A readable foreign file produces “outside this Drive binding,” while an inaccessible ID returns Drive’s 403/404 error. This lets a scoped gadget probe the owner’s access to known files outside its capability. Normalize both outcomes.
5. Medium: Empty Drive searches bypass observation authorization. packages/gatekeeper-google/src/drive-session.ts:251-266, packages/gatekeeper-google/src/cursor.ts:109-115  
Terminal empty pages return null without authorizeObservation(). Negative filename/full-text probes therefore bypass auditing and observer enforcement.
6. Medium: User-facing copy understates the account binding’s authority. packages/gatekeeper-google/src/configurator/drive-account-configurator-ui.tsx:13-19, packages/gatekeeper-google/src/google.ts:3003-3008  
It promises “My Drive and Shared with me” metadata only, but getEntry() reaches any accessible shared-drive ID and fullTextContains probes body/OCR text. The consent surface should describe both.
7. Low: File-scoped list() can return trashed files contrary to its contract. packages/gatekeeper-google/src/drive-session.ts:217-220, packages/gatekeeper-google/src/drive-session.ts:269-276  
The exact-file path uses files.get without requesting/checking trashed, while DriveEntry documents that list() never returns trashed entries.
```

maxwellpeterson (approved):

PR #293, opened 2026-08-21
Stop verifying observers for the Context Library gatekeeper

Maximo-Guk:
Currently observer verifications are required for the context library gatekeeper, however, currently there's no way to share private context collections with others. 

In the near future we're considering making it possible to share your context collections with other users/making sharing configurable on individual skill files. However for the time being we're planning on simply disabling the observer check as a stop gap measure, you still need to explicitly share your workspace with others for them to see your context collection.


ndisidore (approved):

Maximo-Guk:
Closing this for now as we decided to go with a different approach

PR #296, opened 2026-08-21
Add Google Drive native document sessions

ndisidore:
This extends in-scope Google Drive bindings with read-only nested sessions for native Google Docs and Google Sheets. Agents can open a native file only after a fresh metadata lookup, immutable account/shared-drive/file scope enforcement, and observation authorization; Docs return Markdown, Sheets retain bounded A1 range reads, and each child capability is independently disposable. Account and exact-file grants gain the content-specific read scopes while shared-drive grants keep `drive.readonly`, and no Drive path gains writes, blob extraction, or shortcut traversal. The change also gives Docs requests a 30-second timeout and 10 MiB successful-response ceiling, preserves Sheets' existing bounds, redacts provider prose while logging bounded reason codes, and adds coverage for scoping, observer ordering, OAuth upgrades, RPC ownership, streamed limits, and agent-facing declarations.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/296" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


Maximo-Guk:
Some GPT comments:

```
1. High: Docs edits may apply twice after an ambiguous timeout. docs-api.ts:103-105,146-164 applies the new timeout to batchUpdate; if Google commits before aborting, google.ts:2083-2086 leaves the action pending. Retrying an append duplicates it. Add write reconciliation/idempotency.
2. Medium: Multi-tab Docs silently return only the first tab. docs-api.ts:112-117 omits includeTabsContent=true, while google.ts:3139-3145 presents the result as the complete body. Traverse all tabs or document/reject multi-tab Docs.
```

packages/gatekeeper-google/src/drive-types.d.ts
```diff
@@ -155,4 +157,20 @@ export interface GoogleDriveSession {
    * trash, while a direct get does not, and {@link DriveEntry} has no `trashed` field.
    */
   getEntry(fileId: string): Promise<DriveEntry>;
+
+  /**
+   * Open an in-scope native Google Doc with MIME type
+   * `application/vnd.google-apps.document`. Other MIME types, including folders and shortcuts, are
+   * rejected. The returned RPC capability supports promise pipelining and must be disposed when
+   * finished.
```
kentonv: Doc comments shouldn't describe RPC basics -- that should be documented somewhere the agent reads once.

```suggestion
   * Open a Google Doc present in the Drive. The MIME type must be
   * `application/vnd.google-apps.document`. Other MIME types, including folders and shortcuts, are
   * rejected.
```

packages/gatekeeper-google/src/drive-types.d.ts
```diff
@@ -155,4 +157,20 @@ export interface GoogleDriveSession {
...
+   */
+  openGoogleDoc(fileId: string): Promise<GoogleDocReadSession>;
+
+  /**
+   * Open an in-scope native Google Sheet with MIME type
+   * `application/vnd.google-apps.spreadsheet`. Other MIME types, including folders and shortcuts,
+   * are rejected. The returned RPC capability supports promise pipelining and must be disposed when
+   * finished.
```
kentonv:
```suggestion
   * Open a Google Sheet present in the Drive. The MIME type must be
   * `application/vnd.google-apps.spreadsheet`. Other MIME types, including folders and shortcuts,
   * are rejected.
```

packages/gatekeeper-google/src/docs-read-types.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  /** When the document was last modified. */
+  lastModified: Date;
+}
+
+/** Read-only access to one native Google Doc. */
+export interface GoogleDocReadSession {
+  /** Return current document metadata. Works with any number of tabs. */
+  getMetadata(): Promise<DocMetadata>;
+
+  /** Return the document body as Markdown. Requires exactly one tab. */
```
maxwellpeterson, line 15: What's the plan for multi-tab documents? Extending this method with optional args?

It would also be nice to support docx/PDF export that the agent could pipe to gmail as an attachment.

maxwellpeterson (approved):

packages/gatekeeper-google/src/docs-read-types.d.ts
ndisidore: Good q: its not clear here but `getContent` is actually a pre-exsting method now exposed a new interface where this limitation exists.
And yep you nailed it. Probably going to have a `listTabs()` method and then you can optionally pass tabs: []index to this
But punting on that for now since this is pre-existing

PR #298, opened 2026-08-21
Bound every action-log read path

ndisidore:
Reading the action log today means reading all of it: opening a workspace replays every record ever written to each client, the Activity pane holds the whole log in React state, and the auto-approval drain materializes the full table per run. Long-lived workspaces pay for this on every open, and the cost only grows.

This branch bounds every read path and splits the protocol into query-for-state / subscribe-for-deltas:

- `subscribeToActions` delivers live updates only. Clients fetch the current pending set through `listActions({filter: "pending"})` after initiating the subscribe; capnweb e-order makes the pair gapless. The unpaced replay push is gone; initial state flows over pull-paged, client-clocked reads.
- The new `listActions` RPC pages history newest-first: resolved records under a raw-scan cap, so a log buried in resolved records returns short pages with a cursor instead of stalling the DO, and pending records off a new sparse pending-by-gatekeeper index
- The Activity pane demand-loads history one page at a time (type filters, "Load older", failure/retry states). Pending state runs off one ref-counted store shared per overseer stub: it subscribes, then pages the pending filter
- Chat action cards reconcile on reconnect: since the live-only stream never redelivers a resolution missed while disconnected, cached cards still shown pending are re-fetched, a few at a time.
- `AutoApprovalDrainer` reads its gatekeeper's pendings off the index instead of scanning the log.
- Pre-deploy clients that still pass `startAfter` get a paced full replay

Tested with unit suites for the subscription, pagination, migration-backfill, and drain paths, an integration smoke over the paged RPCs, and hook tests for the new frontend state.

### DIFF
--- a/packages/typed-storage/src/index.ts
+++ b/packages/typed-storage/src/index.ts
@@ -118,9 +118,7 @@ export interface Subscriber<T> {
 }
 
 /**
- * A collection of records addressed by primary key. Note: NOT `extends UniqueIndex` -- a
- * collection has the same get/list/delete, but rebuild() (re-deriving an index from the records)
- * is meaningless for the records themselves.
+ * A collection of records addressed by primary key.
  */
 export interface Collection<T extends object, PrimaryKey = string> {
   get(key: PrimaryKey): T | undefined;
--- a/packages/typed-storage/src/index.ts
+++ b/packages/typed-storage/src/index.ts
@@ -61,13 +61,34 @@ export interface UniqueIndex<T, Key> {
   get(key: Key): T | undefined;
   list(options?: ListOptions<Key>): Iterable<T>;
   delete(key: Key): boolean;
+
+  /**
+   * Discard the index's contents and re-derive them from the collection's records. Indexes are
+   * only maintained at write time, so an index declared after records already exist starts empty
+   * (and updates to those records would corrupt it, or throw); a migration must rebuild() such an
+   * index before the records are touched. Throws if two records derive the same key.
+   */
+  rebuild(): void;
 }
 
 /** An index where each key may match multiple records. */
-export interface NonUniqueIndex<T, Key> {
-  get(key: Key): Iterable<T>;
+export interface NonUniqueIndex<T, Key, PK extends string | number = string | number> {
+  /**
+   * List the records matching `key`, ordered by primary key. `options` ranges and pages over the
+   * matching records' primary keys. Note that `limit` here counts records, unlike in a top-level
+   * list(), where it counts index keys.
+   */
+  get(key: Key, options?: ListOptions<PK>): Iterable<T>;
   list(options?: ListOptions<Key>): Iterable<T>;
   delete(key: Key): number;
+
+  /**
+   * Discard the index's contents and re-derive them from the collection's records. Indexes are
+   * only maintained at write time, so an index declared after records already exist starts empty
+   * (and updates to those records would corrupt it, or throw); a migration must rebuild() such an
+   * index before the records are touched.
+   */
+  rebuild(): void;
 }
 
 type Key = string | number;
@@ -86,8 +107,8 @@ type UniqueIndexed<T, Indexes> = {
   [K in keyof Indexes]: UniqueIndex<T, RemoveArray<ReturnType<Indexes[K]>>>
 }
 
-type NonUniqueIndexed<T, Indexes> = {
-  [K in keyof Indexes]: NonUniqueIndex<T, RemoveArray<ReturnType<Indexes[K]>>>
+type NonUniqueIndexed<T, Indexes, PK extends Key = Key> = {
+  [K in keyof Indexes]: NonUniqueIndex<T, RemoveArray<ReturnType<Indexes[K]>>, PK>
 }
 
 export interface Subscriber<T> {
@@ -96,7 +117,16 @@ export interface Subscriber<T> {
   remove(record: T): void;
 }
 
-export interface Collection<T extends object, PrimaryKey = string> extends UniqueIndex<T, PrimaryKey> {
+/**
+ * A collection of records addressed by primary key. Note: NOT `extends UniqueIndex` -- a
+ * collection has the same get/list/delete, but rebuild() (re-deriving an index from the records)
+ * is meaningless for the records themselves.
+ */
+export interface Collection<T extends object, PrimaryKey = string> {
+  get(key: PrimaryKey): T | undefined;
+  list(options?: ListOptions<PrimaryKey>): Iterable<T>;
+  delete(key: PrimaryKey): boolean;
+
   put(value: T): void;
 
   subscribe(subscriber: Subscriber<T>): void;
@@ -168,7 +198,7 @@ type CollectionImpl<T extends object,
                     NonUniqueIndexes> =
     & Collection<T, PrimaryKeyType<T, PrimaryKey>>
     & UniqueIndexed<T, UniqueIndexes>
-    & NonUniqueIndexed<T, NonUniqueIndexes>;
+    & NonUniqueIndexed<T, NonUniqueIndexes, PrimaryKeyType<T, PrimaryKey> & Key>;
 
 type TypedStorageImpl<Collections, Singletons> = TypedStorage
   & {
@@ -281,6 +311,21 @@ class KvPrefixedView<T extends StorageValue> {
     return new KvPrefixedView(this.#kv, `${this.#name}.${name}`);
   }
 
+  /**
+   * Delete every child row (`name.` prefix) and record (`name:` prefix) under this view,
+   * unbuffered -- point deletes are permitted under an open list() cursor. Sweeping the raw key
+   * ranges also reclaims child rows orphaned by earlier inconsistencies, which a walk of the
+   * parent keys would never reach. The `name#` unique-id counter is intentionally kept: ids must
+   * never be reused.
+   */
+  deleteAll(): void {
+    for (let prefix of [`${this.#name}.`, `${this.#name}:`]) {
+      for (let [key, _] of this.#kv.list({prefix})) {
+        this.#kv.delete(key);
+      }
+    }
+  }
+
   getUnidqueId(): number {
     let key = `${this.#name}#`;
     let id = this.#kv.get<number>(key) || 0;
@@ -375,14 +420,15 @@ function createCollection<
 
   // Add a subscriber subscribing on behalf of an index based on the given IndexFunction. This
   // code is shared for unique and non-unique indexes. This code in particular takes care of the
-  // case where the index function returns an array.
+  // case where the index function returns an array. Returns the subscriber's add(), so callers
+  // can also feed pre-existing records into the index (see rebuild()).
   function addIndexSubscriber(
       idx: IndexFunction<T>,
       ops: {
         add(idxKey: Key, pk: Key, type: "Insertion" | "Update"): void;
         remove(idxKey: Key, pk: Key): void;
-      }) {
-    subscribers.add({
+      }): (record: T) => void {
+    let subscriber: Subscriber<T> = {
       add(record: T) {
         let pk = pkForT(record);
         let idxKeys = idx(record);
@@ -451,7 +497,9 @@ function createCollection<
           ops.remove(idxKeys, pk);
         }
       }
-    });
+    };
+    subscribers.add(subscriber);
+    return subscriber.add;
   }
 
   // ---------------------------------------------------------------------------
@@ -459,6 +507,22 @@ function createCollection<
 
   for (let [idxName, idx] of Object.entries(schema.uniqueIndexes || {})) {
     let idxKv = new KvPrefixedView<Key>(storage.kv, `${name}.${idxName}`);
+
+    let addToIndex = addIndexSubscriber(idx as IndexFunction<T>, {
+      add(idxKey: Key, pk: Key, type: "Insertion" | "Update") {
+        let oldValue = idxKv.get(idxKey);
+        if (oldValue !== undefined) {
+          throw new Error(`${type} conflicts with record '${oldValue}' in '${name}.${idxName}'.`);
+        }
+        idxKv.put(idxKey, pk);
+      },
+      remove(idxKey: Key, pk: Key) {
+        if (!idxKv.delete(idxKey)) {
+          throw new Error(
+              `Index '${name}.${idxName}' is inconsistent: removed record is not present.`);
+        }
+      }
+    });
 
     let index: UniqueIndex<T, Key> = {
       get(key: Key): T | undefined {
@@ -484,24 +548,19 @@ function createCollection<
         let pk = idxKv.get(key);
         return pk === undefined ? false : collection.delete(pk);
       },
+      rebuild(): void {
+        // One transaction, so a mid-scan throw (e.g. a key conflict) can't leave the index
+        // partially built after the wipe. The adds are point reads/writes, permitted under the
+        // record scan's open cursor.
+        storage.transactionSync(() => {
+          idxKv.deleteAll();
+          for (let record of collection.list()) {
+            addToIndex(record);
+          }
+        });
+      },
     };
     result[idxName] = index;
-
-    addIndexSubscriber(idx as IndexFunction<T>, {
-      add(idxKey: Key, pk: Key, type: "Insertion" | "Update") {
-        let oldValue = idxKv.get(idxKey);
-        if (oldValue !== undefined) {
-          throw new Error(`${type} conflicts with record '${oldValue}' in '${name}.${idxName}'.`);
-        }
-        idxKv.put(idxKey, pk);
-      },
-      remove(idxKey: Key, pk: Key) {
-        if (!idxKv.delete(idxKey)) {
-          throw new Error(
-              `Index '${name}.${idxName}' is inconsistent: removed record is not present.`);
-        }
-      }
-    });
   }
 
   // ---------------------------------------------------------------------------
@@ -510,12 +569,38 @@ function createCollection<
   for (let [idxName, idx] of Object.entries(schema.nonUniqueIndexes || {})) {
     let idxKv = new KvPrefixedView<number>(storage.kv, `${name}.${idxName}`);
 
+    let addToIndex = addIndexSubscriber(idx as IndexFunction<T>, {
+      add(idxKey: Key, pk: Key, type: "Insertion" | "Update") {
+        let id = idxKv.get(idxKey);
+        if (id === undefined) {
+          id = idxKv.getUnidqueId();
+          idxKv.put(idxKey, id);
+        }
+
+        let child = idxKv.getChild(id.toString());
+        child.put(pk, {});
+      },
+      remove(idxKey: Key, pk: Key) {
+        let id = idxKv.get(idxKey);
+        if (id === undefined) {
+          throw new Error(
+              `Index '${name}.${idxName}' is inconsistent: removed record is not present.`);
+        }
+
+        let child = idxKv.getChild(id.toString());
+        child.delete(pk);
+        if (Array.from(child.list({limit: 1})).length == 0) {
+          idxKv.delete(idxKey);
+        }
+      }
+    });
+
     let index: NonUniqueIndex<T, Key> = {
-      *get(key: Key): Generator<T, void> {
+      *get(key: Key, options?: ListOptions<Key>): Generator<T, void> {
         let id = idxKv.get(key)
         if (id === undefined) return;
         let child = idxKv.getChild(id.toString());
-        for (let pk of child.listKeys()) {
+        for (let pk of child.listKeys(options)) {
           yield collection.get(pk)!;
         }
       },
@@ -562,34 +647,19 @@ function createCollection<
           return count;
         }
       },
+      rebuild(): void {
+        // One transaction, so a mid-scan throw (e.g. a key conflict) can't leave the index
+        // partially built after the wipe. The adds are point reads/writes, permitted under the
+        // record scan's open cursor.
+        storage.transactionSync(() => {
+          idxKv.deleteAll();
+          for (let record of collection.list()) {
+            addToIndex(record);
+          }
+        });
+      },
     };
     result[idxName] = index;
-
-    addIndexSubscriber(idx as IndexFunction<T>, {
-      add(idxKey: Key, pk: Key, type: "Insertion" | "Update") {
-        let id = idxKv.get(idxKey);
-        if (id === undefined) {
-          id = idxKv.getUnidqueId();
-          idxKv.put(idxKey, id);
-        }
-
-        let child = idxKv.getChild(id.toString());
-        child.put(pk, {});
-      },
-      remove(idxKey: Key, pk: Key) {
-        let id = idxKv.get(idxKey);
-        if (id === undefined) {
-          throw new Error(
-              `Index '${name}.${idxName}' is inconsistent: removed record is not present.`);
-        }
-
-        let child = idxKv.getChild(id.toString());
-        child.delete(pk);
-        if (Array.from(child.list({limit: 1})).length == 0) {
-          idxKv.delete(idxKey);
-        }
-      }
-    });
   }
 
   // ---------------------------------------------------------------------------
--- a/packages/workshop-backend/src/overseer.ts
+++ b/packages/workshop-backend/src/overseer.ts
@@ -10,6 +10,7 @@ import {
   RpcTarget as NativeRpcTarget, restore,
 } from "cloudflare:workers";
 import { createTypedStorage, collection, keyString } from "@gadgets/typed-storage";
+import type { ListOptions } from "@gadgets/typed-storage";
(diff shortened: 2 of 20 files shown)


packages/workshop-shared/src/api.ts
```diff
@@ -1841,7 +1844,15 @@ export interface Overseer extends RpcTarget {
 
   /**
    * Subscribe to action adds/updates. Dispose the returned stub to unsubscribe.
-   * If `startAfter` is set, replay actions changed after that timestamp.
+   *
+   * On subscribe, currently-pending records are replayed through the subscriber (the server scans
+   * the log in bounded internal pages; ordering relative to live updates is simply the stream
+   * order), then ready() fires. Resolved history is fetched separately via listActions().
+   *
+   * `startAfter` is deprecated: its presence switches the replay from pending-only to every
```
kentonv: Why this design change to subscriptions?

`subscribeToActions()` was not the problem before, since it is only used to subscribe to _new_ actions as they happen. (The `startAfter` parameter is meant for re-establishing a subscription after disconnecting, without missing anything. It would normally be set to the date of the last action seen.)

With the change, it seems it is no longer possible to subscribe to observations, since they are never "pending".

ndisidore: Hmmm that may have been how it was designed, but that does not match how it was implemented. On current main, [useActions _always_ passes `new Date(0)` on every store open](https://github.com/cloudflare/cloudflare-os/blob/23ff840bc8af035da4079da6b662a48f160e3778/packages/workshop-frontend/src/useActions.ts#L83). There's even a comment that leads me to think this was intentional
```
"startAfter: epoch 0 asks the backend to replay full history through the subscriber, avoiding a stale-state race against a separate listActions()."
```
and even with the intended approach it is still a `O(entire log)` since records aren't indexed by time and overseer does `for (let record of actions.list())`

> "It is no longer possible to subscribe to observations, since they are never 'pending'"

It is :)
Give it a whirl on the preview branch and you'll see this does work
Only the replay becomes pending only - no changes to live subscription


Fair point buried in there though: the subscription by itself no longer catches you up on stuff that resolved while you were disconnected. I moved that job moved to `listActions` paging (plus the reconnect re-fetch for action cards). So the real question may be whether "subscription = live + pending, history = paged reads" is the split we want. But I think that's pretty defensible.
`listActions` was essentially dead code on main - it had zero callers.

kentonv: Yikes, OK, so the UI code was using a subscription-since-time-zero to backfill instead of using listActions()? That definitely wasn't the intended design. The idea was supposed to be that you use `listActions()` to backfill and then `subscribeToActions()` to receive new actions real-time.

> Only the replay becomes pending only - no changes to live subscription

So subscribe() subscribes to all future actions (of any type) _and also_ surfaces historical pending actions?

This seems like a confusing interface. Why combine these? Why not call `listActions()` with a filter requesting pending actions specifically?

Why are pending actions treated specially, anyway? What if the user has a million old pending actions that they intentionally never accepted or rejected? Maybe they are intentionally living in a simulation (there are valid use cases).

I think the right design here is:
- `subscribeToActions()` stays as it is. Perhaps, we modify it to limit how far in the past `startAfter` is allowed to be -- since this was only intended for resuming subscriptions. But I think for short disconnects it's really better to be able to resume the subscription rather than require a separate RPC to check for missed events.
- Use `listActions()` to query history, including of pending actions.

ndisidore: Fair points on all counts. We're basically there now (after another round of changes): 
(1) `subscribe` = deltas only, (2) `listActions()` handles all history queries including pending

- `subscribeToActions()` now delivers live deltas only; nothing is replayed. 
- Pending state comes from `listActions({filter: "pending"})`, same paging/cursor as the rest of history. Clients initiate the subscribe first, then pages
- Backing it is a sparse pending-by-gatekeeper index, which the auto-approval drain now reads too.

On the million-pending workspace: better, but I won't claim it's solved. The wire is now bounded (pull-paged, client-clocked), but each pending page still materializes the full pending set server-side, since the index can't range-read within a group yet (no time index)

The one spot I went a different way is `startAfter` resume. There's a gotcha: hook enable/disable/delete mutate the record with no timestamp field at all (appliedAt only exists on type: "action"), so a resumed subscription would miss hook toggles. Resolutions are fine, they stamp appliedAt. So rather than bound it, I kept startAfter purely as a compat shim for already-deployed clients (they still get the old full replay), with a TODO to delete once they cycle out. 

If we want cheap short-disconnect resume later, it may be worth considering a monotonic revision counter rather than wall-clock because of workerd's frozen clock or otherwise port the `getChatTimestamp` hack into something more general purpose

packages/workshop-shared/src/api.ts
```diff
@@ -1763,10 +1763,21 @@ export interface Overseer extends RpcTarget {
...
+   * currently-pending records instead — the query half of the query-for-state/subscribe-for-deltas
+   * contract (see subscribeToActions()).
+   *
+   * Page size is a server constant, and the server also caps the raw records examined per call,
+   * so a filtered page may be short or even empty while older history remains: absence of
+   * `nextBeforeId` is the only terminator.
+   *
+   * A page reflects call-time state: a record resolved between pages stops appearing in later
+   * "pending" pages (and starts appearing in resolved ones); an action subscription carries its
+   * resolution as a live update.
```
kentonv: This paragraph seems to be stating the obvious, albeit in Claudish so it's hard to tell. I'd delete it.

```suggestion
```

packages/workshop-shared/src/api.ts
```diff
@@ -1841,7 +1852,18 @@ export interface Overseer extends RpcTarget {
...
+   * state, subscribe for deltas: fetch the current pending set via
+   * listActions({filter: "pending"}) and resolved history via the other filters. As with
+   * subscribeToChat(), initiate the subscribe call before those reads — there is no need to
+   * await its return, only to start it first — so nothing can slip between the snapshot the
+   * pages reflect and the stream.
+   *
+   * `startAfter` is deprecated: its presence makes the server replay every record through the
+   * subscriber before ready(), since pre-deploy clients derive their entire history view from
+   * replay. The value itself is ignored. New clients must omit it.
+   * TODO: Delete it once pre-deploy clients have cycled out.
```
kentonv: Please don't deprecate `startAfter`.

```suggestion
   * The `startAfter` parameter is intended to be used when resubscribing after a disconnect:
   * specify the time of the last action seen, in order to ensure no actions were missed during
   * the disconnect. If not specified, the subscription starts from the current time.
   *
   * Do NOT use `startAfter` as a way to enumerate historical data. Use `listActions()` instead.
   * To ensure no holes between a subscription and historical data, call `subscribeToActions()`
   * immediately before `listActions()`, similar to `subscribeToChat()`.
```

packages/workshop-shared/src/api.ts
```diff
@@ -2460,6 +2482,34 @@ export type AiChatHistoryPage = {
   };
 };
 
+/**
+ * Filter for listActions(): "pending" for currently-pending records (of any type), or a resolved
+ * view — one specific record type, or "all" for every resolved record.
```
kentonv: Wait, so there's no way to query both pending and non-pending at the same time? Why?

packages/workshop-shared/src/api.ts
```diff
@@ -2460,6 +2482,34 @@ export type AiChatHistoryPage = {
...
+    filter: ActionHistoryFilter): boolean {
+  return filter === "pending"
+      ? record.state === "pending"
+      : record.state !== "pending" && (filter === "all" || record.type === filter);
+}
+
+/** One page of action history from listActions(). */
+export type ActionHistoryPage = {
+  /** Matching records, descending id (creation order, newest first). May be short or empty while
+   * older history remains — absence of `nextBeforeId` is the only terminator. */
```
kentonv: Claudish.

```suggestion
  /**
   * Matching records, descending id (creation order, newest first).
   *
   * Note that an empty list does NOT mean that there are no more pages. It may be that the server
   * reached the scan limit for one page without finding any matches. Only the absence of
   * `nextBeforeId` indicates that there are no more pages.
   */
```

packages/workshop-shared/src/api.ts
```diff
@@ -3301,6 +3351,12 @@ export type AiChatStreamEvent = {
 /** Interface implemented by the client to receive action-log upserts. */
 export interface ActionsSubscriber {
   entry(record: ActionLogEntry): void;
+
+  /**
+   * @deprecated Fires immediately after a modern (live-only) subscribe registers — it only ever
+   * marked the end of the replay, which now exists solely for the deprecated startAfter path.
+   * Dies with that path; clients should treat it as a no-op.
```
kentonv:
```suggestion
   * @deprecated Fires after the subscription has caught up to the current time. However, this is
   * only a useful signal when a subscription is being used to enumerate past actions using a
   * distant-past `startAfter`. This is not the correct way to use `subscribeToActions()`; use
   * `listActions()` instead.
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -1298,6 +1308,21 @@ const LISTING_REFRESH_BATCH = 16;
 // Longest noun accepted on a format reference. Denormalized display data.
 const MAX_FORMAT_REF_NOUN = 128;
 
+/**
+ * Raw records examined per page of subscribeToActions()'s deprecated startAfter full replay
+ * (legacy-only; dies with that block). Exported for tests.
```
kentonv: There seems to be a lot of effort being made here to handle old clients talking to a new server. This isn't necessary, as clients update the first time they refresh after the server updates.

packages/workshop-backend/src/overseer.ts
```diff
@@ -1298,6 +1308,21 @@ const LISTING_REFRESH_BATCH = 16;
...
+ */
+export const PENDING_SCAN_PAGE_SIZE = 256;
+
+/** listActions() entries returned per page. Exported for tests. */
+export const HISTORY_PAGE_DEFAULT_LIMIT = 50;
```
kentonv: "history" is ambiguous.

```suggestion
export const ACTION_HISTORY_PAGE_DEFAULT_LIMIT = 50;
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -1298,6 +1308,21 @@ const LISTING_REFRESH_BATCH = 16;
...
+
+/**
+ * Cap on raw records a single listActions() call examines, so a filtered scan over a huge log
+ * stays bounded. Exported for tests.
+ */
+export const HISTORY_PAGE_SCAN_CAP = 500;
```
kentonv:
```suggestion
export const ACTION_HISTORY_PAGE_SCAN_CAP = 500;
```

packages/typed-storage/src/index.ts
```diff
@@ -562,34 +602,31 @@ function createCollection<
...
-              `Index '${name}.${idxName}' is inconsistent: removed record is not present.`);
-        }
-
-        let child = idxKv.getChild(id.toString());
-        child.delete(pk);
-        if (Array.from(child.list({limit: 1})).length == 0) {
+      rebuild(): void {
+        // Each list is buffered upfront: the deletes and puts below would otherwise run under an
+        // open list() cursor.
+        for (let idxKey of Array.from(idxKv.listKeys())) {
```
kentonv: This loop to delete the index is overcomplicated and inefficient. Instead of iterating every child explicitly, it could just delete all keys with two prefixes: `<view-name>.` and `<view-name>:`. I would add a method `KvPrefixView.deleteAll()` which does this.

Note that it's perfectly fine to delete rows while list()ing them -- no need to read the entire array into memory upfront. (sqlite explicitly allows deleting a row that was just returned from a SELECT, and then continuing to advance the SELECT.)

packages/typed-storage/src/index.ts
```diff
@@ -562,34 +602,31 @@ function createCollection<
...
-        let id = idxKv.get(idxKey);
-        if (id === undefined) {
-          throw new Error(
-              `Index '${name}.${idxName}' is inconsistent: removed record is not present.`);
-        }
-
-        let child = idxKv.getChild(id.toString());
-        child.delete(pk);
-        if (Array.from(child.list({limit: 1})).length == 0) {
+      rebuild(): void {
```
kentonv, line 648: If we're adding this to `NonUniqueIndex`, we should add it to `UniqueIndex` as well.

packages/typed-storage/src/index.ts
```diff
@@ -562,34 +602,31 @@ function createCollection<
...
+        // nature, but must not materialize an arbitrarily large collection in memory.
+        let cursor: Key | undefined;
+        for (;;) {
+          let page = Array.from(collection.list({startAfter: cursor, limit: REBUILD_PAGE_SIZE}));
+          for (let record of page) {
+            addToIndex(record);
+          }
+          if (page.length < REBUILD_PAGE_SIZE) break;
+          cursor = pkForT(page.at(-1)!);
+        }
```
kentonv: I'm pretty sure this buffering is unnecessary. See if it passes tests like this:

```suggestion

        // Repopulate the index.
        for (let record of collection.list()) {
          addToIndex(record);
        }
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -9748,17 +9828,29 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
...
+    // listActions({filter: "pending"}) after initiating the subscribe (see api.ts).
     if (startAfter !== undefined) {
-      let startAfterTimestamp = startAfter.valueOf();
-      for (let record of actions.list()) {
-        if (disposed) break;
-        let appliedAt = record.type === "action" ? record.appliedAt : undefined;
-        let recordTimestamp = (appliedAt ?? record.createdAt).valueOf();
-        if (recordTimestamp > startAfterTimestamp) {
-          subscriber.entry(actionRecordToLog(record)).catch(unsubscribe);
+      // DEPRECATED: pre-deploy clients pass startAfter and build their entire history view from
```
kentonv: Revert these changes, keep startAfter the way it is.

packages/workshop-backend/src/overseer.ts
```diff
@@ -9364,13 +9411,46 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
     return result;
   }
 
-  async listActions(): Promise<ActionLogEntry[]> {
-    let result: ActionLogEntry[] = [];
-    for (let record of this.impl.storage.actions.list()) {
-      result.push(actionRecordToLog(record));
+  async listActions(options?: {beforeId?: number, filter?: ActionHistoryFilter})
```
kentonv, line 9445: This implementation seems pretty suboptimal, especially for "pending".

There are some really pathological cases here: Most workspaces don't add new hooks often, so if you go to the "activity" view and click "hooks", you almost always end up doing a full table scan. Sure, there is a limit on scan, but then this returns no results, which just causes the client to request the next page -- so you end up doing a full scan anyway, just split across several round trips, making it even slower.

We could fix all of this and simplify the code by adding another non-unique index on `ActionHistoryFilter`. Note that an index function is allowed to return an array of keys, so for a pending action, you could return `["pending", "action"]` to appear in both indexes.

packages/workshop-backend/src/overseer.ts
```diff
@@ -9900,18 +9994,11 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
 
     let self = this;
     function deliverMessage(record: AiChatMessage) {
-      subscriber.message(self.impl.hydrateChatMessageForClient(record)).catch(unsubscribe);
+      subscriber.message(self.#getChatMessageForClient(record)).catch(unsubscribe);
```
kentonv, line 10013: I guess this is a bugfix, made necessary since actions can't be cross-referenced from the action subscription anymore?

packages/workshop-backend/src/overseer.ts
```diff
@@ -9861,7 +9953,9 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
     return msg && this.#getChatMessageForClient(msg);
   }
 
-  async #getChatMessageForClient(msg: AiChatMessage): Promise<AiChatMessage> {
+  // Synchronous so subscribeToChat's deliverMessage can issue message() in the same turn as the
+  // metadata()/deleted() calls around it, preserving cross-callback delivery order.
```
kentonv:
```suggestion
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -10576,7 +10664,12 @@ class UseOverseerInterface extends RpcTarget implements Overseer {
   async newAgentSpawnerGatekeeper(_config: AgentSpawnerConfig): Promise<GatekeeperClient<any>> {
     this.#deny();
   }
-  async listActions(): Promise<ActionLogEntry[]> { this.#deny(); }
+  // Inert (see class comment): the editor pages the action log speculatively before it knows its
+  // role, so answer with an empty terminal page instead of denying.
```
kentonv: Why is the editor loading the log speculatively so early? People almost never look at the activity log. Wouldn't it be better not to try to fetch it at all until someone opens the activity view?

packages/workshop-shared/src/api.ts
kentonv: To expand on this, we should provide whatever the UI needs exactly. If the UI filters pending actions out of the activity log then I guess that's the right thing to do here but I'd be a bit surprised by that -- I'd expect the activity log shows both pending and settled actions.

packages/workshop-backend/src/overseer.ts
ndisidore: This comment could use some work but what is loaded speculatively is the _pending_ actions set, which is always needed to drive the pending action badge and gadget hook state indicators.
History is only pulled on activity view (via the hook)

packages/workshop-backend/src/overseer.ts
ndisidore: Correct: this is a :bug:  squash

packages/workshop-shared/src/api.ts
ndisidore: Correct, "all" mean _all resolved_ (this wasn't something I coined - it was already this way on main)

> If the UI filters pending actions out of the activity log then I guess that's the right thing to do here but I'd be a bit surprised by that

It genuinely does even on current main. pending has its own tab with its own live-updating store.
So this split is exactly what is necessary for each view :slightly_smiling_face: 

If you want I can rename "all" to "resolved" to make this more clear

packages/typed-storage/src/index.ts
ndisidore: Done!

packages/workshop-backend/src/overseer.ts
ndisidore: ✔️  added a `byHistoryFilter` index keyed by the wire filter values, so every filter is one ranged reverse read; pages are full until the last and no more scan cap / empty-page-with-cursor behavior. 

One tweak vs. your sketch: pending records key as `["pending"]` and resolved as `["all", type]`, since the type filters mean resolved history in the UI.

packages/workshop-backend/src/overseer.ts
ndisidore: Updated comment

packages/workshop-backend/src/overseer.ts
ndisidore: done :heavy_check_mark:

packages/typed-storage/src/index.ts
ndisidore: Added the new method :heavy_check_mark:

packages/typed-storage/src/index.ts
```diff
@@ -96,7 +117,16 @@ export interface Subscriber<T> {
   remove(record: T): void;
 }
 
-export interface Collection<T extends object, PrimaryKey = string> extends UniqueIndex<T, PrimaryKey> {
+/**
+ * A collection of records addressed by primary key. Note: NOT `extends UniqueIndex` -- a
+ * collection has the same get/list/delete, but rebuild() (re-deriving an index from the records)
+ * is meaningless for the records themselves.
```
kentonv:
```suggestion
 * A collection of records addressed by primary key.
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -1758,6 +1809,41 @@ class OverseerImpl implements AgentHooks {
...
+    });
+    this.logger.info("backfilled the pending-action index", {
+      event: "storage.migration.pending-index.completed", durationMs: Date.now() - startedAt,
+    });
+  }
+
+  // Version 3 -> 4: backfill the actions `byHistoryFilter` index; same shape and rationale as
+  // #migrateToPendingActionIndex above. The backfill writes ~2 tiny index rows per historical
+  // record in one transaction -- an O(log) one-shot, the same class as the version-3 backfill.
+  #migrateToHistoryFilterIndex(): void {
```
kentonv: We don't need a second migration here -- merge the two.

(Your agent probably conservatively assumed that the previous migration had already been rolled out. Common agent misunderstanding.)

packages/workshop-backend/src/overseer.ts
```diff
@@ -1758,6 +1809,41 @@ class OverseerImpl implements AgentHooks {
...
+  // never-initialized DOs write-free (they stamp the current version at first initialization).
+  #migrateToPendingActionIndex(): void {
+    if (this.storage.version.get() !== 2) return;
+    let startedAt = Date.now();
+    this.ctx.storage.transactionSync(() => {
+      this.storage.actions.pendingByGatekeeper.rebuild();
+      this.storage.version.put(3);
+    });
+    this.logger.info("backfilled the pending-action index", {
+      event: "storage.migration.pending-index.completed", durationMs: Date.now() - startedAt,
```
kentonv: Note that `durationMs` will always be logged as zero due to spectre mitigations.

packages/workshop-shared/src/api.ts
ndisidore: we made the call to deviate from existing functionality and now show pending actions in the history view _in addition_ to the needs attention view.
index is updated appropriately

packages/workshop-backend/src/overseer.ts
```diff
@@ -9759,17 +9841,31 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
...
+    // listActions({filter: "pending"}) after initiating the subscribe (see api.ts).
     if (startAfter !== undefined) {
-      let startAfterTimestamp = startAfter.valueOf();
-      for (let record of actions.list()) {
-        if (disposed) break;
-        let appliedAt = record.type === "action" ? record.appliedAt : undefined;
-        let recordTimestamp = (appliedAt ?? record.createdAt).valueOf();
-        if (recordTimestamp > startAfterTimestamp) {
-          subscriber.entry(actionRecordToLog(record)).catch(unsubscribe);
+      // startAfter supports resubscribing after a disconnect: replay every record whose last
```
kentonv: OK, looking closer at this code, I see now why you (or your agent) is so insistent on changing it even when I keep asking to keep it as it is: The old implementation is quite terrible, listing the whole history from the start in order to find the cutoff point, presumably because actions are not indexed on date. Ouch.

OK, fine, that does indeed need to change.

However, the new implementation here actually isn't any better. It is still iterating from the start of history and filtering stuff that is too old. It loads a page at a time, but this actually doesn't help anything since if the whole page is filtered then there ends up being nothing to wait for, and it immediately begins the next page.

The original proposal was to deprecate `startAfter` entirely and instead use `listActions()` to backfill after a disconnect, but that actually doesn't even work: the reason `startAfter` is a date rather than an action ID is because it also needs to update the subscriber about actions that have changed state, hence it also looks for `appliedAt > startAfter`.

Ugh, we may have no choice here except to index on "last modification time" as well?

packages/workshop-backend/src/overseer.ts
```diff
@@ -1076,7 +1097,24 @@ export function makeOverseerStorage(storage: DurableObjectStorage) {
...
+          pendingByGatekeeper(record: ActionRecord) {
+            return record.state === "pending" ? record.gatekeeperId : null;
+          },
+
+          // Total index keyed by the wire ActionHistoryFilter values, in lockstep with
+          // matchesActionHistoryFilter (api.ts), so every listActions() filter is one ranged
+          // read. Backfilled by the version-4 migration.
+          byHistoryFilter(record: ActionRecord) {
+            return record.state === "pending"
+                ? ["pending", "all", record.type] : ["all", record.type];
```
kentonv: We don't need to index on all "all" anymore. We can just query the collection directly in the "all" case.

```suggestion
            return record.state === "pending" ? ["pending", record.type] : record.type;
```

packages/workshop-backend/src/overseer.ts
ndisidore: bingo. this is what we've been butting up against.

> Ugh, we may have no choice here except to index on "last modification time" as well?

I too was trying to avoid that but I think it is necessary to make things work as intended. The good news is it's pretty cheap: we just need to make it a compound index so avoid frozen time collisions.

Actioned and lookups/resumes should be very cheap now

kentonv (approved):

PR #299, opened 2026-08-21
emphasize that executeCode return values are discarded

### LINKED ISSUE #209: executeCode can drop module return values
## What happened

During a live Cloudflare OS session, several `executeCode` steps completed with no visible output.
That symptom can have more than one cause; this issue does not claim that every blank tool step has
the same cause.

One independently reproducible case is a code-mode module that returns a value without calling
`console.log()`. The dynamic Worker harness awaits the module but does not return its value, while
`executeCodeMode()` formats only the console trace. The invocation therefore completes with blank
output.

## Minimal reproduction

Run a code-mode module containing only:

```js
export default async function() {
  return { ok: true };
}
```

On the current code path, the invocation completes but the returned object is absent from the tool
output.

## Expected behavior

The module return value reaches `executeCode` output alongside any console output. `undefined`
keeps the existing console-only behavior.

## Scope correction

An earlier version of this issue also claimed that `verify()` and `run()` tail events could arrive
in one batch. I did not capture such a batch in production; a synthetic unit invocation does not
establish that the runtime emits it. That unsupported claim has been removed, and #211 has been
closed. The return-value behavior above is independent of that hypothesis and is addressed by #210.

maxwellpeterson:
I often see the Workshop agent try to return values from executeCode calls, only to realize afterward that return values are discarded and it should use console.log() instead. We should emphasize this behavior in the executeCode tool description so the agent gets it right the first time.

Fixes https://github.com/cloudflare/cloudflare-os/issues/209

### DIFF
--- a/packages/workshop-backend/src/agent.ts
+++ b/packages/workshop-backend/src/agent.ts
@@ -750,6 +750,8 @@ Note that this differs from the \`env\` a Gadget's own code sees: a Gadget's ser
 When the user asks you to just do a task that can be done with these bindings, you should use executeCode to perform the task, instead of adding code to a gadget to do it.
 
 The function also receives a \`self\` parameter which is a magic object that points back to this chat thread. Calling any method on \`self\`, like \`self.foo(123)\`, delivers a callback message to this chat and activates you to respond. \`self\` can be passed over RPC (e.g. to a subscription method) and stored in a Durable Object's KV storage for long-term callbacks. When an agent callback is received, it appears in your env under a name like \`PARAMS_1\`, with \`.args\` (the callback arguments), \`.resolve(value)\` (to return a value to the caller), and \`.reject(error)\` (to reject with an error).
+
+The function's return value is discarded and will not appear in the tool result. Use \`console.log()\` for every value you need to inspect. Before the function returns, explicitly log the result of the operation when it will help confirm what happened.
 `.trim();
 
 let LIST_CONNECTABLE_RESOURCES_TOOL_DESCRIPTION = `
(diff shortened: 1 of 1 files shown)


packages/workshop-backend/src/agent.ts
```diff
@@ -831,6 +750,8 @@ Note that this differs from the \`env\` a Gadget's own code sees: a Gadget's ser
 When the user asks you to just do a task that can be done with these bindings, you should use executeCode to perform the task, instead of adding code to a gadget to do it.
 
 The function also receives a \`self\` parameter which is a magic object that points back to this chat thread. Calling any method on \`self\`, like \`self.foo(123)\`, delivers a callback message to this chat and activates you to respond. \`self\` can be passed over RPC (e.g. to a subscription method) and stored in a Durable Object's KV storage for long-term callbacks. When an agent callback is received, it appears in your env under a name like \`PARAMS_1\`, with \`.args\` (the callback arguments), \`.resolve(value)\` (to return a value to the caller), and \`.reject(error)\` (to reject with an error).
+
+The function's return value is discarded and will not appear in the tool result. Use console.log() for every value you need to inspect. Before the function returns, explicitly log the result of the operation when it will help confirm what happened.
```
ndisidore: Possibly worth condensing to what I had originally
```
The exported function's return value is ignored; use \`console.log()\` to produce tool output.
```
either way would suggest wrapping in \`console.log()\` to be consistent with the rest of the block

maxwellpeterson: Added backtick wrappers. I prefer the current prose and feel like it matches the rest of this block better

maxwellpeterson:
Closing in favor of https://github.com/cloudflare/cloudflare-os/pull/300

PR #300, opened 2026-08-21
log executeCode return values

### LINKED ISSUE #209: executeCode can drop module return values

maxwellpeterson:
Kimi likes to return values when using the `executeCode` tool instead of logging them. Return values are currently discarded, which causes confusion. We can automatically log them instead.

This is probably a better alternative to https://github.com/cloudflare/cloudflare-os/pull/299

Fixes #209

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/300" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/packages/gatekeeper-scheduler/src/types.d.ts
+++ b/packages/gatekeeper-scheduler/src/types.d.ts
@@ -280,9 +280,9 @@ export interface ScheduledTaskHook {
  * runs through the restored Gadget and does not need the Scheduler binding.
  *
  * Every successful registration call creates a distinct hook; registration is not idempotent.
- * `executeCode` ignores the exported function's return value, so print the schedule ID with
- * `console.log()`. Empty console output is not evidence that registration failed, and the disabled
- * hook appearing in Connections confirms success. Do not retry solely because no ID was printed.
+ * Return or print the schedule ID so it appears in the `executeCode` output. Empty output is not
+ * evidence that registration failed, and the disabled hook appearing in Connections confirms
+ * success. Do not retry solely because no ID was printed.
  *
  * @example
  * // server.js
@@ -308,7 +308,7 @@ export interface ScheduledTaskHook {
  *   callback,
  *   { title: "Daily brief", description: "Prepare the morning activity summary." },
  * );
- * console.log("Schedule registered:", scheduleId);
+ * return scheduleId;
  */
 export interface ScheduleSession {
   /**
--- a/packages/gatekeeper-scheduler/src/types.d.ts
+++ b/packages/gatekeeper-scheduler/src/types.d.ts
@@ -305,7 +305,7 @@ export interface ScheduledTaskHook {
  *   callback,
  *   { title: "Daily brief", description: "Prepare the morning activity summary." },
  * );
- * return scheduleId;
+ * console.log("Schedule registered:", scheduleId);
  */
 export interface ScheduleSession {
   /**
--- a/packages/gatekeeper-scheduler/src/types.d.ts
+++ b/packages/gatekeeper-scheduler/src/types.d.ts
@@ -280,9 +280,6 @@ export interface ScheduledTaskHook {
  * runs through the restored Gadget and does not need the Scheduler binding.
  *
  * Every successful registration call creates a distinct hook; registration is not idempotent.
- * Return or print the schedule ID so it appears in the `executeCode` output. Empty output is not
- * evidence that registration failed, and the disabled hook appearing in Connections confirms
- * success. Do not retry solely because no ID was printed.
  *
  * @example
  * // server.js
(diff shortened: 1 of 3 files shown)


packages/workshop-backend/src/overseer.ts
```diff
@@ -99,7 +99,8 @@ export default class extends WorkerEntrypoint {
         }
       }
     }
-    await agent(self, env, this.ctx);
+    let result = await agent(self, env, this.ctx);
+    if (result !== undefined) console.log(result);
```
ndisidore: If the agent is doing something sensitive like fetching the body of a secure document, won't this mean its captured in chat output? There's no way to opt in/out?

maxwellpeterson: I'm not sure why the agent would return a value it didn't want to appear in the tool output. I added a line to the tool description to make this behavior more explicit, but I don't think we need an opt in/out mechanism here.

kentonv: Should we prefix this with some text explaining it is the return value of the function, to disambiguate from any logs appearing before that?

packages/workshop-backend/src/agent.ts
```diff
@@ -831,6 +831,8 @@ Note that this differs from the \`env\` a Gadget's own code sees: a Gadget's ser
 When the user asks you to just do a task that can be done with these bindings, you should use executeCode to perform the task, instead of adding code to a gadget to do it.
 
 The function also receives a \`self\` parameter which is a magic object that points back to this chat thread. Calling any method on \`self\`, like \`self.foo(123)\`, delivers a callback message to this chat and activates you to respond. \`self\` can be passed over RPC (e.g. to a subscription method) and stored in a Durable Object's KV storage for long-term callbacks. When an agent callback is received, it appears in your env under a name like \`PARAMS_1\`, with \`.args\` (the callback arguments), \`.resolve(value)\` (to return a value to the caller), and \`.reject(error)\` (to reject with an error).
+
+A non-\`undefined\` value returned by the function is included as the final line of console output, so only return values that should be visible in this output.
```
kentonv: I think we should leave this out. Best not to include text in the system prompt unless the agent really needs to know it. Ideally we want agents to use console.log() and not rely on returns, but we're logging the return value as a courtesy since we see agents commonly do the wrong thing.

I don't think @ndisidore's concern about returning sensitive information is something to worry about. If an agent returns anything here, the only possible intent I can imagine is that they thought it would become part of `executeCode`'s output.

packages/gatekeeper-scheduler/src/types.d.ts
```diff
@@ -280,9 +280,9 @@ export interface ScheduledTaskHook {
  * runs through the restored Gadget and does not need the Scheduler binding.
  *
  * Every successful registration call creates a distinct hook; registration is not idempotent.
- * `executeCode` ignores the exported function's return value, so print the schedule ID with
- * `console.log()`. Empty console output is not evidence that registration failed, and the disabled
- * hook appearing in Connections confirms success. Do not retry solely because no ID was printed.
+ * Return or print the schedule ID so it appears in the `executeCode` output. Empty output is not
+ * evidence that registration failed, and the disabled hook appearing in Connections confirms
+ * success. Do not retry solely because no ID was printed.
```
kentonv: This text is weird. Why are we explaining the UI to the agent?

Did we actually have problems with agents getting confused and thinking that schedules had failed to be registered? If so, is the solution perhaps to simply emphasize that if the call doesn't throw, then the schedule has been registered?

Why does an agent care about the schedule ID at all? What can it do with this? Why are we telling it to log or return it?

ndisidore: The schedule id can be used by the agent to confirm the schedule exists via `list()`
Non-SOTA models would often get confused because they would call `executeCode` but to them nothing happened - it would just get a void output.
The would interpret this as a fail and attempt to re-register (often many times)

This is probably less important now that it won't be a void return, but returning the schedule id may still be useful so it knows how to check if it is still enabled.

packages/workshop-backend/src/agent.ts
maxwellpeterson: Dropped this line in 8aa8e99f4cc5ef0d4eb6064e413ea46e70572965

packages/gatekeeper-scheduler/src/types.d.ts
kentonv: When you experienced this, were those non-SOTA models specifically trying to return the schedule ID? If so, maybe simply the fact that we're logging now will mean that they no longer get confused, without the need for this text?

packages/workshop-backend/src/overseer.ts
maxwellpeterson: Added a "Return value: " prefix in d9541415632ffc3c2670823ea7b9ca6a0e6d20ec

packages/gatekeeper-scheduler/src/types.d.ts
kentonv: Another thing we could do is: If the executeCode logs _nothing_, we could write "(function succeeded with no output)"

kentonv (approved):
Sorry for slow response.

PR #306, opened 2026-08-23
Fix observer verification gaps

Maximo-Guk:
WIP


Maximo-Guk:
The rework landed as #345

PR #309, opened 2026-08-23
Build "deploy" releases in parallel

Maximo-Guk:
Every worker bundle in a release is independent, they can be built in parallel which speeds up building the release manifest significantly.

`node scripts/release/build-release.ts`, on my machine (10 cores), medians of 4 runs:

| | main | this PR |
|---|---|---|
| wall clock | 61.3s | **18.9s** |
| range | 57.0–65.3s | 15.8–25.6s |

Both produce an identical release (18 workers, 87 modules, 29 unique asset blobs). The frontend `vp` build was a cache hit in every run

### DIFF
--- a/scripts/kill-process-tree.test.ts
+++ b/scripts/kill-process-tree.test.ts
@@ -1,7 +1,7 @@
 import assert from "node:assert/strict";
 import { spawn } from "node:child_process";
 import { describe, it } from "node:test";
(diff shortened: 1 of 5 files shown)


ndisidore (approved):
Makes sense!

maxwellpeterson (approved):

PR #317, opened 2026-08-24
Fix blueprint model selector layering

dcartertwo:
AI model options now render above the blueprint configuration dialog instead of behind it, where pointer selection was impossible. The same behavior works with the mobile bottom sheet.

| Before | After |
| --- | --- |
| Model options rendered behind the configuration dialog | Model options rendered above the configuration dialog |


Maximo-Guk:
One GPT finding:

- Medium: useDialogSelectPortalContainer.ts:11 mounts options beside Kumo’s modal portal. Kumo applies aria-hidden="true" to this sibling, hiding all model options from screen readers. Mount the host within the dialog portal hierarchy and assert no option has an aria-hidden ancestor.

Maximo-Guk (approved):

PR #319, opened 2026-08-24
Add local workerd evals for the Workshop agent

AshishKumar4:
This adds local end-to-end evals for the Workshop agent. `pnpm evals` starts the real Workshop backend, Durable Objects and Worker Loader under workerd; creates a fresh account and workspace; sends a prompt to the production agent; then calls the generated Gadget's RPC to check whether the result works.

Only model inference leaves the local process. The suite does not connect to a deployed Workshop, use Cloudflare Access or change the Workshop backend/shared API.

[Local Workshop agent eval architecture](https://raw.githubusercontent.com/cloudflare/cloudflare-os/9e5bbfb2f1438e04a88e2dde4b1542a888cafa0b/docs/images/workshop-evals-architecture.svg)

## How one trial runs

1. `integration-tests` starts a local Workshop with Gadget execution enabled.
2. `AgentSession` signs up a fresh user, selects the model, completes onboarding and opens a workspace.
3. The production agent receives the task prompt and uses its normal tools.
4. The driver waits until `activeAgent` has gone from active to idle and stayed idle for two seconds.
5. The verifier opens the generated Gadget's provisional RPC and runs deterministic checks.
6. `vitest-evals` retains the transcript, evidence, raw measurements, score and trace.
7. The session and local workerd process are disposed.

The idle window covers the top-level agent turn. Persistent callbacks that begin after that window are not part of the current driver; none of the included tasks uses them.

## `AgentSession`

`packages/integration-tests/src/agent-session.ts` is a reusable E2E driver, not scoring code. `packages/workshop-evals` only defines tasks, checks, scores and reports.

### Creating a session

`AgentSession.create()` drives the real local Workshop API:

```text
connect to /api over Cap’n Web
→ allocate and sign up a fresh local user
→ add direct Workers AI configuration when needed
→ list and select the requested model
→ complete first-time onboarding
→ create a fresh workspace
→ subscribe to chat updates
→ subscribe to workpieces
→ wait for bundled output formats
```

The subscriptions start before the first prompt, so a fast agent cannot complete an active-to-idle transition before the driver sees it.

### Running a turn

The first prompt calls `newChat()`; later prompts call `sendChatMessage()` with the same chat ID. The driver does not treat the initial idle state as completion. It must first observe `activeAgent`, then observe it disappear, then see no new active agent for two seconds.

The turn also ends immediately if the chat posts an `error` message. That rule applies before or during the active turn; the driver does not try to classify an error as recoverable. Timeout or cancellation calls `stopAgent()`.

The two-second idle window covers the top-level turn. A persistent callback that starts after the window is outside the current contract. None of the included tasks uses persistent agent callbacks.

After the turn, the session loads every compacted chat-history page in canonical order, reads final chat metadata, and returns:

```ts
{
  history: AiChatMessage[];
  workpieces: WorkpieceSummary[];
  usage: {
    lastStepTokens?: number;
    agentChatCostUsd?: number;
  };
}
```

`lastStepTokens` is intentionally scoped to the final provider-reported model step. It is not a trial token total.

### Checking the generated Gadget

The agent’s changes remain on its provisional chat branch. `connectToGadget()` obtains the workpiece’s `GadgetClient` and connects to the server for that exact chat branch. Checks therefore exercise what the agent just built before a user accepts it.

Disposal releases the chat and workpiece subscriptions, callback stubs, authenticated capabilities and WebSocket. The local target then closes workerd. A failed turn cannot be reused, but its workspace can still be deleted.

[Production agent loop inside an eval trial](https://raw.githubusercontent.com/cloudflare/cloudflare-os/9e5bbfb2f1438e04a88e2dde4b1542a888cafa0b/docs/images/workshop-evals-agent-loop.svg)

## Evals in this PR

| Eval | User turns | Checks |
| --- | ---: | --- |
| `appointment-desk` | 1 | Capacity, cancellation, named errors, simultaneous bookings and duplicate booking IDs. Five checks. |
| `expense-ledger` | 2 | Exact-cent splits, balances, settlement, validation, monthly totals, budgets and preserving data across the follow-up. Nine checks. |
| `project-doc` | 1 | A real standard Doc with the requested heading, list structure and content. Two checks. |

### Appointment desk

The verifier calls the generated `Desk` RPC to check:

1. Sequential bookings stop at capacity and a duplicate slot ID returns `DUPLICATE_SLOT`.
2. Cancellation frees a place and an unknown booking returns `UNKNOWN_BOOKING`.
3. Booking a missing slot returns `UNKNOWN_SLOT`.
4. Ten concurrent booking calls accept exactly three places and reject the rest as `SLOT_FULL`.
5. Concurrent calls sharing one booking ID accept exactly one and return `DUPLICATE_BOOKING` for the rest.

The concurrency check does not inspect synchronization code. A SQL constraint, synchronous storage or another correct implementation can pass.

### Expense ledger

Turn one adds real expenses through the generated `Ledger` RPC and checks:

- uneven integer-cent splits do not lose a cent
- all expected people appear and balances sum to zero
- settlement uses known people, clears every balance and stays short
- invalid inputs return the promised codes without changing state
- monthly totals include only the requested month

Turn two asks the same agent to add budgets, then checks:

- turn-one expenses still exist
- old expenses count as `uncategorized`
- replacing a budget works and remaining value can be negative
- duplicate-ID, balance and settlement behavior from turn one still works

This is a real multi-turn regression check. In the measured GLM 5.2 run, the second code change lost the first turn’s stored expenses and the task scored `0.67`.

### Project Doc

The verifier checks that the workpiece is presented as the standard `document` output, then calls `getDocument()` and requires:

- the exact title `Harbour Refit`
- a heading containing “harbour” and “refit”
- exactly three list items
- scope, budget and timeline each with a sentence of detail

It checks the Docs RPC and output metadata, not visual polish.

### Scoring

The score is the plain fraction of deterministic checks that passed:

```text
score = passed checks / total checks
```

There is no model call in the judge. Every task gates by default. A failed check gives Vitest a nonzero exit code and names the failed checks. Set `WORKSHOP_EVAL_GATING_TASKS=''` only for an explicit reporting-only run.

## Run it

```sh
# AI Gateway
export CF_AI_GATEWAY=...
export CF_AI_GATEWAY_ACCOUNT_ID=...
export CF_AI_GATEWAY_API_TOKEN=...
pnpm evals

# Or direct Workers AI
export CLOUDFLARE_ACCOUNT_ID=...
export CLOUDFLARE_API_TOKEN=...
pnpm evals
```

Controls:

```sh
WORKSHOP_EVAL_MODELS='@cf/zai-org/glm-5.2' pnpm evals
WORKSHOP_EVAL_TRIALS=3 pnpm evals
WORKSHOP_EVAL_GATING_TASKS='appointment-desk,project-doc' pnpm evals
pnpm evals:ui
```

A turn is one user prompt and its top-level agent work. Turns in a task share one chat and workspace. A trial is one complete execution of all turns and checks with one model. Trial count repeats each task-and-model pair; it does not limit the model's internal turns or tool calls.

## What the report contains

Each trial keeps the raw values, not labels such as “recorded”:

- every check result and its evidence
- total wall time
- agent duration and verifier duration for each turn
- model-turn count, defined as canonical agent-authored messages
- tool-call count and tool-error count from those messages
- agent and provider errors
- tokens reported for the final model step
- cumulative agent-chat cost reported by the provider
- normalized user, assistant, tool-call and tool-result events
- trace start, finish and duration
- model, trial, Git commit and prompt-derived `taskVersion`

[Raw measurements and report paths](https://raw.githubusercontent.com/cloudflare/cloudflare-os/9e5bbfb2f1438e04a88e2dde4b1542a888cafa0b/docs/images/workshop-evals-measurement-reporting.svg)

### Actual report sample

This is the raw summary from a local `project-doc` trial with GLM 5.2 at this commit:

| Measurement | Value |
| --- | ---: |
| Score | `1.00` |
| Checks | `2 / 2` |
| Total duration | `32,968 ms` |
| Agent duration | `25,577 ms` |
| Verification duration | `4 ms` |
| Model turns | `4` |
| Tool calls | `5` |
| Tool errors | `0` |
| Last-step tokens | `32,163` |
| Agent-chat cost | `$0.09623464` |
| Agent/provider errors | `0` |

A condensed JSON excerpt from the same result:

```json
{
  "taskId": "project-doc",
  "model": "@cf/zai-org/glm-5.2",
  "trial": 1,
  "score": 1,
  "checks": { "passed": 2, "total": 2, "failed": [] },
  "timings": {
    "totalMs": 32968,
    "turns": [{ "agentDurationMs": 25577, "verificationDurationMs": 4 }]
  },
  "metrics": {
    "modelTurns": 4,
    "toolCalls": 5,
    "toolErrors": 0,
    "lastStepTokens": 32163,
    "agentChatCostUsd": 0.09623464,
    "errors": 0
  },
  "gitCommit": "9e5bbfb2f1438e04a88e2dde4b1542a888cafa0b",
  "taskVersion": "08fff91f7c31d27a1b87122f8ffb9c2e8bbcc44e857e1095e036ea417c12e887"
}
```

The complete JSON also contains the normalized transcript, each check's evidence and the trace spans. `lastStepTokens` deliberately does not claim to be total trial tokens. There is no LLM judge or Braintrust integration in this PR.

## Write another eval

Add `packages/workshop-evals/evals/<name>.eval.ts`:

```ts
const task = defineEvalTask({
  id: "counter",
  turns: [{
    prompt: "Build a Gadget named Counter with increment() and value() RPC methods.",
    verify: async verifier => {
      await verifier.check("increments", async () => {
        using counter = await verifier.connect<CounterApi>("Counter");
        await counter.increment();
        const value = await counter.value();
        return { pass: value === 1, evidence: { value } };
      });
    },
  }],
});

defineTaskEval(task);
```

Checks verify observable behavior, not a particular tool sequence, storage layout or implementation. Multi-turn tasks re-check earlier behavior that the requested change could break.

## CI

The manual **Workshop evals** workflow runs the same `pnpm evals` command in one GitHub Actions job. It starts local workerd, uploads `results.json` and publishes one GitHub report. Repository AI Gateway secrets are used only for model inference. The workflow does not deploy or contact a production Workshop.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/319" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

Wall time: 0.51 seconds


Maximo-Guk:
Initial GPT pass:

```
1. P1 Gating failures leave the workflow green. .github/workflows/workshop-evals.yml:48-50,93-99 tolerates shard failures, while the pinned reporter soft-fails when publishing a Check Run. Set soft-fail: false.
2. P1 Access tokens can be sent over plaintext WebSockets. packages/integration-tests/src/rpc-client.ts:59-67 sends CF_Authorization over ws://, and packages/workshop-evals/src/target.ts:44-51 permits HTTP preview URLs. Require HTTPS whenever an Access token is provided.
3. P1 Agent turn timeouts do not bound the initiating RPC. packages/integration-tests/src/agent-session.ts:157-172 waits for newChat()/sendChatMessage() before observing the timeout promise. A stalled RPC can therefore hang indefinitely despite the advertised hard limit.
4. P1 Failed preview turns cannot delete their workspace. agent-session.ts:187-189,217-220,258-260 marks the session unusable after an error, then rejects deleteWorkspace(). The disposer in packages/workshop-evals/src/target.ts:108-113 consequently leaks the preview workspace.
5. P1 Gadget verification has no deadline. packages/workshop-evals/src/harness.ts:35-40 and src/verifier.ts:69-82 allow a generated RPC to remain pending beyond Vitest’s timeout. The async disposal scope never unwinds, leaking the session/workspace.
6. P2 The final ledger regression check passes vacuously. packages/workshop-evals/evals/expense-ledger.eval.ts:343-363 accepts empty balances() and settlement() arrays, allowing broken turn-one methods to receive a perfect score.
7. P2 Reported total tokens contain only the last model step. packages/workshop-evals/src/harness.ts:31-54 overwrites usage each turn and publishes AiChatMetadata.totalTokens, which explicitly represents only the conversation’s last step.
8. P2 The 500 ms idle debounce can finish before a callback turn starts. packages/integration-tests/src/agent-session-internals.ts:71-77 treats temporary inactivity as completion, but callback continuation performs asynchronous context loading before restoring activeAgent.
```

Maximo-Guk:
Second GPT pass on efficiency:

```
1. High: Failed preview trials cannot delete their workspace. run() marks the session failed, while cleanup calls deleteWorkspace(), which rejects failed sessions. This leaks workspaces and may mask the original failure. packages/integration-tests/src/agent-session.ts:187-220, packages/workshop-evals/src/target.ts:108-113.
2. High: Adding .concurrent now is unsafe and inefficient. Every trial starts another full Wrangler/workerd harness, whose custom build writes shared generated files. Share one harness per file/shard, then create a fresh AgentSession per trial. packages/workshop-evals/src/target.ts:139-147, packages/workshop-backend/wrangler.jsonc:4-7.
3. Medium: These are not hermetic like the integration tests. The integration suite installs NetworkInterceptor and rejects all unexpected egress; evals allow unrestricted outbound traffic, including agent webFetch calls. Permit only required AI Gateway/inference and cost endpoints, rejecting everything else. Preview evals cannot be made hermetic this way. packages/integration-tests/src/network-interceptor.ts:1-6, packages/workshop-evals/src/target.ts:118-139.
4. Medium: The cleanup timeout reserve is not enforced. Only agent turns are timed; setup and verifier RPCs can consume the outer Vitest timeout and prevent cleanup. Use one absolute trial deadline and propagate its remaining time/signal into turns and verification. packages/workshop-evals/src/harness.ts:18-40, packages/workshop-evals/src/config.ts:9-14.
5. Medium: Scoring does not exactly match the prompts. DUPLICATE_SLOT is required but untested; the expense UI requirements are untested; the Doc structure check only verifies word presence. Conversely, the full Doc rewrite tests platform behavior not explicitly requested. appointment-desk.eval.ts:56-58, expense-ledger.eval.ts:80-86, project-doc.eval.ts:61-93.
6. Low: Every session waits for output formats although only project-doc requires them. Gateway mode also performs an unmeasured title-generation inference per trial. Make format readiness task-specific and disable/mock auxiliary title generation. packages/integration-tests/src/agent-session.ts:133-135, packages/workshop-backend/src/overseer.ts:5238-5242.
```

packages/integration-tests/src/rpc-client.ts
```diff
@@ -43,11 +47,27 @@ export function nextUsernames(...prefixes: string[]): string[] {
...
+    baseUrl: URL, options: WorkshopConnectionOptions = {}): RpcStub<PublicApi> {
   const wsUrl = new URL("/api", baseUrl);
   wsUrl.protocol = wsUrl.protocol === "https:" ? "wss:" : "ws:";
-  return newWebSocketRpcSession<PublicApi>(wsUrl.toString());
+  if (options.accessToken === undefined) {
+    return newWebSocketRpcSession<PublicApi>(wsUrl.toString());
+  }
+  const nodeSocket = new NodeWebSocket(wsUrl.toString(), {
+    origin: baseUrl.origin,
+    headers: { Cookie: `CF_Authorization=${options.accessToken}` },
```
ndisidore: this seems like like it could leak the token - previews can be http
probably want to enforce https when this is present

PR #320, opened 2026-08-24
wait for every connection request before resuming the agent

bjesus:
When the agent asks for several connections in one turn, accepting the first card resumed it right away, so it carried on as if the rest had been refused. Now the turn resumes only once every request it raised has been decided and all were accepted; a denial leaves the turn ended, the same rule an awaited action already follows.


ndisidore (approved):
thanks!

PR #325, opened 2026-08-25
Preserve Cloudflare OAuth token exchange errors

### LINKED ISSUE #323: Preserve Cloudflare OAuth token exchange errors
The Cloudflare Gatekeeper collapses every non-2xx token response into `null`, so `acceptAuthCode()` reports the same "exchange failed or returned no refresh token" error for invalid client credentials, rejected grants, and provider failures.

This makes production OAuth failures unnecessarily difficult to diagnose. In one deployment, the provider returned `401 invalid_client`, but the actionable cause was discarded before it reached Worker logs.

Expected behavior:

- Preserve the HTTP status and OAuth `error` / `error_description` fields for rejected token exchanges.
- Do not include client credentials, authorization codes, access tokens, or refresh tokens in the error.
- Keep successful token parsing unchanged.

The provider error payload is the standard OAuth response and gives operators enough information to distinguish client configuration from grant and scope failures.

atsaplin:
Closes #323

## What does this change?

Cloudflare OAuth token redemption currently collapses every non-2xx response to `null`. That makes invalid client credentials indistinguishable from a successful response that omitted an access token. This patch throws an error containing only the HTTP status and standard OAuth `error` and `error_description` fields, and adds a focused rejection-path test.

The change affects one non-2xx branch in the token exchange helper. The test supplies a 401 `invalid_client` response and asserts the exact sanitized error. Credentials and token values are never read into the message. Successful response handling is unchanged.

## Verification

- `pnpm --filter @gadgets/cloudflare-gatekeeper test:run`

## Checklist

### DIFF
--- a/packages/gatekeeper-cloudflare/src/oauth.ts
+++ b/packages/gatekeeper-cloudflare/src/oauth.ts
@@ -138,8 +138,18 @@ async function redeem(config: CloudflareOAuthConfig, body: URLSearchParams): Pro
     body,
   });
   if (!resp.ok) {
-    resp.body?.cancel();
-    return null;
+    const providerError: unknown = await resp.json().catch(() => null);
+    const errorCode = providerError !== null && typeof providerError === "object" &&
+        "error" in providerError && typeof providerError.error === "string"
+      ? ` ${providerError.error}`
+      : "";
+    const errorDescription = providerError !== null && typeof providerError === "object" &&
+        "error_description" in providerError && typeof providerError.error_description === "string"
+      ? `: ${providerError.error_description}`
+      : "";
+    throw new Error(
+      `Cloudflare OAuth token exchange failed (${resp.status}${errorCode})${errorDescription}`,
+    );
   }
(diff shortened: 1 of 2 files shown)


atsaplin:

packages/gatekeeper-cloudflare/src/oauth.ts
```diff
@@ -138,8 +138,18 @@ async function redeem(config: CloudflareOAuthConfig, body: URLSearchParams): Pro
     body,
   });
   if (!resp.ok) {
-    resp.body?.cancel();
-    return null;
+    const providerError: unknown = await resp.json().catch(() => null);
```
ndisidore, line 141: `redeem()` is also used by `refreshTokens()`. Previously a rejected refresh returned null, allowing `getAccessToken()` to call `credentialsExpired()`. with this change it now throws so an expired/revoked refresh token will keep failing without prompting reconnection.
If you are able to fix this happy to merge!

PR #331, opened 2026-08-25
add preview support to google oauth flow

maxwellpeterson:
Upgrade the OAuth state parameter into a signed JWT that can optionally carry a preview callback URL for completing the flow. Dynamic redirection is only allowed when the `OAUTH_ALLOW_PREVIEW_REDIRECTS` variable is set, and preview redirect URLs must be suffixed by the main deployment URL.

This allows a single callback URL to be registered with the OAuth application that covers all preview deployments.


packages/gatekeeper-google/src/google.ts
```diff
@@ -215,7 +223,21 @@ export default {
       let doId = path[0];
       let initiationNonce = path[1];
       let stub = ctx.exports.UserAccount.get(ctx.exports.UserAccount.idFromString(doId));
-      let begun = await stub.beginOAuthFlow(initiationNonce);
```
ndisidore, line 218: nit: you might be able to skip most of this and just go straight to
```ts
const begun = await stub.beginOAuthFlow(initiationNonce, oauthRedirectUri);
newUrl.searchParams.set("redirect_uri", oauthRedirectUri);
```

fetch() already has the immutable oauthRedirectUri. beginOAuthFlow() only needs to persist it with the nonce

ndisidore (approved):

PR #340, opened 2026-08-25
Govern sharing of restricted data by observer verification

Maximo-Guk:
Why:

Before this PR: Once a user has observed a sensitive data source, we block sharing the resource altogether.

We decided we can lift this restriction now that we have mechanisms in place to verify the observers permissions.

What:

This PR sets out to rename `prohibitAllSharing` to `containsRestrictedData` and makes it so that now when restricted data is observed, the workspace can still be shared by verifying the observers access against the resource by reusing many of the same mechanisms we had already introduced in the original observer verification PR.

authorizeCollaborator() is now responsible for being the single gate for checking whether observers have access to underyling data. Everything routes through here, including `receiveExternalMessage()` which previously only checked the role and never checked whether the observer had access to the underlying data.

We also decided to abort the DO in order to revoke the RPC capabilities, we already use the same mechanism when revoking collaborators and we can just the same mechanism here. We made this a little less disruptive by checking graph first to see if there's any other collaborators before aborting the DO

Testing:

I added extensive integration tests in `sensitive-observations.test.ts`, `observer-role-scope.test.ts` and `external-message-verification` ( even though it's actually used yet in any gatekeepers ) for testing the observer verification logic

I also have two preview links mantainers can test it out on

- [My own private bigquery dataset](https://maximo-preview-public-followups-gadgets.gadgets-staging.workers.dev/workspace/c7816a4fe46df5813692ad40666a2f6c074e8bf18b78f9a862adf005f9f39009#share=c39083e300e2b6580e2bae54ab648456)

- [A shared bigquery dataset](https://maximo-preview-public-followups-gadgets.gadgets-staging.workers.dev/workspace/fad56e0d35903a24ad941fc4a62342cd8f3e1bf8ed8dd72cf0cc6ff22b6ec7e6#share=de137e27896c37f7b2dbb0551873de6a)

Please note these previews are built off of #308 (which is stacked on this PR) since this PR contains no UI changes

Out of scope:

There were a few pre-existing issues which my agent flagged, I left/flagged them as TODO's in the PR.

Frontend behavior is unchanged here, the share modal edits are the rename only, so the UI still declines to share a workspace even though the server now permits it. The UI will be done in https://github.com/cloudflare/cloudflare-os/pull/308


packages/workshop-backend/src/overseer.ts
```diff
@@ -4975,30 +5136,61 @@ class OverseerImpl implements AgentHooks {
...
+  // Fire-and-forget: the callers are synchronous (bindWorkpiece) or already past their last write,
+  // and getSharingManager() is async, so failures are logged rather than left as an unhandled
+  // rejection. Failing to restart is fail-open for the widened scope, hence the `error` level.
+  //
+  // Note that ensureAmbientCapsules() calls addGatekeeper() from inside open(), so on a shared
+  // workspace the first open after an ambient capsule appears bounces itself once; the capsule
+  // exists by then, so the client's retry is clean.
+  #restartIfShared(reason: string): void {
+    this.getSharingManager().then(sharing => {
+      if (sharing.listCollaborators().length === 0) return;
```
Maximo-Guk, line 5206: This checks collaborators instead of just observers as my agent had a concern around collaborators who are parked mid-open() in the configuration model, and don't have a committed record yet.

Maximo-Guk:
Rebased off main

Maximo-Guk:
> **Findings**
> 
> 1. **P1** `overseer.ts:4527,7987,11404`: Restricted data from unbound producers can reach gadget state/UI, but `use` collaborators are never verified against those producers. Verify all restricted producers or enforce provenance isolation.
> 2. **P1** `gatekeeper-google/src/google.ts:983,3399`: BigQuery verification only calls `datasets.get`. This does not prove table-data access or equivalent row/column policy results, allowing collaborators to see owner-only query output.
> 3. **P1** `overseer.ts:4388-4407`: `addGatekeeper()` publishes the connection before awaiting `describe()`, but restarts collaborators afterward. Existing build sessions can access the predictable ID during this unbounded window without verification.
> 4. **P1** `overseer.ts:8200,8319-8328`: Failed re-verification scrubs coverage, then may wait indefinitely for the configuration callback and cleanup before scheduling restart. Previously opened sessions remain usable after revocation is detected.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/33180531453)


From scanning these findings with my agent:

1) Already documented as "Known security risk — never-bound producers"

2) This is addressed in #372 , we're only touching kernel changes here

3) This should be fixed in my [last commit](https://github.com/cloudflare/cloudflare-os/pull/340/commits/935430ff7698170be8448a92319d9fb22b26e372)

4) The re-prompt stall is equivalent to never re-opening (edge case 3)

packages/workshop-backend/src/overseer.ts
```diff
@@ -1394,6 +1404,21 @@ export function sanitizeMessageFormatRefs(
   return accepted.toSorted((a, b) => a.position - b.position);
 }
 
+// Action records that predate the flag's rename carry `containsRestrictedData` under its old
+// name, `prohibitAllSharing`. Records are data at rest and are never rewritten, so the
+// tolerance can never be removed.
+type LegacyObservationDescription = ObservationDescription & { prohibitAllSharing?: boolean };
```
kentonv, line 1407: This is pretty ugly. We should either do a migration to update all existing records, or we should continue using the old name in the meantime. (There's other precedent for continuing to use old names in the storage schemas already. I am sort of inclined to extend typed-storage to support specifying a "legacy name" for a collection/singleton/index/etc., in order to avoid migrations...)

Maximo-Guk:
Superseded by a three-way split, stacked in this order: #380

PR #341, opened 2026-08-26
Clean up loose ends in git-storage, improving crash recovery

kentonv:
Last week's changes to transition to git storage had some obscure issues.

When an agent message contained a write/edit tool call, the resulting edits would be persisted to `chatChanges` (which contains the OT change log) at the time the tool executed, but the overall agent message was not persisted to the actual chat history until the end of the step. A poorly-timed crash could leave the edit in `chatChanges` while the agent message was never persisted. When the DO started back up, the agent would run again, and very likely try to make the same changes again, resulting in the changes either being double-applied or confusingly failing.

We now instead buffer changes to an in-memory buffer, which gets persisted as part of the same transaction that persists the chat log message. Either everything gets persisted together, or nothing does.

Moreover, this means that the `chatChanges` contains ONLY user-authored edits, which is closer to how things worked back in the Yjs days. Agent-authored changes go into the in-memory buffer, and then are persisted directly to a "changes" message in the chat log.

These changes fell out of my upcoming work to enable agents to create git worktrees. In that work, the agent will gain the ability to edit files programmatically within executeCode. Such edits will also need to land in the in-memory buffer so that they can be persisted transactionally.

`plans/step-transactionality.md` contains a more complete description of this change.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/341" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/workshop-shared/src/code-change.ts
+++ b/packages/workshop-shared/src/code-change.ts
@@ -150,6 +150,45 @@ export const MAX_CODE_CHANGE_SIZE = 2 * 1024 * 1024;
 const FILE_ENTRY_SIZE_OVERHEAD = 16;
 const EDIT_SECTION_SIZE_WEIGHT = 4;
 
+// Structure overhead granted per file entry and per edit section by codeChangeSerializedSize:
+// V8 writes a handful of header/varint bytes per object, array, and string, so a generous
+// fixed share per node keeps the estimate an upper bound without measuring anything.
+const SERIALIZED_ENTRY_OVERHEAD = 32;
+const SERIALIZED_SECTION_OVERHEAD = 16;
+
+/**
+ * Upper bound on the bytes a `CodeChange` contributes to a V8-serialized storage record --
+ * the form Durable Object storage actually holds change rows and "changes" messages in,
+ * subject to its 2MB value cap. V8 writes each string raw (Latin-1 at one byte per UTF-16
+ * code unit, or UTF-16 at two) plus small structure headers, so two bytes per code unit plus
+ * a fixed per-entry/per-section overhead never undercounts; the estimate reads only string
+ * lengths, costing O(entries), and is immune to the escape-sequence and multi-byte inflation
+ * a JSON-text measure would suffer. Callers use it to keep *compositions* of changes inside
+ * one storable record; `MAX_CODE_CHANGE_SIZE` separately bounds a single change with its own
+ * unit-weighted measure at validation time.
+ */
+export function codeChangeSerializedSize(change: CodeChange): number {
+  let bytes = 0;
+  for (let entries of Object.values(change)) {
+    for (let [path, fileChange] of entries) {
+      bytes += SERIALIZED_ENTRY_OVERHEAD + 2 * path.length;
+      if ("set" in fileChange) {
+        bytes += 2 * fileChange.set.length;
+      } else if ("edit" in fileChange) {
+        for (let section of fileChange.edit) {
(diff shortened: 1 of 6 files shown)


packages/workshop-shared/src/code-change.ts
```diff
@@ -150,6 +150,45 @@ export const MAX_CODE_CHANGE_SIZE = 2 * 1024 * 1024;
...
+ * the form Durable Object storage actually holds change rows and "changes" messages in,
+ * subject to its 2MB value cap. V8 writes each string raw (Latin-1 at one byte per UTF-16
+ * code unit, or UTF-16 at two) plus small structure headers, so two bytes per code unit plus
+ * a fixed per-entry/per-section overhead never undercounts; the estimate reads only string
+ * lengths, costing O(entries), and is immune to the escape-sequence and multi-byte inflation
+ * a JSON-text measure would suffer. Callers use it to keep *compositions* of changes inside
+ * one storable record; `MAX_CODE_CHANGE_SIZE` separately bounds a single change with its own
+ * unit-weighted measure at validation time.
+ */
+export function codeChangeSerializedSize(change: CodeChange): number {
```
ndisidore, line 172: Would it be worth adding a test for this? The clankers are suggesting there is bug here
> `codeChangeSerializedSize` can drastically undercount edit sections made of many short/blank lines. Both new byte budgets trust its "never undercounts" guarantee, and with the chunking fallback deleted, an undercounted composition permanently wedges materialization — it retries the same oversized compose forever

Here's a test it says would fail
```ts
describe("codeChangeSerializedSize", () => {
  it("upper-bounds the V8-serialized size even for edits of many empty lines", async () => {
    // Each file inserts 512K-1 newlines as 512K empty lines: schema-valid (the
    // validator charges the separators), and every empty line costs >=2 bytes in
    // V8's wire format -- the estimator must not price them at zero.
    const LINES = 512 * 1024;
    const mkSection = (): TextChange => [[0, ...Array(LINES).fill("")]];
    const change: CodeChange = {
      1: [["a", { edit: mkSection() }], ["b", { edit: mkSection() }], ["c", { edit: mkSection() }]],
    };
    validateCodeChangeSchema(change); // in budget: separators cost ~1.5MB of the 2MB cap

    // Ground truth is the very write the estimate protects: a DO storage record.
    const stub = env.TEST_OVERSEER.getByName("size-estimate-demo");
    await runInDurableObject(stub, async (_instance, state) => {
      await expect(state.storage.put("change", change)).rejects.toThrow(/2097152|larger/);
    });

    // The record above did not fit, so the estimate must say it doesn't fit.
    expect(codeChangeSerializedSize(change)).toBeGreaterThan(MAX_CODE_CHANGE_SIZE);
  });
});
```

kentonv: I pushed a commit to fix this.

I did ask Claude to write a test, but because workerd doesn't expose any way to invoke V8 serialization directly (in order to find out the actual size), Claude wanted to write a test using node, and set up a bunch of new infrastructure for that. I decided it wasn't worth it. The test can only prove that we covered the cases we thought of, which is easy enough to judge by looking at the (pretty simple) code. The test can't really prove that there are no cases we failed to think of.

Honestly though the correctness of this little function isn't nearly as important as all these agents seem to think.

ndisidore (approved):

kentonv:
The test timeout in CI legitimately highlights some quadratic behavior introduced in this PR which I'm fixing now.

PR #344, opened 2026-08-26
Keep pre-existing observer registrations when re-verification fails

ndisidore:
:bug: fix. It's actually just 1 line (the rest is supporting tests/comments)

`ensureObserver` mutated its snapshot of which gatekeepers were registered before the call.
a repaired pre-existing binding got misclassified as newly added, and if any other binding then failed terminally, the rollback deleted it
if e.g. your Slack token expired and you re-authed when the open re-prompted you, but some other connection also failed, the failed open would roll back and delete your Slack registration too, even though you'd had it for weeks.  
The bad part is nothing crashes, but the observer record and the gatekeeper now disagree, so the gatekeeper can't exclude you from data it read before your next good open.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/344" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/workshop-backend/src/overseer.ts
+++ b/packages/workshop-backend/src/overseer.ts
@@ -7862,10 +7862,10 @@ class OverseerImpl implements AgentHooks {
     let record = this.storage.observers.get(profileId);
     let accountChoices: {[gatekeeperId: number]: number} = {...record?.accountChoices};
 
-    // Gatekeeper ids whose account choice came from the persisted record (vs. configured during
-    // this call). On a verification failure we only roll back observers we registered *this* call,
-    // leaving pre-existing registrations intact (rollback restores the pre-call state).
-    let preConfigured = new Set<number>(
+    // Gatekeeper ids registered before this call (their account choice came from the persisted
+    // record). An immutable snapshot: on a verification failure we roll back only observers we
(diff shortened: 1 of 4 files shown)


packages/workshop-backend/src/overseer.ts
```diff
@@ -7985,13 +7985,13 @@ class OverseerImpl implements AgentHooks {
...
-          // pass makes -- otherwise a gatekeeper could be left admitting an observer on a choice we
-          // never persisted.
+          // Drop the failed choices so the re-prompt asks about exactly these bindings. Failed
+          // bindings stay in `registeredBeforeCall`: a registration repaired on the re-prompt must
+          // survive a later rollback -- removing it would break excludeObservers while the
+          // persisted record still asserts it exists. Its unpersisted account choice self-corrects
+          // on the next open (re-verification fails the stale persisted choice and re-prompts).
           for (let id of failures.keys()) {
             delete accountChoices[id];
-            preConfigured.delete(id);
```
ndisidore, line 7994: This is, essentially, the entire fix

Maximo-Guk (approved):

kentonv (approved):

PR #351, opened 2026-08-26
Test Workshop sharing and presence over public RPC

AshishKumar4:
Stacked on #349, this adds deterministic functional coverage for collaborator roles, presence, share-link redemption, and revocation. The tests use the same local workerd Workshop and public Cap’n Web API as a browser client. They do not start an agent or call a real model.

## Use-only collaborator

```text
owner creates an active workspace
→ unrelated user is denied
→ owner grants use-only access
→ collaborator opens with role = use
→ collaborator cannot rename the workspace
→ owner removes the collaborator
→ future opens are denied
```

The denial assertions inspect the public `WORKSPACE_ACCESS_DENIED` code rather than matching an internal storage state.

## Share-link lifecycle

```text
owner creates one use-role share link
→ owner mints a second key for the same link
→ two users redeem the two keys
→ both open with role = use
→ revocation preview names both users
→ one revoke removes both permission edges
→ both keys fail on the next open
```

This checks the user-visible invariant that keys belong to a link and revoking the link revokes every key.

## Presence lifecycle

```text
owner grants build access
→ collaborator opens the workspace
→ presence init contains both users
→ collaborator closes the capability
→ presence emits removal
```

The presence callback target is created through `stubFor()` from the existing integration toolkit, ensuring the Cap’n Web stub belongs to the same RPC runtime as the session.

## Existing infrastructure

The stack continues to use:

- one shared `startHarness()` workerd instance per file
- fresh users and workspaces per concurrent case
- `connect()` and `signUp()` over the real WebSocket
- `NetworkInterceptor` with no allowed outbound traffic
- public permission, share-link, and presence capabilities

The only toolkit change is re-exporting its canonical `RpcTarget` so callback implementations do not import a second Cap’n Web runtime.

## Failure diagnostics

The three tests are named after their public contracts:

```text
grants and revokes a use-only collaborator
revokes every key and recipient of one share link
reports collaborator presence and removes it when the capability closes
```

Failures include the test name, exact assertion, expected and received roles or user IDs, stable open-error code, or the `waitFor()` state that never arrived.

## Files

```text
packages/integration-tests/__tests__/workshop-sharing.test.ts
packages/integration-tests/__tests__/workshop-presence.test.ts
packages/integration-tests/src/rpc-client.ts
```

Run the focused stack with:

```sh
pnpm exec vp run -F @gadgets/workshop-backend build
pnpm --filter @gadgets/integration-tests test:run -- \
  __tests__/workshop-sharing.test.ts \
  __tests__/workshop-presence.test.ts
```

No Workshop production code or public API changes are included.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/351" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

Wall time: 0.56 seconds

Wall time: 0.79 seconds

### DIFF
--- a/packages/integration-tests/__tests__/workshop-presence.test.ts
+++ b/packages/integration-tests/__tests__/workshop-presence.test.ts
@@ -1,0 +1,73 @@
+import { afterAll, beforeAll, expect, it } from "vitest";
+import type { PresenceParticipant, PresenceSubscriber } from "@gadgets/workshop-shared/api";
+import { type Harness, startHarness } from "../src/harness.js";
+import { NetworkInterceptor } from "../src/network-interceptor.js";
+import { mockChatCompletion } from "../src/mock-model.js";
+import { RpcTarget, connect, nextUsernames, signUp, stubFor, waitFor } from "../src/rpc-client.js";
+
+class PresenceRecorder extends RpcTarget implements PresenceSubscriber {
+  readonly participants = new Map<string, PresenceParticipant>();
+  init(participants: PresenceParticipant[]): void {
+    this.participants.clear();
+    for (const participant of participants) this.participants.set(participant.key, participant);
+  }
+
+  add(participant: PresenceParticipant): void {
+    this.participants.set(participant.key, participant);
+  }
+
+  remove(key: string): void {
+    this.participants.delete(key);
+  }
+}
+
+let harness: Harness | undefined;
+const network = new NetworkInterceptor([mockChatCompletion("Test chat")]);
+
+beforeAll(async () => {
+  network.install();
+  harness = await startHarness({ gatekeepers: [] });
+});
+
+afterAll(async () => {
+  try {
+    await harness?.server.close();
+    expect(network.getUnmockedCalls()).toEqual([]);
+  } finally {
+    network.uninstall();
+  }
+});
+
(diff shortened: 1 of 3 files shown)


packages/integration-tests/__tests__/workshop-presence.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+beforeAll(async () => {
+  network.install();
+  harness = await startHarness({ gatekeepers: [] });
+});
+
+afterAll(async () => {
+  try {
+    await harness?.server.close();
+    network.takeUnmockedCalls("/workers-ai/v1/chat/completions");
```
Maximo-Guk: Same comment here https://github.com/cloudflare/cloudflare-os/pull/349#discussion_r3867118522

Maximo-Guk (approved):

packages/integration-tests/__tests__/workshop-presence.test.ts
AshishKumar4: Yep, fixed. Both the sharing and presence tests now use the same `mockChatCompletion()` handler from the base PR. I tested them with a fake root `.dev.vars`; the title requests were handled locally and there were no unhandled outbound requests.

PR #361, opened 2026-08-27
Add local Workshop trajectory evals

AshishKumar4:
This PR adds `pnpm evals` on top of `WorkshopAgentSession`. An eval defines one or more prompts and deterministic checks against the Gadget the agent created. The runner starts an isolated Workshop under local workerd, runs each model and trial, calls the generated Gadget through its RPC API, and cleans up the workspace afterward.

The report shows whether the task passed, which checks failed, and the evidence returned by each check. It also records the canonical conversation, timings, model turns, tool calls, tool errors, final-step tokens, observed chat cost, model, trial, Git commit, and prompt hash.

The first eval asks the agent to create a Project Doc. Only model inference leaves the machine. This PR does not add a production target, Braintrust, or an LLM quality judge.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/361" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

Wall time: 0.45 seconds

Wall time: 2.19 seconds

Wall time: 0.48 seconds

### DIFF
--- a/packages/integration-tests/src/agent-session.ts
+++ b/packages/integration-tests/src/agent-session.ts
@@ -1,8 +1,8 @@
 import type { RpcPromise, RpcStub } from "capnweb";
 import type {
   ActionHistoryFilter, ActionHistoryPage, AiChatAuthorInfo, AiChatHistoryPage, AiChatMessage,
-  AiChatMetadata, AiChatStreamEvent, AiChatSubscriber, AiModelConfig, AuthenticatedApi, Overseer,
-  PublicApi, WorkpieceId, WorkpieceSummary, WorkpiecesSubscriber,
+  AiChatMetadata, AiChatStreamEvent, AiChatSubscriber, AiModelConfig, AuthenticatedApi, GadgetClient,
+  Overseer, PublicApi, WorkpieceId, WorkpieceSummary, WorkpiecesSubscriber,
 } from "@gadgets/workshop-shared/api";
 import type { CodeChange } from "@gadgets/workshop-shared/code-change";
 import {
@@ -54,6 +54,9 @@ export type AgentTurnResult = {
 /** Filters accepted by the public action-history API. */
 export type ActionListOptions = { beforeId?: number; filter?: ActionHistoryFilter };
 
+/** Gadget client and chat branch needed to connect to provisional agent code. */
+export type ProvisionalGadget = { client: RpcStub<GadgetClient>; chatId: number };
+
 /**
  * Drives a fresh local Workshop account and workspace through the public Cap'n Web API.
  * `runTurn()` observes one active-to-idle agent activation. It does not claim that late callbacks
@@ -66,6 +69,7 @@ export interface WorkshopAgentSession extends AsyncDisposable {
       ids: readonly [number, ...number[]], options?: AgentTurnOptions): Promise<AgentTurnResult>;
   listActions(options?: ActionListOptions): Promise<ActionHistoryPage>;
   connectedAccount(vendorId: string): ConnectedAccount;
+  openGadget(id: WorkpieceId): Promise<ProvisionalGadget>;
   close(): Promise<void>;
 }
 
@@ -444,6 +448,13 @@ class WorkshopAgentSessionImpl implements WorkshopAgentSession {
     return account;
   }
 
+  async openGadget(id: WorkpieceId): Promise<ProvisionalGadget> {
+    this.#assertNotClosed();
+    const chatId = this.#chatId;
+    if (chatId === undefined) throw new Error("The session has no chat branch");
+    return { client: await this.#workspace.getGadget(id), chatId };
+  }
+
   close(): Promise<void> {
     this.#closePromise ??= this.#close();
     return this.#closePromise;
@@ -695,8 +706,12 @@ class WorkshopAgentSessionImpl implements WorkshopAgentSession {
   }
 
   #assertOpen(): void {
-    if (this.#closed) throw new Error("WorkshopAgentSession is closed");
+    this.#assertNotClosed();
     if (this.#terminal) throw new Error("WorkshopAgentSession cannot continue after interruption");
+  }
+
+  #assertNotClosed(): void {
+    if (this.#closed) throw new Error("WorkshopAgentSession is closed");
   }
 }
 
--- a/packages/integration-tests/src/network-interceptor.ts
+++ b/packages/integration-tests/src/network-interceptor.ts
@@ -22,13 +22,28 @@ export type Handler = (
   request: Request,
 ) => Response | null | Promise<Response | null>;
 
+/** Decides whether one request may use the real network. */
+export type AllowRequest = (url: URL, method: string, headers: Headers) => boolean;
+
+type NetworkInterceptorOptions = {
+  handlers?: Handler[];
+  allow?: AllowRequest;
+  allowLoopback?: boolean;
+};
+
 export class NetworkInterceptor {
   readonly #handlers: readonly Handler[];
+  readonly #allow: AllowRequest | undefined;
+  readonly #allowLoopback: boolean;
   #realFetch: typeof globalThis.fetch | null = null;
   #unmockedCalls: string[] = [];
 
-  constructor(handlers: Handler[] = []) {
+  constructor({
+    handlers = [], allow, allowLoopback = true,
+  }: NetworkInterceptorOptions = {}) {
     this.#handlers = [...handlers];
+    this.#allow = allow;
+    this.#allowLoopback = allowLoopback;
   }
 
   install(): void {
@@ -44,8 +59,11 @@ export class NetworkInterceptor {
         : input.url;
       const url = new URL(raw);
 
-      // The harness dispatches its own traffic over loopback; let that through untouched.
-      if (url.hostname === "localhost" || url.hostname === "127.0.0.1" || url.hostname === "[::1]") {
+      // Test clients use loopback by default. Security-sensitive callers can route it through
+      // their handlers instead so model-authored requests cannot reach host services.
+      if (this.#allowLoopback &&
+          (url.hostname === "localhost" || url.hostname === "127.0.0.1" ||
+           url.hostname === "[::1]")) {
         return realFetch(input, init);
       }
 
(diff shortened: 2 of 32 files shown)


packages/workshop-evals/src/target.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function configureGateway(
+    config: WorkerConfig, access: Extract<LocalModelAccess, { kind: "gateway" }>): void {
+  config.vars = {
+    ...config.vars,
+    CF_AI_GATEWAY: access.gateway,
+    CF_AI_GATEWAY_ACCOUNT_ID: access.accountId,
+    CF_AI_GATEWAY_API_TOKEN: access.apiToken,
+    CF_AI_GATEWAY_PROVIDERS: "cloudflare",
+    CF_AI_GATEWAY_USE_BINDING: "false",
```
Maximo-Guk: Why don't we use the binding if we're only using workers ai models for now?

packages/integration-tests/src/network-interceptor.ts
```diff
@@ -22,13 +22,18 @@ export type Handler = (
...
+export type AllowRequest = (url: URL, method: string, headers: Headers) => boolean;
+
 export class NetworkInterceptor {
   readonly #handlers: readonly Handler[];
+  readonly #allow: AllowRequest | undefined;
   #realFetch: typeof globalThis.fetch | null = null;
   #unmockedCalls: string[] = [];
 
-  constructor(handlers: Handler[] = []) {
+  constructor(handlers: Handler[] = [], allow?: AllowRequest) {
```
Maximo-Guk: Could we use named parameters here ({handlers, allow}) to make it a little more obvious what we're passing in the parameters when we're creating NetworkInterceptor

packages/workshop-evals/vitest.eval.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "vitest/config";
+import { EVAL_TEST_TIMEOUT_MS } from "./src/config.js";
+
+export default defineConfig({
+  test: {
+    include: ["evals/**/*.eval.ts"],
+    globalSetup: ["../integration-tests/src/global-setup.ts"],
+    environment: "node",
+    fileParallelism: false,
```
Maximo-Guk: Any reason we're disabling fileParallelism on the evals?

packages/workshop-evals/src/config.ts
```diff
@@ -0,0 +1,50 @@
...
+  const commit = configured ?? readLocalCommit();
+  if (!GIT_SHA_PATTERN.test(commit)) {
+    throw new Error("WORKSHOP_EVAL_COMMIT must be a full 40-character Git SHA");
+  }
+  return commit;
+}
+
+/** Parse the model and repetition controls before a trial can spend inference. */
+export function evalMatrix(environment: NodeJS.ProcessEnv = process.env): EvalMatrix {
+  const models = commaList(environment.WORKSHOP_EVAL_MODELS ?? "");
```
Maximo-Guk: Hmm where do we set WORKSHOP_EVAL_MODELS right now? Should be that be a const set in vitest.config.ts?

packages/integration-tests/src/network-interceptor.ts
AshishKumar4: Agreed. `NetworkInterceptor` now takes `{ handlers, allow, allowLoopback }`, and all call sites use the named options.

packages/workshop-evals/vitest.eval.config.ts
AshishKumar4: No remaining reason. Each eval file gets its own Vitest fork, workerd harness, storage, ports, and process-local fetch interceptor. I removed `fileParallelism: false`.

packages/workshop-evals/src/config.ts
AshishKumar4: `WORKSHOP_EVAL_MODELS` is set by the developer's shell or the manual workflow in #362. `config.ts` owns and tests the fallback model because it also parses the model/trial matrix. Putting the value in Vitest config would split that policy between two places, so I kept the current single source of truth.

packages/workshop-evals/src/transcript.ts
```diff
@@ -0,0 +1,104 @@
...
+          : { ...toolCall, arguments: argumentsValue });
+
+        if (call.error !== undefined) {
+          events.push({
+            type: "tool_result",
+            toolCallId: call.toolCallId,
+            name: call.toolName,
+            error: { name: "Error", message: call.error },
+            metadata,
+          });
```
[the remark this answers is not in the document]
AshishKumar4: Fixed. Batched assistant tool calls are all emitted before their corresponding results, with a regression test.

packages/integration-tests/src/agent-session.ts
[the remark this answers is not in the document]
AshishKumar4: Fixed in the parent session driver. Cancelled operations now return the last canonical history, workpieces, and usage.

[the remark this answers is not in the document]
AshishKumar4: Fixed in the parent session driver. `deleteSelf()` always gets a fresh bounded attempt after stop and drain cleanup.

packages/workshop-evals/src/harness.ts
```diff
@@ -0,0 +1,178 @@
...
+      const success = runError === undefined && cleanupError === undefined &&
+        turns.length === task.turns.length &&
+        turns.every(turn => turn.outcome.status === "completed") &&
+        checks.length > 0 && checks.every(check => check.pass);
+      const usageMetadata: Record<string, JsonValue> = {};
+      if (usage.lastStepTokens !== undefined) usageMetadata.lastStepTokens = usage.lastStepTokens;
+      if (usage.observedCumulativeChatCostUsd !== undefined) {
+        usageMetadata.observedCumulativeChatCostUsd = usage.observedCumulativeChatCostUsd;
+      }
+      const events: TranscriptEvent[] = toTranscriptEvents(history);
```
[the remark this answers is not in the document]
AshishKumar4: This is intentional and bounded to developer-authored eval data. Each run creates a fresh local account and workspace; no production or user traffic is read. Reports contain the task prompt and generated fixture data so failures are diagnosable, while credentials and headers are never added to transcript events.

packages/workshop-evals/src/config.ts
Maximo-Guk: I feel like it would be fine to keep some defaults in the vitest.config.ts, so the developer doesn't need to set it manually, if they want to override the env or CI I think that's fine, but I think it makes sense to have a default

packages/workshop-evals/src/target.ts
Maximo-Guk: Why not provision a remote workers_ai binding? It should be fairly trivial, just adding the binding

Maximo-Guk: >. The remote option exists only on a full Wrangler project config

Are we sure? Is there an open-issue for this in workers-sdk?

packages/workshop-evals/src/config.ts
Maximo-Guk: Got it!

Maximo-Guk (approved):

packages/workshop-evals/src/target.ts
Maximo-Guk: Main reason it would be nice to get binding working is then you don't even need tokens in local dev to run the eval.

AshishKumar4: Have implemented it via workers ai binding IF CF_AI_GATEWAY_API_TOKEN isn't available, so we should now be able to use it in local or remote same cf account but also be able to test it against other non workers ai gateway models cross account (the way we have our AI gateway for prod right now)

packages/workshop-evals/src/target.ts
```diff
@@ -0,0 +1,166 @@
...
+function configureGateway(
+    config: WorkerConfig, access: Extract<LocalModelAccess, { kind: "gateway" }>): void {
+  config.vars = {
+    ...config.vars,
+    CF_AI_GATEWAY: access.gateway,
+    CF_AI_GATEWAY_ACCOUNT_ID: access.accountId,
+    CF_AI_GATEWAY_PROVIDERS: "cloudflare",
+    ...(access.apiToken === undefined ? {} : { CF_AI_GATEWAY_API_TOKEN: access.apiToken }),
+  };
+  if (access.apiToken === undefined) {
```
Maximo-Guk: To be honest this should probably also check for `CF_AI_GATEWAY_USE_BINDING=false / not existing` rather than just empty api token

AshishKumar4: tbh my agent actually proposed that but why would we want that though? feels redundant state?

Maximo-Guk: Main reason is because we use this elsewhere in the repo, and what if you have an API token already in env ( i.e CI org secrets which get injected in to vars ) but you want to opt out?

AshishKumar4: If an api token named CF_AI_GATEWAY_API_TOKEN is already in the env, that indicates we can use any arbitary gateway model, in which case it would be on us to configure what model to use - workers AI or gpt 5.6
But I would implement that env var regardless

Maximo-Guk: Right, but what if you want to opt-out and use the binding, you would have no way to do that if the secrets are injected at the org level this gives you a clean escape hatch

packages/workshop-evals/src/config.ts
```diff
@@ -0,0 +1,74 @@
...
+ * to Workers AI (every Workers AI ID is `@cf/...`, and cloudflare is the only provider the direct
+ * transport serves), so an unknown model runs exactly as a catalog Workers AI model does.
+ */
+export function resolveEvalModel(modelId: string): EvalModel {
+  for (const [provider, models] of Object.entries(SUGGESTED_MODELS)) {
+    if (Object.hasOwn(models, modelId)) {
+      return { provider: provider as AiModelProvider, model: modelId };
+    }
+  }
+  return { provider: "cloudflare", model: modelId };
```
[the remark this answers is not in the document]
Maximo-Guk: Fixing this

Maximo-Guk:
I added a small commit @AshishKumar4 which makes it so we inherit types from SUGGESTED_MODELS used by workshop-backend, so that way agents can quickly figure out what models they can test with. This is purely additive, the evals can also define their own models

Maximo-Guk (approved):

PR #367, opened 2026-08-27
enhanced gmail gatekeeper support

maxwellpeterson:
This PR adds a bunch of new features to the Gmail portion of the Google gatekeeper. The existing gatekeeper connection scopes (whole account, search query, specific label) are unchanged.

For human reviewers, I would focus on the gatekeeper interface changes in `packages/gatekeeper-google/src/types.d.ts` and let your agents handle the rest.

### Features added by this PR

- Richer lookup/search: `searchThreads()`, `listMessages()`, `searchMessages()`, `getMessage()`, and `getThread()`
- Message-level mutations, plus `star()`, `unstar()`, `applyLabel()`, and `removeLabel()`
- Message headers and attachment APIs, including attachment metadata and binary content reads
- Expanded sending and replies with CC, BCC, HTML, and explicit recipient overrides
- Gmail-style inline forwarding, including original body and attachments. Previously forwarding attached the original message as an `.eml`
- Complete draft lifecycle: list, get, create, update, delete, and send drafts, including reply/forward draft creation
- Label management: list, create, rename, delete, apply, and remove labels


packages/gatekeeper-google/src/types.d.ts
```diff
@@ -4,122 +4,487 @@ export type { Cursor };
...
+
+  /**
+   * Reopen a draft by a stable {@link GmailDraftInfo.id} previously returned
+   * by this binding. Pending updates are reflected by the returned capability.
+   */
+  getDraft(id: string): Promise<GmailDraft>;
+
+  /**
+   * Create a draft in the connected mailbox and return a capability for it.
+   * Only available for a whole-mailbox binding.
```
ndisidore, line 306: To me this says maybe we can break out the session types?
i.e.

```ts
export interface GmailScopedSession { /* list/search/get/getDraft/listDrafts */ }
export interface GmailMailboxSession extends GmailScopedSession {
  send(...); createDraft(...); listLabels(); createLabel(...); renameLabel(...); deleteLabel(...);
}
```
will keep agent context down too, as that way  an agent on a label-scoped binding doesn't need to read documentation for six methods that throw

packages/gatekeeper-google/src/types.d.ts
```diff
@@ -4,122 +4,487 @@ export type { Cursor };
...
+/** An attachment entry containing metadata and a content capability. */
+export type GmailAttachmentEntry = {
+  /** Metadata for this MIME part. */
+  info: GmailAttachmentInfo;
+  /** Capability for reading this MIME part's content. */
+  attachment: GmailAttachment;
+}
+
+/** Access to the mailbox, search, or label selected for this Gmail binding. */
 export interface GmailSession {
```
ndisidore: fwiw I think it could be good to have a method for getting the email of the attached mailbox
maybe something like

```ts
getMailboxAddress(): Promise<EmailAddress>
```

I think the agent will struggle to differentiate between an inbound and outbound email in a thread as it is and to e.g. not include itself in a recipient list, etc.

packages/gatekeeper-google/src/types.d.ts
```diff
@@ -4,122 +4,487 @@ export type { Cursor };
...
+  /**
+   * Reply to the sender. `body` is plain text and may be empty; `options.html`,
+   * when provided, is its HTML alternative. Supplying any recipient option replaces
+   * the complete recipient set calculated from the source message. For received mail,
+   * reply uses Reply-To when present and otherwise From. For mail sent by the
+   * connected mailbox, it uses the first original To recipient. Calculated
+   * lists remove the connected mailbox, omit original BCC recipients, and
+   * remove duplicates.
+   */
+  reply(body: string, options?: GmailReplyOptions): Promise<void>;
```
ndisidore: send/reply/replyAll/forward ... all these methods return void.
it seems like this is to workaround some simulation quirks?
since you already mint `newGmailMessageId()` you may want to at the least return a message-id here, which the agent could then use if it wanted to get the GmailMessage capability

packages/gatekeeper-google/src/types.d.ts
```diff
@@ -4,122 +4,487 @@ export type { Cursor };
...
   name?: string;
 }
 
+/**
+ * One recipient mailbox, either a bare address or a display-name form such as
+ * `Person <person@example.com>`. Line breaks and recipient groups are rejected.
+ */
+export type EmailRecipient = string;
+
+/** Metadata describing a Gmail thread. */
```
ndisidore, line 27: Pre-existing so nb, but this seems awfully thin for a list method.
No timestamp, no participants, no unread flag, no labels... not super useful and is going to require the agent to do a lot of sequential lookups

ndisidore (commented):
First pass focusing mostly on the api. But looking really good here! There's a lot to these google things

packages/gatekeeper-google/src/gmail.ts
```diff
@@ -0,0 +1,3753 @@
...
+function gmailAutoApprovalMetadata(
+    action: GmailAction,
+): Partial<Pick<ActionDescription, "actionKind" | "autoApprovable">> {
+  switch (action.type) {
+  case "send":
+  case "draftSend":
+    return {};
+  case "messageMutation":
+    return {
+      actionKind: {tag: action.operation, label: MESSAGE_MUTATION_LABELS[action.operation]},
```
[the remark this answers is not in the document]
vultuk: Fixed in 1aed85de. The bash correctness scenario now forces a dispatching preserved-close intent through retry and verifies dead_letter, retry_count 1, and a null execution ticket; the scenario passes.

packages/gatekeeper-google/src/types.d.ts
maxwellpeterson: Split out a separate `GmailScopedSession` interface in 7509b65a604f5d4f9d01e6ede6534034b964b986

packages/gatekeeper-google/src/types.d.ts
maxwellpeterson: Added in 7509b65a604f5d4f9d01e6ede6534034b964b986

packages/gatekeeper-google/src/types.d.ts
maxwellpeterson: Add all of your suggested properties in 7509b65a604f5d4f9d01e6ede6534034b964b986

packages/gatekeeper-google/src/gmail.ts
```diff
@@ -0,0 +1,3753 @@
...
+      autoApprovable: true,
```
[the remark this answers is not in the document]
maxwellpeterson: This is intentional -- simulating message mutations is quite complicated and this PR is already massive. I'll consider addressing in a future PR.

packages/gatekeeper-google/src/types.d.ts
maxwellpeterson: Yep, added in 6003637c6d65c6bc6eaec79c7f70bb74b6d5f16b

ndisidore (commented):
💥  this is beautiful. API looks great.
Happy to approve once bonk is satiated (which is a chore 😅 )

ndisidore (approved):

PR #368, opened 2026-08-27
Run Workshop integration files in parallel

AshishKumar4:
The Workshop integration suite grew from two files to six. Five files start a local Workshop, and `fileParallelism: false` made those startup costs add up. The suite took 28–43 seconds on current main and would reach about 50 seconds with the agent-test stack.

The old config comment said parallel files would share ports and storage. They do not. Vitest runs each file in an isolated fork, and every `createTestHarness()` call starts separate workerd processes with dynamic ports and in-memory Durable Object, KV, and R2 state.

The real shared resource is generated code. `capnweb-validate` deletes and rewrites `packages/workshop-backend/.wrangler/validate`, while Wrangler reads its generated `server.ts`. This PR gives the format, browser-runtime, and validated-Worker builds one Vite+ task owner. The integration task builds that entrypoint once, then isolated file processes read it. Direct runs use the same task. Watch mode registers the complete Worker input closure, including a configured external blueprint directory, and rebuilds before each rerun. Generated outputs are excluded so a build cannot trigger itself.

The build steps keep the repository's 60-second idle and 600-second total watchdog. The integration task is uncached because backend source reaches it through the gitignored validated output; running this fast suite is safer than maintaining a second source fingerprint.

Measured locally:

```text
Current main, sequential              28–43s
Current main, default parallel      2.67–3.32s
Expanded stack, default parallel    2.89–3.64s
```

Five repeated runs of both the current and expanded suites passed with Vitest’s default worker count. An explicit cap made both suites slower, so `maxWorkers` remains unset.

Run it with:

```sh
pnpm --filter @gadgets/integration-tests test:run
```

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/368" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

Wall time: 0.54 seconds

Wall time: 0.47 seconds

### DIFF
--- a/packages/integration-tests/__tests__/workshop-sharing.test.ts
+++ b/packages/integration-tests/__tests__/workshop-sharing.test.ts
@@ -6,7 +6,7 @@ import {
 import { type Harness, startHarness } from "../src/harness.js";
 import { mockChatCompletion } from "../src/mock-model.js";
 import { NetworkInterceptor } from "../src/network-interceptor.js";
-import { connect, nextUsernames, signUp } from "../src/rpc-client.js";
+import { connect, logIn, nextUsernames, signUp, waitFor } from "../src/rpc-client.js";
 
 let harness: Harness | undefined;
 const network = new NetworkInterceptor([mockChatCompletion("Test chat")]);
@@ -48,10 +48,52 @@ async function expectOpenDenied(
   expect(getOpenGadgetErrorCode(denied)).toBe(OPEN_GADGET_ERROR_CODES.workspaceAccessDenied);
 }
 
+async function withAuthenticated<T>(
+    username: string, body: (authenticated: RpcStub<AuthenticatedApi>) => Promise<T>): Promise<T> {
+  using publicApi = connect(requireHarness().url);
+  using authenticated = await logIn(publicApi, username);
+  const result = await body(authenticated);
+  return result;
+}
+
 async function activate(workspace: RpcStub<Overseer>): Promise<string> {
   const { id } = await workspace.getMetadata();
   await workspace.newChat("Make this workspace visible without an agent", null);
   return id;
+}
+
+async function expectOpenDeniedAfterRestart(
+    username: string, workspaceId: string, shareKey?: string): Promise<void> {
+  const result = await waitFor("revoked workspace access after restart", async () => {
+    try {
+      return await withAuthenticated(username, async authenticated => {
+        try {
+          using _workspace = await authenticated.openGadget(workspaceId, shareKey);
+        } catch (error) {
+          return getOpenGadgetErrorCode(error) === OPEN_GADGET_ERROR_CODES.workspaceAccessDenied
+            ? { denied: true }
+            : null;
+        }
+        return { denied: false };
+      });
+    } catch {
+      return null;
+    }
+  });
+  expect(result.denied).toBe(true);
+}
+
+async function listShareLinksAfterRestart(username: string, workspaceId: string) {
+  return waitFor("owner workspace after revocation restart", async () => {
+    try {
+      return await withAuthenticated(username, async authenticated => {
+        using workspace = await authenticated.openGadget(workspaceId);
+        return workspace.listShareLinks();
+      });
+    } catch {
+      return null;
+    }
+  });
 }
 
 it.concurrent("grants and revokes a use-only collaborator", async () => {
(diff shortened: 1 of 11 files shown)


packages/workshop-backend/vite.config.ts
```diff
@@ -22,28 +22,27 @@ export default {
...
-       * would run it in vp's clean environment. `cache: false` because `build-browser-runtime.mjs`
-       * writes into the same package tracking hashes as its input, which vp declines to cache.
-       */
+      'build:browser-runtime': {
+        command: 'node ../../scripts/with-timeout.ts --idle 60 --max 600 -- node build-browser-runtime.mjs',
+        cache: false,
+      },
+      /** Builds the validated entrypoint shared by integration-test file workers. */
+      'build:integration-worker': {
+        command: 'node ../../scripts/with-timeout.ts --idle 60 --max 600 -- pnpm exec capnweb-validate build --out .wrangler/validate',
```
Maximo-Guk: Why aren't we using `vitest-task-vite-config.ts` / hardcoding these values here ?

packages/integration-tests/src/global-setup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+function runNode(args: string[]): void {
+  execFileSync(process.execPath, [
+    WATCHDOG, "--idle", "60", "--max", "600", "--", process.execPath, ...args,
+  ], { cwd: WORKSHOP_DIR, stdio: "inherit" });
+}
+
+function buildWorkshop(): void {
+  runNode(["scripts/build-format-blueprints.mjs"]);
+  runNode(["build-browser-runtime.mjs"]);
```
Maximo-Guk: Why are we building here, and in the vp+ task config

packages/integration-tests/vitest.config.ts
```diff
@@ -3,13 +3,12 @@ import { defineConfig } from "vitest/config";
...
     hookTimeout: 120_000,
-    // One harness at a time: the workers bind fixed DO namespaces and KV, so files running together
-    // would race on ports and persisted state. This serialises the files -- one runs to completion
-    // before the next starts, though they may share a worker process -- and leaves what happens
-    // inside a file alone: only the cases written as `it.concurrent` run together.
-    fileParallelism: false,
+    // Vitest isolates files in fork processes. Two workers overlap workerd startup without letting
+    // the full repository test run create one workerd fleet per CPU.
+    fileParallelism: true,
+    maxWorkers: 2,
```
Maximo-Guk: What is this default? Why not set it to default, and could we set this to a higher worker limit?

packages/integration-tests/src/harness.ts
```diff
@@ -83,6 +83,9 @@ function workshopConfig(
     gatekeepers: { binding: string; name: string }[],
     patch?: (config: WorkerConfig) => void): WorkerConfig {
   const config = readWorkerConfig(WORKSHOP_DIR);
+  // globalSetup completed the destructive shared `.wrangler/validate` build before file workers
+  // started. Rebuilding it in each fork would race on that directory.
+  if (process.env.WORKSHOP_INTEGRATION_PREBUILT === "1") delete config.build;
```
Maximo-Guk, line 88: Why do we delete config.build? How does that prevent rebuilding?

packages/integration-tests/__tests__/workshop-sharing.test.ts
```diff
@@ -83,15 +90,23 @@ it.concurrent("grants and revokes a use-only collaborator", async () => {
     role: "use",
   });
   await expect(collaboratorWorkspace.setTitle("Forbidden rename")).rejects.toThrow();
-
   const affected = await workspace.removeCollaborator(added.profile.id, []);
+  collaboratorWorkspace[Symbol.dispose]();
```
Maximo-Guk: Maybe we could use using here?

packages/integration-tests/__tests__/workshop-sharing.test.ts
```diff
@@ -54,79 +55,101 @@ async function activate(workspace: RpcStub<Overseer>): Promise<string> {
   return id;
 }
 
+// Revocation aborts the workspace DO after its RPC response. Close linked sessions first, then
+// reconnect after the scheduled restart instead of treating that expected disconnect as a failure.
+function waitForRevocationRestart(): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, REVOCATION_RESTART_SETTLE_MS));
```
Maximo-Guk: Is there some way we can do this without a setTimeout wait? Feels a little hacky not being able to await a teardown and having a hardcoded time here

packages/integration-tests/vitest.config.ts
Maximo-Guk: Did we find this out?

packages/integration-tests/package.json
```diff
@@ -10,8 +10,9 @@
   },
   "scripts": {
     "build": "tsc --noEmit",
-    "test:run": "vitest run",
-    "test:watch": "vitest"
+    "test:prebuild": "vp run -F @gadgets/workshop-backend build:integration-worker",
```
Maximo-Guk, line 13: Why isn't this a vp+ task?

packages/integration-tests/package.json
```diff
@@ -10,8 +10,9 @@
...
+    "test:run": "pnpm run test:prebuild && vitest run",
```
Maximo-Guk, line 14: Should these also be vp+ tasks similar to all the other test packages?

packages/integration-tests/vitest.config.ts
AshishKumar4: It's because the maxWorkers defaults to:                                                                                                                                                                                                                                  
   - Test run: available CPU count - 1                                                                                                                                                                                                                      
   - Watch mode: half the available CPUs                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                            
 That default is too high here. Each test file starts about two workerd processes, so it it can lead to deminishing returns with high cores as overheads mount.                                                        

But probably can increase to 3

 Measured:                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                            
 ```text                                                                                                                                                                                                                                                    
2 workers: 6.5–7.2s
3 workers: 5.4–6.2s
4 workers: 5.4–6.2s
 ```

packages/integration-tests/package.json
AshishKumar4: build:integration-worker is already the Vite+ task, owned by workshop-backend because that package owns .wrangler/validate. Adding another Vite+ task in integration-tests would give the same build two task names. test:prebuild only composes that task into the direct and watch commands.

packages/integration-tests/package.json
AshishKumar4: `packages/integration-tests/vite.config.ts` already defines the Vite+ `test` task and makes it depend on the backend build task. `test:run` remains the direct, uncached Vitest path, matching the repository convention; `vp run -F @gadgets/integration-tests test` is the Vite+ path. Both use the same backend prebuild task, so the build command still has one owner.

packages/integration-tests/vitest.config.ts
AshishKumar4: Setting to default

Maximo-Guk (approved):

packages/integration-tests/src/global-setup.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/** Share one validated Worker build across isolated test-file processes. */
+export default function setup(project: TestProject): () => void {
+  if (!existsSync(VALIDATED_ENTRY)) {
+    throw new Error("The integration-test Workshop build did not produce its validated entrypoint");
+  }
+  process.env.WORKSHOP_INTEGRATION_PREBUILT = "1";
+  // Vite+ owns the initial build. A watch process stays alive, so later reruns invoke that task here.
+  project.onTestsRerun(rebuildWorkshopForWatch);
```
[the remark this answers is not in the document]
AshishKumar4: Agreed. Added `forceRerunTriggers` for the Workshop source, browser runtime, format blueprints, Wrangler config, shared API, backend utilities, and integration fixtures. I verified this by changing `workshop-backend/src/server.ts` during `test:watch`: Vitest invoked the shared `build:integration-worker` task and reran all six files.

packages/integration-tests/vitest.config.ts
```diff
@@ -3,13 +3,16 @@ import { defineConfig } from "vitest/config";
 export default defineConfig({
   test: {
     include: ["__tests__/**/*.test.ts"],
-    // Each harness boots real workerd instances, so these run far slower than unit tests.
+    globalSetup: ["./src/global-setup.ts"],
+    forceRerunTriggers: [
+      "../workshop-backend/{src,browser,format-blueprints}/**",
```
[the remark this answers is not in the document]
AshishKumar4: Agreed. The trigger no longer watches `src/generated/**`. It watches root hand-authored `src/*.{ts,txt}` files plus the `auth` and `ai-gateway-billing` source trees, so browser and blueprint outputs cannot schedule a second destructive rebuild.

packages/integration-tests/vitest.config.ts
```diff
@@ -1,15 +1,33 @@
...
-    // Each harness boots real workerd instances, so these run far slower than unit tests.
+    globalSetup: ["./src/global-setup.ts"],
+    // Wrangler loads these outside Vitest's module graph. Absolute source paths make watch mode
+    // rebuild before rerunning; generated outputs are omitted so a build cannot trigger itself.
+    forceRerunTriggers: [
+      workspacePath("packages/workshop-backend/src/*.{ts,txt}"),
+      workspacePath("packages/workshop-backend/src/auth/**"),
+      workspacePath("packages/workshop-backend/src/ai-gateway-billing/**"),
+      workspacePath("packages/workshop-backend/browser/**"),
+      workspacePath("packages/workshop-backend/format-blueprints/**"),
```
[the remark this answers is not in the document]
AshishKumar4: Agreed. The configured directory is now resolved exactly as the generator resolves it: relative to `workshop-backend`, or directly when absolute. That literal directory is registered with Vite’s watcher and its absolute `/**` pattern forces the rerun. I verified this with an external `/tmp/cfos-watch-blueprints` directory; changing its sidecar rebuilt the Worker and reran all six files.

packages/integration-tests/vitest.config.ts
```diff
@@ -1,15 +1,39 @@
...
+).replaceAll("\\", "/");
+
 export default defineConfig({
   test: {
     include: ["__tests__/**/*.test.ts"],
-    // Each harness boots real workerd instances, so these run far slower than unit tests.
+    globalSetup: ["./src/global-setup.ts"],
+    // Wrangler loads these outside Vitest's module graph. Absolute source paths make watch mode
+    // rebuild before rerunning; generated outputs are omitted so a build cannot trigger itself.
+    forceRerunTriggers: [
```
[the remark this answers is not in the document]
AshishKumar4: Agreed. The config now registers literal source directories and files through a small Vite plugin, then uses absolute `forceRerunTriggers` only to filter their events. Hand-authored backend source entries are enumerated while `src/generated` is excluded, so generated output cannot trigger the build that wrote it.

packages/integration-tests/vitest.config.ts
```diff
@@ -1,15 +1,71 @@
...
+const WORKSPACE_DIR = resolve(dirname(fileURLToPath(import.meta.url)), "../..");
+const workspacePath = (path: string) =>
+  resolve(WORKSPACE_DIR, path).replaceAll("\\", "/");
+const formatBlueprintsPath = resolve(
+    WORKSPACE_DIR,
+    "packages/workshop-backend",
+    process.env.FORMAT_BLUEPRINTS_DIR ?? "format-blueprints",
+).replaceAll("\\", "/");
+const backendSourceRoot = workspacePath("packages/workshop-backend/src");
+const backendSourceInputs = readdirSync(backendSourceRoot, { withFileTypes: true })
```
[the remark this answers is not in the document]
AshishKumar4: Agreed. The watcher now registers `backendSourceRoot` itself. The rerun filters match root entries and nested entries outside `generated`, so source added after Vitest starts is covered without watching generated output. I verified this by creating a new top-level `src/watch-trigger-smoke/new-source.ts` while watch mode was running; it rebuilt the Worker and reran all six files.

PR #369, opened 2026-08-27
Keep tool picker position when selecting tools

dcartertwo:
Selecting a tool rebuilt the configurator form and reset the nested tool list to the top. Preserve each checkbox list's position across selection-driven rerenders; changing the filter still starts the filtered results at the top.

The regression test executes the generated configurator runtime and exercises both interactions against real DOM elements.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/369" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/scripts/build-gatekeeper-configurator.ts
+++ b/scripts/build-gatekeeper-configurator.ts
@@ -208,7 +208,7 @@ function setValues(patch) {
   const active = getFocusState();
   values = { ...values, ...patch };
   postSelectionState();
-  render(active);
+  render(active, true);
 }
 
 function clearFields(...names) {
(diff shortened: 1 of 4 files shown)


scripts/build-gatekeeper-configurator.ts
```diff
@@ -648,16 +648,28 @@ RadioCards = components.RadioCards;
 CheckboxList = components.CheckboxList;
 Autocomplete = components.Autocomplete;
 
-function render(focusState = undefined) {
+function render(focusState = undefined, preserveCheckboxScroll = false) {
   if (!root || !spec) return;
   if (focusState === undefined) focusState = getFocusState();
+  const checkboxScrollOffsets = preserveCheckboxScroll
```
ndisidore, line 654: Hmmm, this seems like a hack - by default selecting a checkbox should't scroll, so we're causing the scroll and then working around that by scolling back.

This probably means we're re-rendering the container on every select which is iffy state mgmt.

I had an agent poke and it confirms this
```ts
CheckboxList.onchange → emit() → configurator onChange → setValues() → render() → root.replaceChildren(...)
```

however I didn't look into how invasive reworking this would be or if there are tradeoffs with this approach so consider this non-blocking because this does work as-is

ndisidore (approved):

scripts/build-gatekeeper-configurator.ts
dcartertwo: agree it is a little hacky

PR #373, opened 2026-08-27
Simplify repository build scripts & make cross platform

ndisidore:
This removes run-local's repository-wide hash and stamp, and uses Vite+'s existing task cache for typed-storage and frontend assets instead. The frontend bundle tasks now share cache metadata, enforce production mode from Vite config, and clean dist without relying on POSIX shell syntax. Release and preview scripts now share one deployable-package reader and gatekeeper naming helpers, while unused app-watch aliases and the test-only oxlint rule export are gone.

### DIFF
--- a/packages/workshop-frontend/vite.config.ts
+++ b/packages/workshop-frontend/vite.config.ts
@@ -5,16 +5,35 @@ import tsconfigPaths from 'vite-tsconfig-paths'
 import { TanStackRouterVite } from '@tanstack/router-plugin/vite'
 import { vitestTask } from '../../scripts/vitest-task-vite-config.js'
 
-// `dist/` is this package's own build output, excluded from the inputs of both tasks: vp declines to
-// cache a task that reads a path it also wrote. Package-relative rather than workspace-wide --
-// `typed-storage` emits, and its `exports` resolves to `dist/index.js`, so a pattern matching every
-// package's `dist` would drop a real input.
+// `dist/` is this package's own build output, excluded from the inputs of the bundle and test
+// tasks: vp declines to cache a task that reads a path it also wrote. Package-relative rather than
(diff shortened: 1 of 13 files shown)


scripts/run-local.ts
```diff
@@ -41,109 +30,16 @@ try {
...
-  // Build only what's required to run locally (no full-repo type-check / no frontend tsc).
-  runPnpm(["--filter", "@gadgets/typed-storage", "build"]);
-  runPnpm(["--filter", "@gadgets/workshop-frontend", "exec", "vite", "build"]);
-
-  // Record the stamp only after a successful build so an interrupted build retries next time.
-  writeFileSync(STAMP_PATH, sourceHash + "\n");
-} else {
-  console.log("No source changes since last build; skipping build.");
-}
+runPnpm(["install"]);
```
ndisidore, line 40: `pnpm install` already caches. this is a cheap call and caching the cache is not worth it imo

packages/workshop-frontend/vite.config.ts
```diff
@@ -27,34 +44,25 @@ const runConfig = {
...
+        ...frontendBundleTaskOptions,
       },
       test: vitestTask('vitest run', [ownDist]),
     },
   },
 }
 
-export default defineConfig(({ mode }) => {
+export default defineConfig(({ command, mode }) => {
+  if (command === 'build') process.env.NODE_ENV = 'production'
```
Maximo-Guk: GPT comment:

- P2 Set NODE_ENV before Vite starts packages/workshop-frontend/vite.config.ts:64  
Vite snapshots whether NODE_ENV was initially set before loading this callback. If an .env* file specifies NODE_ENV=development, Vite resets it after this assignment, producing development semantics. The prior command-scoped assignment prevented that. Use a cross-platform wrapper that sets NODE_ENV=production before importing Vite.

Maximo-Guk (approved):

PR #377, opened 2026-08-28
Fix stale chat metadata rebuild race

ndisidore:
We've seen  a few sentry errors in production where a chat change message can arrive before metadata containing its pin, causing the OT client to rebuild from an empty file map and throw `edit of absent file`.

Currently the `ChatOtClient` latches the exception as fatal. When the matching metadata arrives moments later, that client does not retry; the error remains until it is recreated.

This defers snapshots whose materialization watermark is ahead of metadata; the matching metadata update then recomputes and applies the change.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/377" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/packages/workshop-frontend/src/otClient.ts
+++ b/packages/workshop-frontend/src/otClient.ts
@@ -259,6 +259,8 @@ export class ChatOtClient {
   setDurableState(durable: ChatDurableCode): void {
     this.#enqueue(async () => {
       const codeBase = durable.codeBase ?? { pins: [], generation: 0, revision: 0 }
+      // Messages can arrive before the metadata that covers their materialization watermark.
+      if (durable.rowsThrough > codeBase.revision) return
       this.#latestDurable = durable
 
       if (!this.#ready) {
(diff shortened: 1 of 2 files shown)


packages/workshop-frontend/src/otClient.ts
```diff
@@ -259,6 +259,8 @@ export class ChatOtClient {
   setDurableState(durable: ChatDurableCode): void {
     this.#enqueue(async () => {
       const codeBase = durable.codeBase ?? { pins: [], generation: 0, revision: 0 }
+      // Messages can arrive before the metadata that covers their materialization watermark.
+      if (durable.rowsThrough > codeBase.revision) return
```
ndisidore, line 263: The early return is safe here because the ot client can't action revision without metadata 

 Concrete sequence:

 1. View shows revision 10.
 2. Revision-11 message arrives; guard ignores it because metadata still says revision 10.
 3. Metadata revision 11 arrives; snapshot retries and view updates.

 The server queues steps 2 and 3 back-to-back, so normally this is milliseconds.

There are other more involved fixes available here but this minimal guard should handle most

Maximo-Guk (approved):

PR #380, opened 2026-08-28
Observer verification: fix coverage bookkeeping, restart sessions when scope widens

Maximo-Guk:
Why:

Before this PR: We didn't sever existing observers, this was just actually documented in the original plan "existing observers see an incremental modal for just the new binding on their next open..."

The main gap we have right now is adding a binding does not restart live sessions, so an already-open collaborator is only verified against it at their next open. 

What:

authorizeCollaborator() is now responsible for being the single gate for checking whether observers have access to underyling data. Everything routes through here, including `receiveExternalMessage()` which previously only checked the role and never checked whether the observer had access to the underlying data.

We also decided to abort the DO in order to revoke the RPC capabilities, we already use the same mechanism when revoking collaborators and we can just the same mechanism here. We made this a little less disruptive by checking graph first to see if there's any other collaborators before aborting the DO

Testing:

I added extensive integration tests in `observer-role-scope.test.ts`, `observer-reverification.test.ts` and `external-message-verification` ( even though it's actually used yet in any gatekeepers ) for testing the observer verification logic


packages/workshop-backend/src/overseer.ts
```diff
@@ -4339,20 +4368,35 @@ class OverseerImpl implements AgentHooks {
...
+    // #restartIfShared severs the sessions that were never verified against it: the DO's input gate
+    // is open across the await, ids are allocated sequentially, so a live build session can guess
+    // this one, and getGatekeeperById (OverseerClientInterface) gates on nothing but existence.
+    let facet = this.getGatekeeperFacet(id, cls);
     try {
       let description = await facet.describe();
       gatekeeperRecord.resourceTitle = description.title;
       gatekeeperRecord.resourceUrl = description.url;
       gatekeeperRecord.hasSlashCommands = description.hasSlashCommands;
       this.storage.gatekeepers.put(gatekeeperRecord);
```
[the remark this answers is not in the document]
kentonv: Note that, obviously, any writes have to happen before the abort or they will be lost. Moreover, the grace period is important to make sure writes actually settled. An abort() breaks the output gate, causing existing in-flight writes never to land. At the very least, we have to `await ctx.storage.sync()`, but that also gives a chance for a client to jump in with a request.

Perhaps the easiest solution is to use `blockConcurrencyWhile()`.

```
await this.ctx.blockConcurrencyWhile(async () => {
  await this.ctx.storage.sync();

  // Throwing from blockConcurrencyWhile also resets the DO.
  throw new Error("Gadget restarted because a new connection was added.");
});
```

Another option would be to store in-memory that the given gatekeeper should be blocked from being accessed -- the mark naturally goes away after restarting. But, this seems more complicated, and blockConcurrencyWhile() should do the trick I think.

packages/workshop-backend/src/overseer.ts
```diff
@@ -4975,30 +5019,61 @@ class OverseerImpl implements AgentHooks {
...
+  // Fire-and-forget: the callers are synchronous (bindWorkpiece) or already past their last write,
+  // and getSharingManager() is async, so failures are logged rather than left as an unhandled
+  // rejection. Failing to restart is fail-open for the widened scope, hence the `error` level.
+  //
+  // Note that ensureAmbientCapsules() calls addGatekeeper() from inside open(), so on a shared
+  // workspace the first open after an ambient capsule appears bounces itself once; the capsule
+  // exists by then, so the client's retry is clean.
+  #restartIfShared(reason: string): void {
+    this.getSharingManager().then(sharing => {
+      if (sharing.listCollaborators().length === 0) return;
```
kentonv: Instead of checking whether any collaborators exist on the sharing graph, we should check whether anyone is actually _connected_. We can keep a count of the number of live `OverseerClientInterface` and `UseOverseerInterface` objects (increment count in constructor, decrement in disposer) in order to decide whether anyone is connected. (We can even count connections by the owner separately from collaborators.)

packages/workshop-backend/src/overseer.ts
```diff
@@ -7861,14 +7978,31 @@ class OverseerImpl implements AgentHooks {
...
+    //    Rebinding a connection keeps its gatekeeper id, so a stale entry from before an unbind
+    //    would otherwise re-register them off a choice made for a scope the workspace no longer
+    //    has, instead of asking them again.
     let record = this.storage.observers.get(profileId);
+    if (record) {
+      let inScopeIds = new Set(inScope.map(gk => gk.id));
+      let pruned = false;
+      for (let key of Object.keys(record.accountChoices)) {
+        if (!inScopeIds.has(Number(key))) {
+          delete record.accountChoices[Number(key)];
```
kentonv: This removes the account choice from the observer record, but does not actually remove the observer from the gatekeeper. They are still registered there and as a result will still be surfaced in `excludeObservers`, and since they still have the "use" role this will still cause said observations to be blocked.

I actually don't think I agree with the premise of the comment above explaining this new section. If I understand the Claudish correctly, it's saying:

* Gatekeeper 1 is bound to a gadget.
* New observer arrives, chooses an account for gatekeeper 1 and verifies.
* Gatekeeper 1 is unbound from the gadget.
* Gatekeeper 1 is later rebound against a different gadget.
* Should the observer be expected to choose an account again or do they keep the choice they made earlier?

IMO it's fine to keep the choice they made earlier. We should just leave the `accountChoices` entry in there.

If, however, the gatekeeper flags the user in `excludedObservers` for some observation, we should _at that time_ determine if the user no longer has any visibility to the gatekeeper (i.e. the gatekeeper is no longer bound to any gadgets and the user has use-only access). In that case, we should remove the user as an observer on the gatekeeper, and remove the `accountChoices` record so that they have the option to choose a different account next time...

(But I may be missing something, let me know if you or your agent sees a problem with this...)

packages/workshop-backend/src/overseer.ts
```diff
@@ -7970,25 +7973,41 @@ class OverseerImpl implements AgentHooks {
 
           let fail = (reason: string, err?: unknown) => {
             failures.set(gk.id, {accountId, reason});
+            // The persisted record is what asserts this collaborator was verified for this
```
kentonv: I think this is incorrect. The existence of an entry in `accountChoices` doesn't assert anything about the user's verification status. It simply records the user's choice, so that we don't have to ask them again.

A secondary, though less important, role of `accountChoices` is to record which gatekeepers may have already had the user registered via a successful `addObserver()`. Those gatekeepers potentially need later `removeObserver()` calls to clean up. However, this isn't strictly necessary -- if we accidentally leave a user registered on `removeObserver()` when they no longer have access, then when they show up in `excludeObservers` in the future, we can detect that the observer is no longer present and remove them at that time. So existence in `accountChoices` is more of a hint for this purpose.

But regardless, nothing should be assuming that an entry in `accountChoices` means the user is actually verified.

Under this understanding, I believe this commit can be removed, since there's not necessarily any need to delete a choice just because it failed re-verification. If we do want to delete it, though, we would also need to call `removeObserver()` on the gatekeeper to fully remove this observer. But if we do that, we *also* need to make sure that the observer doesn't have any other live connections open, because if they do, those connections would then be able to bypass `excludeObservers`.

So I think we should actually just remove this commit.

packages/workshop-backend/src/overseer.ts
kentonv: Similar to the other comment -- this seems to misunderstand `accountChoices`.

I think we can leave the record alone. If the gatekeeper ever becomes in-scope again, then the user will probably want to keep the same choice they made earlier.

Note, though, that in the meantime, the user is still registered as an observer on the gatekeeper (and even with the change written here, they still would be -- there's no call to `gatekeeper.removeObserver()` here). This means they may show up in `excludeObservers` spuriously. We should have the `excludeObservers` handling detect that the gatekeeper is no longer in-scope for the user, and then remove the observer from the gatekeeper, even though the observer is still in the workspace's observers table. Currently (unless it changed elsewhere in this PR that I haven't seen yet), when the user shows up in `excludeObservers` it will actually cause the observation to be blocked, which isn't right if the gatekeeper is no longer in-scope for the user...

kentonv (commented):
I started reviewing this but I see you're making other changes so decided to post what I have and come back later.

kentonv (commented):
> Bugfix: Don't let a delayed abort swallow a concurrent access change.

This commit is suspicious. It seems to be complaining that a randomly-placed abort() could leave the DO in an inconsistent state, if _something else_ is happening in the meantime that gets interrupted in the middle.

But if that's the case, then that _something else_ is the buggy bit. Every piece of code should be tolerant of a random crash interrupting it. Usually, DO's atomicity guarantees take care of it: any writes performed without an intervening `await` are automatically coalesced into a transaction. But anything that does writes around an async operation needs to consider what happens if the async operation never returns (due to a crash).

Of course, adding and removing observers from a gatekeeper is inherently async, and we can't even wrap it in a transaction since the gatekeeper itself is modifying its own storage. (I do want to create cross-facet transactions at some point... but we don't have them yet.)

The rule to follow here is: Ensure the gatekeeper _overestimates_. That is:
* addObserver() must succeed before the observer is granted any access.
* remoteObserver() must not be called until the observer's access has been revoked.

Since it's an over-estimate, in the event of a badly-timed crash, a gatekeeper can end up having an observer registered who doesn't actually have access. That is OK: it only causes that observer to possibly show up in `excludeObservers` in the future. At that point, the overseer checks and discovers that the excluded observer doesn't have access in the first place, and can then remove them from the gatekeeper, thus bringing it into sync.

In any case, though, I don't quite understand the situation that the first paragraph of the commit description is worried about. If `removeCollaborator()` has already removed the sharing edge (and the observer has also been removed from the sharing table), then it doesn't really matter if we fail to finish removing the observer from all gatekeepers -- the gatekeepers can be left over-estimating.

packages/workshop-backend/src/overseer.ts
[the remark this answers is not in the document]
Maximo-Guk: I ended up adopting the `blockConcurrencyWhile()` approach so now writes are flushed before the reset, but I wanted to avoid putting the `scheduler.wait(100)` inside of the blockConcurrencyWhile, so I used some in-memory markers (`#gatekeepersPendingRestart`) to block the new connection during the ~100ms pre-reset window (since the sessions it's about to sever stay live until then).
 
https://github.com/cloudflare/cloudflare-os/pull/380/changes#diff-b21fc87bbe11ba6ce2f1fef7efe5dd1df886282ec94c5450a2cf8c0acf084fccR5180

To be honest I'd argue that if someone severs a connection, I think it would be reasonable to expect a delay up to a few seconds for the access to get revoked but it seemed trivial enough to add `#gatekeepersPendingRestart` so I don't feel strongly

packages/workshop-backend/src/overseer.ts
```diff
@@ -4975,30 +5019,61 @@ class OverseerImpl implements AgentHooks {
...
+  // and getSharingManager() is async, so failures are logged rather than left as an unhandled
+  // rejection. Failing to restart is fail-open for the widened scope, hence the `error` level.
+  //
+  // Note that ensureAmbientCapsules() calls addGatekeeper() from inside open(), so on a shared
+  // workspace the first open after an ambient capsule appears bounces itself once; the capsule
+  // exists by then, so the client's retry is clean.
+  #restartIfShared(reason: string): void {
+    this.getSharingManager().then(sharing => {
+      if (sharing.listCollaborators().length === 0) return;
+      return this.scheduleAccessRestart(reason);
```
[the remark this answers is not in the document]
Maximo-Guk: See above comment https://github.com/cloudflare/cloudflare-os/pull/380#discussion_r3904124228, final sync is now in blockConcurrencyWhile()

packages/integration-tests/__tests__/external-message-verification.test.ts
```diff
@@ -0,0 +1,205 @@
...
+          .resolves.toMatchObject({
+            accepted: false, message: expect.stringMatching(/AI model/i) });
+
+      // The gatekeeper now revokes Bob's underlying access. His persisted observer record is
+      // untouched, so only a live addObserver re-verification on this submission can notice --
+      // and the gatekeeper's own refusal reason appearing in the reply is the proof that round
+      // trip happened, since nothing persisted in the Workshop contains it. An implementation
+      // that merely checked the record would keep accepting him here.
+      await setVerifyOutcome(accountLabel(bobAccount), { allow: false, reason: DENIED_REASON });
+      const revoked = await submitExternalMessage({ callerEmail: bob, gadgetKey, prompt: "hi" });
```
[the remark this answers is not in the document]
Maximo-Guk: This got dropped after https://github.com/cloudflare/cloudflare-os/pull/380#discussion_r3896728247, so doesn't apply anymore

packages/workshop-backend/src/overseer.ts
```diff
@@ -8023,9 +8205,29 @@ class OverseerImpl implements AgentHooks {
...
+      // would have excluded them from is admitted with nothing left to block it.
+      let rollback = returningObserver ? newlyAdded : new Set([...newlyAdded, ...invalidated]);
+      await this.#removeObserverFromGatekeepers(observerId, [...rollback]);
+
+      // This open is being denied, but the collaborator may hold other sessions that opened while
+      // the scrubbed choice still verified them. Sever every session so they must re-verify --
+      // whoever can't will simply be denied their next open. This cannot loop: a second identical
+      // failure finds the entry already scrubbed, so no flag and no restart. A re-prompt that
+      // repairs the failure never reaches here, and step 6 re-persists full coverage.
+      if (scrubbedCoverage) {
```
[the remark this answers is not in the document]
Maximo-Guk: This got dropped after https://github.com/cloudflare/cloudflare-os/pull/380#discussion_r3896728247, so doesn't apply anymore

packages/workshop-backend/src/overseer.ts
```diff
@@ -4975,30 +5023,75 @@ class OverseerImpl implements AgentHooks {
...
+  // Fire-and-forget: the callers are synchronous (bindWorkpiece) or already past their last write,
+  // and getSharingManager() is async, so failures are logged rather than left as an unhandled
+  // rejection. Failing to restart is fail-open for the widened scope, hence the `error` level.
+  //
+  // Note that ensureAmbientCapsules() calls addGatekeeper() from inside open(), so on a shared
+  // workspace the first open after an ambient capsule appears bounces itself once; the capsule
+  // exists by then, so the client's retry is clean.
+  #restartIfShared(reason: string): void {
+    this.getSharingManager().then(sharing => {
+      if (sharing.listCollaborators().length === 0) return;
```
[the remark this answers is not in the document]
Maximo-Guk: This was fixed in `Bugfix: Restart only the role whose verification scope actually widened` 

So now #restartIfSessionsAffected(reason, affectedRole?) now takes the widened role, and each trigger passes the one it grows, if affectRole is omitted it will just sever regardless

packages/workshop-backend/src/overseer.ts
Maximo-Guk: We now have a `#liveSessions`, `joinSession(kind)` and a `#hasCollaboratorSession(role?)` which "joins"/increments count at all the sites collaborators can have access through and leaves when disposer is called

```
OverseerClientInterface
UseOverseerInterface
GadgetClientImpl / GatekeeperClientImpl
UseGadgetClientInterface
authorizeCollaborator
receiveExternalMessage
Subscription leases (#subscriptionLease, both interfaces)
```

https://github.com/cloudflare/cloudflare-os/pull/380/changes#diff-b21fc87bbe11ba6ce2f1fef7efe5dd1df886282ec94c5450a2cf8c0acf084fccR1543

`#restartIfSessionsAffected` calls (!this.#hasCollaboratorSession(affectedRole)) rather than checking the sharing graph, and the owner is counted as it's own kind and never triggers a restart

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Removed / dropped that commit! Thank you for the explanation

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Dropped the commit so we're no longer touching record https://github.com/cloudflare/cloudflare-os/pull/380#discussion_r3904790637

And pushed a commit for "Fix excludeObservers instead: detect the gatekeeper is out of scope for that user, de-register them from the gatekeeper, don't block." which should make it so observations will no be blocked for gatekeeper which are no longer in-scope

packages/workshop-backend/src/overseer.ts
```diff
@@ -4975,30 +5097,94 @@ class OverseerImpl implements AgentHooks {
...
+  // collaborator's until some gadget binds it, and binding one enters "use" scope having been in
+  // "build" scope since it was created. A session holding the other role therefore has no new
+  // verification requirements, and severing it would buy nothing. Omit the argument for a restart
+  // that isn't a widening at all (one that must sever regardless of role).
+  //
+  // Note that ensureAmbientCapsules() calls addGatekeeper() from inside open(), so on a shared
+  // workspace the first open after an ambient capsule appears bounces itself once; the capsule
+  // exists by then, so the client's retry is clean.
+  #restartIfSessionsAffected(reason: string, affectedRole?: CollaboratorRole): void {
+    if (!this.#hasCollaboratorSession(affectedRole)) return;
```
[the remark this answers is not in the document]
Maximo-Guk: This was addressed in "hold an admission lease across authorization" in 6401a607

packages/workshop-backend/src/overseer.ts
```diff
@@ -10575,10 +10847,12 @@ class UseOverseerInterface extends RpcTarget implements Overseer {
         this.impl.logger);
   }
 
+  #leaveSession: () => void;
   #leavePresence: () => void;
   #leaveOutputsFanout: () => void;
 
   [Symbol.dispose]() {
+    this.#leaveSession();
```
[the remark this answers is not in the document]
Maximo-Guk: This should be fixed in 6401a607, we're now keeping the lease alive per descendant and we added the following tests "a retained connection capability holds the gate open after its interface is gone" and "a retained gadget capability from a use session holds the gate open"

packages/workshop-backend/src/overseer.ts
```diff
@@ -4578,39 +4663,76 @@ class OverseerImpl implements AgentHooks {
...
     let gatekeeperIds = [...this.storage.gatekeepers.list()].map(gk => gk.id);
-    for (let observerId of observerIds) {
+    for (let observerId of unauthorized) {
       let observer = this.storage.observers.byObserverId.get(observerId);
       if (!observer) continue;
       this.storage.observers.delete(observer.profileId);
       await this.#removeObserverFromGatekeepers(observerId, gatekeeperIds);
     }
+    for (let observerId of outOfScope) {
+      await this.#removeObserverFromGatekeepers(observerId, [gatekeeperId]);
```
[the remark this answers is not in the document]
Maximo-Guk: 6401a607 does the per-removal recheck and 63fdaafa adds the serialization option too

Maximo-Guk:
> **Devin Review** found 4 new potential issues.
> 
> 🐛 4 issues in files not directly in the diff
> ### 🐛 [Failed restart permanently blocks connections](https://github.com/cloudflare/cloudflare-os/blob/ce025bb56f5facecf2fa52cacb0c5233d86bd3fe/packages/workshop-backend/src/overseer.ts#L5270-L5276)
> If `storage.sync()` rejects, `scheduleAccessRestart` swallows the failure and retains its settled promise. Quarantined connections stay inaccessible, while later changes cannot retry the reset.
> 
> ### ⚠️ [Concurrent disable leaves hook enabled](https://github.com/cloudflare/cloudflare-os/blob/ce025bb56f5facecf2fa52cacb0c5233d86bd3fe/packages/workshop-backend/src/overseer.ts#L2377-L2382)
> If `disableHook` runs while `enableHook` awaits its controller, it sees the old disabled record and returns. `enableHookRecord` then enables the hook, losing the disable request.
> 
> ### 🟥 [Retained bindings bypass observer enforcement](https://github.com/cloudflare/cloudflare-os/blob/ce025bb56f5facecf2fa52cacb0c5233d86bd3fe/packages/workshop-backend/src/overseer.ts#L4819-L4820)
> After unbinding, a retained loopback still opens the connection because `startGatekeeperSession` never revalidates its edge. Exclusion removes the collaborator’s registration, allowing continued reads without observer checks.
> 
> ### 🟨 [Hook enable misses retained gadget capabilities](https://github.com/cloudflare/cloudflare-os/blob/ce025bb56f5facecf2fa52cacb0c5233d86bd3fe/packages/workshop-backend/src/overseer.ts#L2381-L2382)
> A collaborator can retain a gadget-minted child after disposing counted wrappers. Enabling a hook then skips the restart, leaving that child authorized against newly introduced data.
> 
> 2 flags not posted on this PR by your GitHub settings — view them in Devin Review. ([Configure](https://cloudflare.devinenterprise.com/org/cloudflare/settings/review))
> 
> [ Devin Review ](https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/380)

> Failed restart permanently blocks connections
 
The stuck state described requires the object to *survive* the failed flush, and it cannot. Any exception escaping `blockConcurrencyWhile()`'s callback terminates and resets the DO.

> Concurrent disable leaves hook enabled:

The state is visible in the UI and retryable, and I think is super unlikely to happen in practice, deferring

> Retained bindings bypass observer enforcement

Documented in docs/observers.md ("Known gaps")

> Hook enable misses retained gadget capabilities

Documented in docs/observers.md ("Known gaps")

packages/workshop-backend/src/overseer.ts
```diff
@@ -7757,28 +8131,91 @@ class OverseerImpl implements AgentHooks {
...
+  // Gatekeeper ids reachable from a "use" collaborator's session, and therefore all of their
+  // verification scope: everything bound by some non-provisional gadget (the gadget UI they drive
+  // can invoke it), plus every connection with an *enabled* hook -- a hook is a live write channel
+  // into the gadget it wakes, delivering the connection's data into state that collaborator can
+  // open, regardless of binding edges -- plus, transitively, every env target of a reachable
+  // agent spawner: connectToGadget mints the bound spawner's loopback with no role filter
+  // (getEnvForLoader), and spawn/spawnCallable seeds the spawned agent's bindings from
+  // config.env (spawnAgent), handing the collaborator an agent that reads those connections
+  // with the creator's authority and returns their data.
+  #useScopeGatekeeperIds(): Set<WorkpieceId> {
```
Maximo-Guk, line 9093: This method now does all the checks for use scope, goes over all gadgets with visible bindings, all enabled hooks and agent spawners

packages/workshop-backend/src/overseer.ts
```diff
@@ -7849,6 +8318,43 @@ class OverseerImpl implements AgentHooks {
...
+  // The authorization gate every non-owner entry point (open(), receiveExternalMessage()) must
+  // pass through: resolve the caller's effective role, then verify them as an observer of
+  // everything this workspace has read. Returns null for no access; verification failures throw.
+  // A caller that requires at least `requireRole` (e.g. receiveExternalMessage needs "build")
+  // passes it so an insufficient role is denied *before* verification runs -- otherwise the caller
+  // would be verified (real addObserver calls, a persisted observer record) only to be turned
+  // away, or worse, told to fix a verification failure that can never grant them access.
+  // `configureCb` is forwarded to ensureObserver to prompt for unconfigured account choices;
+  // without it, verification is non-interactive and an unconfigured binding denies access.
+  async authorizeCollaborator(
```
Maximo-Guk, line 9287: This is now the canonical authorization check for entrypoints like open(), receiveExternalMessage etc so they don't have to duplicate their own authorization checks. Note: receiveExternalMessage wasn't even verifying scopes before this

packages/workshop-backend/src/overseer.ts
```diff
@@ -8430,21 +8955,66 @@ export class OverseerDurableObject extends DurableObject<Cloudflare.Env> {
...
+    // this path never constructs one of the counted client interfaces, so without a lease a scope
+    // widening mid-call would find no session to sever and the reply would be produced under
+    // stale verification. With it, the widening resets the DO, this call dies with it, and the
+    // caller retries. The lease deliberately starts only after authorizeCollaborator returns:
+    // verification itself is covered by that call's own internal lease, and a caller it turns
+    // away must never have counted at all -- a stranger racing an addGatekeeper would otherwise
+    // cause a needless workspace reset. The handoff between the two leases is microtask-only,
+    // the same argument as open()'s.
+    //
+    // KNOWN GAP, tolerable only while nothing calls this endpoint: the lease ends when this call
```
Maximo-Guk, line 9927: We aren't use externalMessage anywhere yet, and this endpoint wasn't even checking scopes before this PR, so I'd like to defer any of this extra work AI reviewers flagged for now

packages/workshop-backend/src/overseer.ts
```diff
@@ -9097,6 +9678,35 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
...
+  // interface that minted it is disposed, so one that escaped the count would let a scope
+  // widening find no session to sever while e.g. a chat or action subscription kept streaming
+  // gatekeeper-derived data. Applied to every subscription-returning method uniformly -- one
+  // invariant for every export, rather than per-subscription reasoning about which could carry
+  // sensitive data. The owner's subscriptions pass through uncounted.
+  #subscriptionLease(subscription: RpcStub<{}>): RpcStub<{}> {
+    let kind = this.#mintedCapabilityKind();
+    if (!kind) return subscription;
+    let leave = this.impl.joinSession(kind);
+    // @ts-expect-error Bugs in native RPC types make this not work currently.
```
Maximo-Guk, line 10668: I'm not sure about this ^^ will double check what is going on here

packages/workshop-backend/src/overseer.ts
```diff
@@ -10590,6 +11220,22 @@ class UseOverseerInterface extends RpcTarget implements Overseer {
...
 
+  // Count a subscription handle toward #hasCollaboratorSession for its own lifetime, exactly as
+  // OverseerClientInterface.#subscriptionLease does for "build" sessions -- a client can dispose
+  // this interface while retaining the subscription. Applied uniformly, including to the inert
+  // subscriptions: every export minted into a collaborator session counts, rather than
+  // per-subscription reasoning about which could carry data.
+  #subscriptionLease(subscription: RpcStub<{}>): RpcStub<{}> {
+    let leave = this.impl.joinSession("use");
+    // @ts-expect-error Bugs in native RPC types make this not work currently.
+    return new NativeRpcStub<{}>({
```
Maximo-Guk, line 12201: Same here

packages/workshop-backend/src/overseer.ts
```diff
@@ -11207,24 +11898,85 @@ class SlashCommandAuthorizerImpl extends NativeRpcTarget implements ObservationA
...
+// issued it could never be revoked, and a holder would keep a live write channel into the gadget
+// after a disable/delete shrank every collaborator's verification scope.
+function makeHookFiringCallback(impl: OverseerImpl, hookId: number): NativeRpcStub<RpcTarget> {
+  // The proxy target must be callable for the `apply` trap to ever fire (a Proxy over a
+  // non-callable target is itself non-callable), and the bindHook contract allows the bound
+  // callback to be a function type, invoked by calling the firing's callback directly. An arrow
+  // function also has no `prototype` own-property to conflict with the wildcard `get` below.
+  // TODO: Same workerd bug as startGatekeeperHook: a Proxy returned as an RpcTarget is judged
+  //   non-pipelineable, so wrap it in a stub manually.
+  return new NativeRpcStub(new Proxy((() => {}) as unknown as RpcTarget, {
```
Maximo-Guk, line 12908: Hmm, not sure why we had to go with this proxy approach here, will also investigate this further

packages/workshop-backend/src/overseer.ts
```diff
@@ -1523,6 +1528,32 @@ class OverseerImpl implements AgentHooks {
...
+      if (left) return;
+      left = true;
+      this.#liveSessions[kind]--;
+    };
+  }
+
+  // Whether some non-owner session of `role` (or of either role, when omitted) is live right now.
+  // The owner is never an observer, so their session is counted separately and never counts here.
+  #hasCollaboratorSession(role?: CollaboratorRole): boolean {
+    if (role === undefined) return this.#liveSessions.build > 0 || this.#liveSessions.use > 0;
```
kentonv: This line risks failing to be updated if a new role is added. Perhaps best to iterate over the properties?

packages/workshop-backend/src/overseer.ts
```diff
@@ -4679,42 +4694,55 @@ class OverseerImpl implements AgentHooks {
...
   //     their observer record (best-effort removeObserver on all gatekeepers). They are no longer
   //     set up to observe; if they regain access they reconfigure from scratch (Step 3).
-  // If no named observer can reach the observation, it is allowed.
+  // If no named observer can reach the observation, it is allowed. The teardown yields on every
+  // removal, so each observer is re-classified against current state adjacent to their own -- an
+  // observer put back in scope mid-teardown blocks the observation instead of being de-registered.
   async #enforceExcludeObservers(gatekeeperId: number, observerIds: string[]): Promise<void> {
     let sharing = await this.getSharingManager();
 
+    // Classify one named observer against *current* state. All reads are synchronous, so a
```
kentonv: This change adds a lot of complexity and inefficiency to re-scan the state for each iteration of the loop.

A simpler and more efficient approach would be to make all the calls to `#removeObserverFromGatekeepers` synchronously, and then await all the promises in a single `Promise.all()`.

This, I think, closes the race just as well. (It's not really closed until the locking changes in the next commit, of course.)

docs/observers.md
```diff
@@ -414,14 +415,26 @@ gadget-minted children alike:
...
+| `bindWorkpiece()` for a permanent (non-`chatId`) edge onto a vendor-backed connection — or onto a legacy (pre-`creationSpec`) one, which nobody *can* be verified against, so it restarts and quarantines and fresh `use` opens then fail closed on the reconnect-required error | **use** scope — the gadget UI a `use` session drives can now invoke it |
 | A merge that promotes a pending gadget or a pending binding edge into `use` scope | **use** scope, same reason |
-| `enableHook` on a vendor-backed connection not already in `use` scope | **use** scope — the hook delivers the connection's data into a gadget a `use` session can open (a hook waking a still-provisional gadget stays out of `use` scope until promotion; the merge diff reports that widening) |
+| `enableHook` on a vendor-backed connection not already in `use` scope | **use** scope — the hook delivers the connection's data into a gadget a `use` session can open. Every enabled hook widens: the record's `gadgetId` is bookkeeping-only and can diverge from the callback's real target, so no provisional-gadget exemption is derived from it |
```
kentonv: This is pedantic. It is building complexity around a known, obscure bug which is likely to be fixed at some point. It's not worth worrying about. Trusting `gadgetId` is fine.

kentonv (commented):
> Bugfix: Don't let a delayed abort swallow a concurrent access change.
> Bugfix: Close the restart barrier with blockConcurrencyWhile().

These two commits do not actually achieve anything. Opus seems very confused in the comments.

The fact of the matter is that if there is any concurrent activity in-flight when the abort() hits, that concurrent activity may be canceled -- obviously.

Adding a second sync() doesn't change this, because nothing guarantees that the concurrent action was "complete" before the sync() began -- the sync() could interject in the middle.

Adding a blockConcurrencyWhile() also doesn't change this, because that also does not guarantee that the concurrent action was "complete" -- it actually prevents the concurrent action from making any more progress.

So despite these commits, the status quo is the same: any concurrent activity might be arbitrarily canceled in the middle.

I think that Opus is confused about output gates, and about my statement "An abort() breaks the output gate, causing existing in-flight writes never to land."

It's true that if you perform a synchronous write immediately followed by an abort(), the write will never land. However, it's also the case that no outside observer can ever know that you performed said write. Effectively, the write never happened in the first place.

I think Opus is worried that a client might concurrently make some change and *get a successful response*, but then the abort() rolls back that change. This cannot happen: the client cannot receive a successful response unless all writes actually completed. That's the whole point of output gates.

In any case, these two commits should be removed as they add complexity and misleading documentation for no actual value.

kentonv (approved):

Maximo-Guk:
> > Bugfix: Don't let a delayed abort swallow a concurrent access change.
> > Bugfix: Close the restart barrier with blockConcurrencyWhile().
> 
> These two commits do not actually achieve anything. Opus seems very confused in the comments.
> 
> The fact of the matter is that if there is any concurrent activity in-flight when the abort() hits, that concurrent activity may be canceled -- obviously.
> 
> Adding a second sync() doesn't change this, because nothing guarantees that the concurrent action was "complete" before the sync() began -- the sync() could interject in the middle.
> 
> Adding a blockConcurrencyWhile() also doesn't change this, because that also does not guarantee that the concurrent action was "complete" -- it actually prevents the concurrent action from making any more progress.
> 
> So despite these commits, the status quo is the same: any concurrent activity might be arbitrarily canceled in the middle.
> 
> I think that Opus is confused about output gates, and about my statement "An abort() breaks the output gate, causing existing in-flight writes never to land."
> 
> It's true that if you perform a synchronous write immediately followed by an abort(), the write will never land. However, it's also the case that no outside observer can ever know that you performed said write. Effectively, the write never happened in the first place.
> 
> I think Opus is worried that a client might concurrently make some change and _get a successful response_, but then the abort() rolls back that change. This cannot happen: the client cannot receive a successful response unless all writes actually completed. That's the whole point of output gates.
> 
> In any case, these two commits should be removed as they add complexity and misleading documentation for no actual value.

Makes sense I should have caught this myself too thanks for the explanation, I dropped the commits in cb9a6bd5.

packages/workshop-backend/src/overseer.ts
Maximo-Guk: This is fine

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Done in 8f80fe7d it iterates `#liveSessions` and skips only the owner's count

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Thank you, done in a7802504

packages/workshop-backend/src/overseer.ts
```diff
@@ -7757,28 +8072,93 @@ class OverseerImpl implements AgentHooks {
...
+    // and chat bindings, and chats are owner/build-only), and a spawner's env is fixed at
+    // creation -- so every widening lands on the diffs around the sites that grow the roots
+    // (bindWorkpiece, mergeChatChanges, enableHookRecord), with no dedicated trigger.
+    // Spawner-to-spawner env edges are legal, hence the worklist; an env target that is a gadget
+    // is skipped -- env gadgets are non-pending, so their bindings are covered by the first loop.
+    let pending = [...ids];
+    while (pending.length > 0) {
+      let spec = this.storage.gatekeepers.get(pending.pop()!)?.creationSpec;
+      if (spec?.type !== "agentSpawner") continue;
+      for (let target of Object.values(spec.config.env)) {
```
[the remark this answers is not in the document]
Maximo-Guk: This isn't real, from Fable: Not a bug. Spawned agents cannot call `requestConnection`: when `agentContext.spawnerConfig` is set

packages/workshop-backend/src/overseer.ts
```diff
@@ -11207,24 +11840,85 @@ class SlashCommandAuthorizerImpl extends NativeRpcTarget implements ObservationA
...
+      return Reflect.apply(record.callback as any, undefined, args);
+    },
+    get(_target, prop) {
+      // All wildcard properties of a stub appear as functions, so `then` must come back
+      // undefined (this is not a thenable) and symbols are never RPC methods -- the same
+      // dispositions as getGadgetFacet's proxy over the gadget facet.
+      if (typeof prop === "symbol" || prop === "then") return undefined;
+      return async (...args: unknown[]) => {
+        let record = requireLiveHook(impl, hookId);
+        return Reflect.apply((record.callback as any)[prop], record.callback, args);
```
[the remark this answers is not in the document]
Maximo-Guk: Observer enforcement catches mistakes, it is not a critical security boundary. The bindHook contract explicitly permits hooks to pass and return capabilities, and the holder of a firing is a gatekeeper bound by the session contract ("MUST use a HookInitiator to obtain a fresh callback")

packages/workshop-backend/src/overseer.ts
Maximo-Guk: it exists so the stored persistent callback stub never leaves the DO, since once handed out it could never be revoked similar to getGadgetHookEntrypoint()

Maximo-Guk:
> * **[P1] `packages/workshop-backend/src/overseer.ts:12845`**: `openSession()` returns an independently retained capability without a session lease. After disposing its parent, a build collaborator can retain an agent-spawner session; subsequent connection additions see no live build session, skip restart/reverification, and the spawner can access the new connection transitively through a gadget. Lease the returned capability graph or enforce an authorization generation on every call.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/33807550861)

Same reply as this one https://github.com/cloudflare/cloudflare-os/pull/380#discussion_r3929254546, this is ok

Maximo-Guk:
Validated sharing in previews with @ndisidore too

PR #382, opened 2026-08-28
Restricted data: govern restricted reads by observer verification

Maximo-Guk:
Stacked on #381.

Why:

Before this PR: Once a user has observed a sensitive data source, we block sharing the resource altogether. We decided we can lift this restriction now that we have mechanisms in place to verify the observers permissions.

What:

Makes it so that now when restricted data is observed, the workspace can still be shared by verifying the observers access against the resource by reusing the same mechanisms we had already introduced in the original observer verification PR.

Frontend behavior is unchanged here, so the UI still declines to share a workspace even though the server now permits it. The UI will be done in https://github.com/cloudflare/cloudflare-os/pull/308


Maximo-Guk:
> * **High:** `packages/workshop-backend/src/overseer.ts:9685` denies only the failed re-open while prior collaborator capabilities remain live. After `addObserver()` confirms revoked access, those owner-backed sessions can continue reading restricted data. Terminal re-verification failures must revoke existing sessions.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/34541232095)

This is the lazy-revocation we already document explicitly, at the throw site, in [docs/sharing.md](https://github.com/cloudflare/cloudflare-os/blob/4a886852edb623c10b3b08747a2334e0b8cb7422/docs/sharing.md#lazy-revocation) [docs/observers.md edge case 3](https://github.com/cloudflare/cloudflare-os/blob/4a886852edb623c10b3b08747a2334e0b8cb7422/docs/observers.md#6-edge-cases), and in the Known limitations of [plans/restricted-data-sharing.md](https://github.com/cloudflare/cloudflare-os/blob/4a886852edb623c10b3b08747a2334e0b8cb7422/plans/restricted-data-sharing.md). The integration test "a failed re-verification denies that open and nothing else" also pins it.

Maximo-Guk:
> * **High:** `packages/workshop-backend/src/overseer.ts:8028` excludes unbound producers from `use` verification. Restricted data can be copied into gadget state/UI, then exposed to a `use` collaborator who never passes that producer’s `addObserver()`. Include all restricted producers in `use` scope or prevent their data reaching gadget-visible state.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/33207434161)

This is the "never-bound producers" item under Accepted tradeoffs in plans/restricted-data-sharing.md, described as a known security risk and marked at the site by the TODO(known-risk)

Maximo-Guk:
> * **High:** `packages/workshop-backend/src/overseer.ts:5819` ignores observer IDs before their record is persisted. If one `addObserver()` finishes while another stalls, a restricted observation can exclude that pending observer, be admitted, then become visible after verification completes. Track pending IDs before registration.

Known it's the mid-registration item under Known limitations in both docs/observers.md and the plan doc, marked by the TODO(observer-races)

> * **High:** `packages/workshop-backend/src/overseer.ts:5863` treats an unbound producer as unreachable, but retained binding loopbacks remain callable after unbind. An excluded collaborator can directly receive the restricted observation. Revalidate the binding edge per call or keep them in scope.

This is the Step 5 gap in docs/observers.md (the "left their scope does not yet imply cannot reach" paragraph)

> * **Medium:** `packages/workshop-backend/src/overseer.ts:5706` ignores active share links. An unverifiable restricted read succeeds, then permanently makes an already-distributed link unredeemable while it remains active and listed. Include outstanding links in this check.

Fixed on the #assertUnverifiableProducerUnshared now treats any outstanding share link as "shared", the same predicate `removalBlockedByRestrictedData` uses

> * **Medium:** `packages/workshop-frontend/src/ShareModal.tsx:778` still hides all sharing controls after restricted data is observed. This prevents the newly supported verified sharing and also blocks removing collaborators or revoking links.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/34535471304)

The UI changes land in a follow up PR

Maximo-Guk:
> * **High:** `packages/workshop-backend/src/overseer.ts:9479` allows concurrent first opens to mint different observer IDs. The last persisted record orphans the other live ID, so later exclusions for it are ignored. Serialize verification per profile.

Known, race documented in known limitations in observer.md doc, marked by the TODO at the site. ( pre-existing won't fix in this PR )

> * **High:** `packages/workshop-backend/src/overseer.ts:5563` assumes scope restarts reverify every consumer, but persisted collaborator-initiated agent turns resume without authorization. They can send newly reachable restricted data to the collaborator’s model or external response target. Persist turn ownership and reauthorize on resume.

Gap is documented at receiveExternalMessage (overseer.ts and in docs/observers.md under "Known gap — the agent turn an external message starts". Nothing yet calls the external chat endpoint, so I'm continuing to defer this. And platform-gateway mode should be used if you want to restrict what AI providers your deployment uses.

> * **Medium:** `packages/workshop-backend/src/overseer.ts:7578` awaits after snapshotting ambient accounts. A stale reconciliation can subsequently delete a newly installed restricted producer, leaving its historical ID missing and permanently blocking future sharing. Fetch the sharing manager first or revalidate the account snapshot.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/34536383111)

In ensureAmbientCapsules the only await after the accounts snapshot is getSharingManager(), which this branch added so that the guard could run synchronously: the gatekeeper records are snapshotted after it, and the loop from snapshot through removalBlockedByRestrictedData to removeGatekeeper has no await at all

Maximo-Guk:
> * **High:** `packages/workshop-backend/src/overseer.ts:7067` does not count active collaborator agent turns as live sessions. After disconnect, an owner can add/bind a producer before the turn lazily seeds bindings, exposing restricted data to the collaborator’s model without verification. Hold a role lease through `#unregisterRunningAgent()`.

This is pre-existing same root cause as the [resumed-turn finding above](https://github.com/cloudflare/cloudflare-os/pull/382#issuecomment-5627780266) as the comment in overseer.ts already says. The window is narrower than "any time after disconnect", deferring and on top of that if the deployment admin wants to restrict model egress they can just use platform gateway mode

> * **High:** `packages/workshop-backend/src/overseer.ts:2734` only restarts on hook enable when counted wrappers remain. Gadget-returned child capabilities are independently retained, so a collaborator can dispose wrappers, avoid restart, and read restricted hook data through the child. Unconditionally restart the target gadget facet when enabling a hook.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/34537893863)

Documented in docs/observers.md "Known gap — enableHook neither counts nor aborts gadget-minted children", deferring

packages/workshop-backend/src/overseer.ts
```diff
@@ -9167,6 +9330,14 @@ class OverseerImpl implements AgentHooks {
...
   // open, since nothing they can reach needs verification against it. An in-scope one still
   // throws, fail-closed (and "build" scope is everything, so it always throws there).
+  //
+  // TODO(known-risk): scoping by role means a "use" collaborator is never verified against a
+  //   producer outside their scope (one no gadget binds and no enabled hook feeds) -- yet
+  //   restricted data read from such a producer can reach gadget state and the UI they drive,
+  //   because provenance is not tracked past the observation. Deliberately accepted for v1; the
+  //   required fix (verify every restricted producer, or isolate restricted data by provenance) is
+  //   recorded under "Known security risk -- never-bound producers" in
+  //   plans/restricted-data-sharing.md, and worked through in docs/observers.md edge case 4.
```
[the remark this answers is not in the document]
Maximo-Guk: This is the "never-bound producers" item under Accepted tradeoffs in plans/restricted-data-sharing.md, marked at this site by the TODO(known-risk) comment a few lines up and worked through in docs/observers.md edge case 4

packages/workshop-backend/src/overseer.ts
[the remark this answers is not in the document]
Maximo-Guk: Known and documented as the Step 5 gap in docs/observers.md ("left their scope does not yet imply cannot reach").

packages/workshop-backend/src/overseer.ts
```diff
@@ -5683,6 +5826,11 @@ class OverseerImpl implements AgentHooks {
     let outOfScope: string[] = [];
     for (let observerId of observerIds) {
       let observer = this.storage.observers.byObserverId.get(observerId);
+      // TODO(observer-races): a first-time ensureObserver registers a freshly minted id with the
+      // gatekeepers *before* the record (and so byObserverId) is persisted, so an id named here
+      // during that window -- which can park on the config modal -- reads as unknown and the
+      // observation is admitted to a collaborator it names. Fix: an in-memory map of pending ids
+      // consulted here, failing closed. Lands with observer-verification-fixes (7ba93821).
       if (!observer) continue;  // not an active observer -> ignore
```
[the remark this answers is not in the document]
Maximo-Guk: Known. This is the mid-registration item under Known limitations in docs/observers.md (Step 5, item 1) and in plans/restricted-data-sharing.md, marked here by the TODO(observer-races) comment directly above the continue. A first-time ensureObserver registers a freshly minted id with the gatekeepers before the record is persisted, so an id named during that window reads as unknown and the observation is admitted

packages/workshop-backend/src/overseer.ts
```diff
@@ -11998,13 +12167,12 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
...
-          "shared.");
-    }
-
-    return (await this.impl.getSharingManager()).addCollaborator({
+    let sharing = await this.impl.getSharingManager();
+    // Asserted in the same synchronous block as the grant's storage write (after every await): a
+    // check ahead of the awaits above could pass, a concurrent producer-connection removal land
+    // during the yield, and the grant still be written past it.
+    this.impl.assertNewSharingAllowed();
+    return sharing.addCollaborator({
```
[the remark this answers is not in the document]
Maximo-Guk: Fixed

docs/observers.md
```diff
@@ -672,6 +732,31 @@ already in the JSDoc in `gatekeeper.ts`; add anything missing there rather than
...
    handles: the gatekeeper still knows the id, but the observer can no longer reach what it
    produces, so they are de-registered from it instead of blocking.
+8. **Removing a connection that read restricted data** — while the workspace is latched
+   (`containsRestrictedData`) *and* shared, `GatekeeperClient.remove()` refuses for the
+   *producer* connections — those through which restricted data was actually read, derived from
+   the permanent action log (`restrictedProducerIds`); non-producers stay removable. The record
+   is what observer verification runs against, and the restricted data outlives it in chat
+   history and storage, so deleting it would let a never-verified collaborator open unchecked.
+   Outstanding share links block removal the same way: their keys never expire, and redemption
+   is gated at open() only while the record exists. The remedy is to remove collaborators and
```
Maximo-Guk, line 742: Not a big fan of this line so I'll likely reword it as it makes it seem like you'll be able to share again, but you won't, it's just the remedy to get the owner working again. This workspace will now be owner only forever until we add an escape hatch

packages/workshop-backend/src/overseer.ts
```diff
@@ -5529,6 +5539,31 @@ class OverseerImpl implements AgentHooks {
       await this.#enforceExcludeObservers(gatekeeperId, description.excludeObservers);
     }
 
+    if (description.containsRestrictedData) {
+      // Resolved here rather than up front: on a cold DO this is an RPC to the owner's User DO,
+      // and an ordinary unrestricted observation must not pay for it. The producer record is read
+      // *after* that await, so the check below and the latch are one synchronous block -- a record
+      // read before the await could be stale by the time it is checked, and latching against a
+      // stale one permanently bricks sharing.
```
kentonv: slop comment

packages/workshop-backend/src/overseer.ts
```diff
@@ -5529,6 +5539,31 @@ class OverseerImpl implements AgentHooks {
...
+      let sharing = await this.getSharingManager();
+      let producer = this.storage.gatekeepers.get(gatekeeperId);
+
+      // An in-flight facet RPC can outlive removeGatekeeper, so a restricted observation can
+      // arrive naming a connection this workspace no longer has. Latching a missing producer id
+      // permanently bricks sharing (assertNewSharingAllowed's missing-record branch), so refuse
+      // the read instead -- including on an unshared workspace, where nothing else would stop it.
+      // This same read is what refuses a connection removed during the exclusion awaits above,
+      // where the latch is not yet set and so removalBlockedByRestrictedData does not yet protect
+      // the producer.
```
kentonv: slop

packages/workshop-backend/src/overseer.ts
```diff
@@ -5529,6 +5539,31 @@ class OverseerImpl implements AgentHooks {
...
+      if (!producer) {
+        throw new Error(
+            "This observation was blocked because it contains sensitive data, but the " +
+            "connection it was read through has been removed from this workspace.");
+      }
+      this.#assertUnverifiableProducerUnshared(producer, sharing);
```
kentonv, line 5563: Just delete all this, and delete `#assertUnverifiableProducerUnshared`, the check it is performing is not important and all this complexity and these slop comments are not worth it.

```suggestion
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -5657,6 +5692,114 @@ class OverseerImpl implements AgentHooks {
...
+    // otherwise list collaborators -- so it reports only that the workspace is shared, naming
+    // neither the collaborators nor their profile ids (the full email on OAuth/CF Access
+    // deployments).
+    throw new Error(
+        "This observation was blocked because it contains sensitive data, but it was read " +
+        "through a connection that cannot verify anyone's access to that data, and this " +
+        "workspace is shared. Its collaborators must be removed and its share links revoked " +
+        "before this data can be read.");
+  }
+
```
kentonv, line 5735: This method is pedantic. We don't have to handle the case that aiModel and agentSpawner set `constainsRestrictedData`, because they never do. Legacy records with no creationSpec are also really not worth this much complexity -- users with very old workspaces will just have to remake those workspaces.

```suggestion
```

packages/workshop-backend/src/overseer.ts
```diff
@@ -5657,6 +5692,114 @@ class OverseerImpl implements AgentHooks {
...
+  // The connection ids this workspace has read restricted data through: the producers the latch
+  // guards. Derived by scanning the action log for observations whose description carries
+  // `containsRestrictedData`, since nothing else records which connection a latched read came
+  // through.
+  restrictedProducerIds(): Set<WorkpieceId> {
+    let producers = new Set<WorkpieceId>();
+    for (let record of this.storage.actions.list()) {
```
kentonv, line 5742: This loop is unacceptably expensive.

packages/workshop-backend/src/overseer.ts
```diff
@@ -5657,6 +5692,114 @@ class OverseerImpl implements AgentHooks {
...
+    let producers = this.restrictedProducerIds();
+    if (producers.size > 0 && !producers.has(id)) return false;
+    return sharing.listCollaborators().length > 0 || sharing.listShareLinkRecords().length > 0;
+  }
+
+  // Refuse a new sharing grant once the workspace has read restricted data through a connection
+  // that can no longer verify a recipient's access to it -- one that has since been removed, or
+  // that never had a vendor account behind it. Every other producer verifies its collaborators at
+  // each open, so sharing stays available.
+  assertNewSharingAllowed(): void {
```
kentonv, line 5779: This is too weird and pedantic.

I'm pretty sure at present there's actually no UI to delete a gatekeeper.

How about if and when we create one, we make the user check a box saying: "I certify that no sensitive data from this gatekeeper has been retained in this workspace."

This should apply to any gatekeeper, not just ones that have generated observations with `containsRestrictedData`.

kentonv (commented):
I think this change should be significantly simplified.

After simplification the change should be mostly (entirely?) deletions.

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Deleted!

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Deleted, sounds good!

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Nothing scans the action log anymore since the removalBlockedByRestrictedData and assertNewSharingAllowed are now deleted, so this should be fixed now

packages/workshop-backend/src/overseer.ts
Maximo-Guk: Sounds good, deleted it and added your certification idea in to docs/observers.md

packages/workshop-backend/src/overseer.ts
```diff
@@ -5529,6 +5520,10 @@ class OverseerImpl implements AgentHooks {
       await this.#enforceExcludeObservers(gatekeeperId, description.excludeObservers);
     }
 
+    if (description.containsRestrictedData) {
+      this.storage.containsRestrictedData.put(true);
```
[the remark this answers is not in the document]
Maximo-Guk: Intentional. We're planning on adding a certification UI when removing unbinding/removing connections

kentonv (approved):

PR #387, opened 2026-08-29
Usability - LLAMA: Allow context window and output limit to be set per model

metahertz:
## What does this change?

Self-hosted models reached through the `ollama` provider have no pi catalog entry (`catalogModel` returns undefined for it) and no `SUGGESTED_MODELS` row, so `modelTokenWindow` always falls through to its generic defaults of 128k context and 4096 output tokens regardless of what the server actually serves. The Add Model dialog offers no way to correct this, so the values are unreachable for exactly the models that need them most.

Both fields are optional and blank means "infer as before", so existing configurations are unaffected. A supplied value is validated as a positive integer at submit time, since a bad one would otherwise surface much later as a provider-side rejection.

Tests cover the fallback, the override, and an override winning over a catalogued model's own limits. Note that I could not execute the suite in my environment: `vitest run` fails at import for every workshop-backend test file, including on a pristine checkout, so the failure is unrelated to this change. The three affected packages typecheck.

## Checklist

- [X]  This is a small, concrete change; it is not a feature, refactor, or low-value cleanup.
- [X]  I understand that maintainers decide whether the change is obviously correct and trivially verifiable.
- [X]  I have read and followed the contribution guidelines.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/387" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/packages/workshop-backend/__tests__/ai-models.test.ts
+++ b/packages/workshop-backend/__tests__/ai-models.test.ts
@@ -461,6 +461,56 @@ describe("getModel direct routing (no gateway)", () => {
     expect(handle.model.baseUrl).toBe("http://my-ollama:11434/v1");
   });
 
+  it("falls back to default token limits for an uncatalogued Ollama model", () => {
+    // The ollama provider has no pi catalog and no SUGGESTED_MODELS row, so an unconfigured
+    // self-hosted model lands on the generic defaults -- which is exactly why the overrides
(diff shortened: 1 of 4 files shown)


metahertz:
Welp, there ends that experiment, next harness!

PR #397, opened 2026-08-31
Make release manifest shortNames legal install slugs

Maximo-Guk:
The deploy wizard sends a gatekeeper's manifest shortName to the deploy service as the install slug verbatim, and the service turns that slug into a GATEKEEPER_<SLUG> binding name - so it validates against `/^[a-z][a-z0-9]*$/`. gatekeeper-mcp-portal was the only installable gatekeeper whose derived shortName carried a hyphen, so mcp-portal failed zod validation on the first line of install(), before the workflow was ever created. Users saw a redacted "Something unexpected went wrong on our side." with no workflow logs to debug from.

- Adds releaseShortName() (manifest-lib.ts:338), which folds the package name into the slug charset by stripping everything outside [a-z0-9]. ( illegal characters as they're not supported in binding name for router worker )
- Adds collision guard in generateManifest() throwing if two gatekeepers fold to the same slug.

Also added tests for `shortName is a legal slug and matches its BASE_URL path`, `releaseShortName() folds correctly and no-ops on conforming names`, `shortNames are unique across the real manifest`

Also made mcp-portal and cf-portal singleton gatekeepers in the manifest

### DIFF
--- a/scripts/release/manifest-lib.test.ts
+++ b/scripts/release/manifest-lib.test.ts
@@ -178,6 +178,12 @@ test("worker entries carry the deploy contract", () => {
     }
   }
 
+  // The MCP connectors are install-once for SINGLETON's second reason: they take no inputs, so a
+  // second install could not differ from the first — it would only mint a second vendor id.
+  assert.equal(workers["gatekeeper-mcp"].singleton, true);
+  assert.equal(workers["gatekeeper-mcp-portal"].singleton, true);
+  assert.equal(workers["gatekeeper-homeassistant"].singleton, true);
+
   // Module blobs are content-addressed.
   for (const [name, entry] of Object.entries(workers)) {
     assert.ok(entry.modules.some((m) => m.name === entry.mainModule),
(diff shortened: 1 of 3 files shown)


Maximo-Guk:
> * **[P2] Keep the MCP portal vendor ID consistent across deployment modes** (`scripts/release/manifest-lib.ts:499`): Releases derive `mcpportal` from `GATEKEEPER_MCPPORTAL`, while dev/preview derive `mcp_portal` from `GATEKEEPER_MCP_PORTAL`. Blueprints persist this ID and require exact matching, so MCP portal blueprints exported between those Workshop instances cannot be configured. Align binding-derived IDs across deployment modes or separate the canonical vendor ID from the install slug.
> 
> [github run](/cloudflare/cloudflare-os/actions/runs/33693565704)

This isn't a real issue. The deploy flow already provisions gatekeepers with prefixes ( i.e gk-google-2 etc ) and regardless no prebuilt blueprint currently declare bindings anyways and mcpportal doesn't work at all currently in deploy gatekeeper install flow

maxwellpeterson (approved):

PR #427, opened 2026-09-02
Add createExternalResource agent tool

ndisidore:
Adds the `createExternalResource` tool as discussed with respect to Google Drive: rather than granting whole-Drive access so the agent can make one document, the agent creates a new resource of a creatable type through an already-connected account.

This adds the other half of `requestConnection`: the agent mints a brand-new resource of a type the vendor marks creatable, through an account the user already connected.  The binding is live immediately (simliar to what `createGadget` does), the turn continues, and the gatekeeper simulates the resource until the user approves the creation, which queues first and applies first, so dependent edits are safe to stack behind it.

:lock_with_ink_pen:   One thing worth noting: thus far we have been following that we use the minting account's credentials, and for creations that account is the turn initiator's. This means a collaborator-driven creation lands in the collaborator's provider account. If we want this to be the gadget owner's we can make that swap.

Auto approval is not supported yet.

Screenshot from 2026-09-07 16-50-55
Screenshot from 2026-09-07 16-53-21

First consumer is Google Doc creation, stacked as #428

### DIFF
--- a/packages/workshop-backend/src/agent-compaction.ts
+++ b/packages/workshop-backend/src/agent-compaction.ts
@@ -1,4 +1,5 @@
-import {SUGGESTED_MODELS, WORKERS_AI_OUTPUT_LIMIT, type AiChatMessage, type AiModelConfig}
+import {SUGGESTED_MODELS, WORKERS_AI_OUTPUT_LIMIT, isCreatedResourceSuccess, type AiChatMessage,
+        type AiModelConfig}
   from "@gadgets/workshop-shared/api";
 import {composeCodeChange, type CodeChange} from "@gadgets/workshop-shared/code-change";
 import type {Api, Message, Model} from "@earendil-works/pi-ai";
@@ -201,7 +202,8 @@ export function legacyChatBaseVersion(
  * Earliest turn a checkpoint cannot absorb, or undefined if none. A pending connection request
  * carries live accept/deny state that only its own message can answer, so the boundary stays behind
  * it. Provisional gadget creations and binding additions need no such protection: the checkpoint
- * records them, and the registry rows they name are untouched by compaction.
+ * records them, and the registry rows they name are untouched by compaction. (Nor do pending
+ * creation actions: their decision arrives as a durable agentNudge, an ordinary message.)
  */
 export function findProtectedFromSequence(messages: AiChatMessage[]): number | undefined {
   let protectedIndex = messages.findIndex(
@@ -393,6 +395,11 @@ export function buildCompactionState(
         } else if (call.toolName === "createWorktree" && call.output !== undefined) {
           chatBindings.set(call.input.bindingName,
               {type: "workpiece", id: call.output.worktreeId});
+        } else if (call.toolName === "createExternalResource" &&
+                   isCreatedResourceSuccess(call.output)) {
+          // Mirrors replay's re-establishment of the binding from the recorded output.
+          chatBindings.set(call.input.bindingName,
+              {type: "workpiece", id: call.output.gatekeeperId});
         }
       }
     } else if (message.type === "agentCallback") {
--- a/packages/workshop-backend/src/overseer.ts
+++ b/packages/workshop-backend/src/overseer.ts
@@ -1,6 +1,6 @@
 import { RpcCompatible, RpcStub, RpcTarget } from "capnweb";
 import { validateRpc } from "capnweb-validate";
-import { Overseer, GadgetMetadata, UiBundle, WorkpieceId, WorkpieceSummary, WorkpiecesSubscriber, GadgetClient, GadgetBindingInfo, GatekeeperClient, ActionState, ActionLogEntry, ActionsSubscriber, ActionHistoryFilter, ActionHistoryPage, ChatGadgetPin, ChatCodeBase, ChatGadgetPinState, CodeChangeSubmission, CommitIdentity, CommitInfo, MergeChangesResult, AiChatMetadata, AiChatMessage, AiChatHistoryPage, AiChatSubscriber, AiChatAuthorInfo, AiModelConfig, AiChatMessageBody, AgentSpawnerConfig, ConsoleLogSubscriber, ConsoleLogEvent, CapsuleSpecifier, CollaboratorInfo, CollaboratorRole, AffectedCollaborator, ShareLinkInfo, GatekeeperCreationSpec, ObserverConfigCallback, ObserverBindingNeed, ObserverBindingFailure, BlueprintBindingAnnotation, BlueprintBinding, BlueprintMetadata, BlueprintOutput, MessageFormatRef, isOutputIcon, SpawnerEnvTarget, BlueprintGadgetSummary, AiChatStreamEvent, BlueprintScreenshotUpload, BLUEPRINT_SCREENSHOT_R2_PREFIX, blueprintScreenshotUrl, ChatAttachmentUpload, ChatAttachmentHandle, ChatAttachmentRef, BoundHookInfo, PreApprovableAction, PresenceParticipant, PresenceSubscriber, SlashCommandChoice, SlashCommandRequest, validateBindingName, createOpenGadgetError, OPEN_GADGET_ERROR_CODES, resolveSiteName, actionChangeTime } from '@gadgets/workshop-shared/api';
+import { Overseer, GadgetMetadata, UiBundle, WorkpieceId, WorkpieceSummary, WorkpiecesSubscriber, GadgetClient, GadgetBindingInfo, GatekeeperClient, ActionState, ActionLogEntry, ActionsSubscriber, ActionHistoryFilter, ActionHistoryPage, ChatGadgetPin, ChatCodeBase, ChatGadgetPinState, CodeChangeSubmission, CommitIdentity, CommitInfo, MergeChangesResult, AiChatMetadata, AiChatMessage, AiChatHistoryPage, AiChatSubscriber, AiChatAuthorInfo, AiModelConfig, AiChatMessageBody, AgentSpawnerConfig, ConsoleLogSubscriber, ConsoleLogEvent, CapsuleSpecifier, CollaboratorInfo, CollaboratorRole, AffectedCollaborator, ShareLinkInfo, GatekeeperCreationSpec, ObserverConfigCallback, ObserverBindingNeed, ObserverBindingFailure, BlueprintBindingAnnotation, BlueprintBinding, BlueprintMetadata, BlueprintOutput, MessageFormatRef, isOutputIcon, SpawnerEnvTarget, BlueprintGadgetSummary, AiChatStreamEvent, BlueprintScreenshotUpload, BLUEPRINT_SCREENSHOT_R2_PREFIX, blueprintScreenshotUrl, ChatAttachmentUpload, ChatAttachmentHandle, ChatAttachmentRef, BoundHookInfo, PreApprovableAction, PresenceParticipant, PresenceSubscriber, SlashCommandChoice, SlashCommandRequest, type CreatedResourceOutput, isCreatedResourceSuccess, validateBindingName, createOpenGadgetError, OPEN_GADGET_ERROR_CODES, resolveSiteName, actionChangeTime } from '@gadgets/workshop-shared/api';
 import { applyCodeChange, changedGadgets, codeChangeSerializedSize, composeCodeChange, diffFiles,
   transformCodeChange, validateCodeChangeContent, validateCodeChangeSchema,
   type CodeContent, type CodeChange } from "@gadgets/workshop-shared/code-change";
@@ -30,7 +30,7 @@ import {
   getAiGatewayLogCost,
   type AiGatewayLogRoute,
 } from "./ai-gateway";
-import { AgentGadgetInfo, AgentHooks, AiChatAgentContext, CHAT_CHANGE_MESSAGE_BUDGET, ChatBindingEntry, SeedBindingInfo, runAgent, makeStorableArgs, summarizeArgs, type AgentStepChange, type AiChatMessageBodyWithModelData, type CompactionCheckpoint, type StoredAssistantMessage, type WorktreeTurnAccess } from "./agent";
+import { AgentGadgetInfo, AgentHooks, AiChatAgentContext, CHAT_CHANGE_MESSAGE_BUDGET, ChatBindingEntry, SeedBindingInfo, runAgent, makeStorableArgs, summarizeArgs, type AgentStepChange, type AiChatMessageBodyWithModelData, type CompactionCheckpoint, type CreateExternalResourceInput, type StoredAssistantMessage, type WorktreeTurnAccess } from "./agent";
 import { WorktreeSessionImpl } from "./worktree-session";
 import WORKTREE_BINDING_TYPES from "./worktree-binding.txt";
 import { deploymentOutputForBlueprint, FormatOffer, listFormatOffers, readAdminConfig } from "./admin-config";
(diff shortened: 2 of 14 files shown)


packages/workshop-backend/src/agent-compaction.ts
```diff
@@ -200,12 +201,17 @@ export function legacyChatBaseVersion(
...
+ * registry rows they name are untouched by compaction.
  */
-export function findProtectedFromSequence(messages: AiChatMessage[]): number | undefined {
-  let protectedIndex = messages.findIndex(
-      message => message.type === "connectionRequest" && message.state === "pending");
+export function findProtectedFromSequence(
+    messages: AiChatMessage[], pendingCreationSequence?: number): number | undefined {
+  let protectedIndex = messages.findIndex(message =>
+      (message.type === "connectionRequest" && message.state === "pending") ||
+      (pendingCreationSequence !== undefined && message.sequence >= pendingCreationSequence));
```
ndisidore: Note for reviewers: now a pending resource create also blocks compaction (following the connectionRequest pattern)

packages/workshop-frontend/src/ChatInterface.tsx
```diff
@@ -897,15 +916,15 @@ function buildToolCallGroups(
     const summary = getToolCallSummary(toolCalls[0], outputOf);
     labelParts.push(`${summary.verb}${summary.target ? ` ${summary.target}` : ""}`);
   } else if (toolCalls.length > 1 && distinctToolNames.length === 1) {
-    const summary = getToolCallSummary(toolCalls[0], outputOf);
+    // Label by the last call: for same-target retries the final outcome wins (a failed create
+    // retried successfully is "Created", not "Tried").
+    const summary = getToolCallSummary(toolCalls[toolCalls.length - 1], outputOf);
```
ndisidore, line 921: so `createExternalResource` is the one tool where a "failure" is just a normal string result the model reads and retries on 
every other tool gets away with a dumb count label because the error machinery covers their failures separately. we can't just set `tc.error` on rejections either - the tool call itself succeeded, it returned guidance the model is supposed to act on, and flagging that as an error would paint the designed fix-and-retry flow red like something broke. 
so if we did "created 2 external resources" when one was actually rejected, the group header would straight up lie about the thing the user is being asked to approve - hence this one case counts successes

open to other options here

packages/workshop-backend/src/overseer.ts
```diff
@@ -2661,6 +2736,26 @@ class OverseerImpl implements AgentHooks {
...
+    // Undecided creations barriered to this chat outlive the pending marker, and once the log
+    // is deleted an approvable card would mint a provider resource nothing can address -- so
+    // settle them, unless a surviving gadget edge still binds the resource (merged or another
+    // chat's; this chat's pending edges were severed above), which keeps the card meaningful.
+    // Approved creations stay -- the provider resource is real.
+    for (let record of Array.from(this.storage.gatekeepers.list())) {
+      if (record.creation?.chatId !== chatId) continue;
+      let bound = [...this.storage.gadgets.list()].some(gadget =>
+          gadget.type === "gadget" &&
+          Object.values(gadget.bindings).some(edge => edge.target === record.id));
```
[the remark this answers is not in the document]
ndisidore: This is not unique to this code:  dangling outcome is the designed, platform-wide degradation

packages/workshop-backend/src/overseer.ts
```diff
@@ -5255,6 +5371,94 @@ class OverseerImpl implements AgentHooks {
...
     });
+
+    // The decision nudge precedes the best-effort refresh below: a reset during describe() must
+    // not cost the model the verdict -- the recorded tool result permanently says the resource
+    // doesn't exist yet, and nothing later re-emits the decision.
+    let gatekeeperRecord = this.storage.gatekeepers.get(record.gatekeeperId);
+    let creationId = gatekeeperRecord?.creation?.actionId;
+    if (record.id === creationId && record.caller.from === "agent" &&
+        gatekeeperRecord !== undefined) {
+      this.nudgeCreationDecision(record.caller.chatId, gatekeeperRecord, "approved", resolvedBy);
```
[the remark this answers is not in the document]
ndisidore: I don't think this is accurate: under the DO's write-coalescing guarantee they commit as one atomic batch.

packages/workshop-backend/src/overseer.ts
```diff
@@ -5255,6 +5371,103 @@ class OverseerImpl implements AgentHooks {
...
+          if (gatekeeperRecord.pending === undefined) {
+            this.#restartIfSessionsAffected(
+                "Gadget restarted because a created resource now exists at the provider.");
+          }
+        }
+      } catch (error) {
+        this.logger.warn("failed to refresh created resource description after apply", {
+          event: "gatekeeper.created.describe.refresh.failed",
+          gatekeeperId: record.gatekeeperId, error,
+        });
```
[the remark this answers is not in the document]
ndisidore: Eh, this is documented best-effort with a bounded cosmetic residual. it requires `describe()` failing twice, milliseconds after `applyAction` to the same vendor succeeded

packages/workshop-backend/src/overseer.ts
```diff
@@ -11227,19 +11658,49 @@ class OverseerClientInterface extends RpcTarget implements Overseer {
...
+    // nudge land in one synchronous step -- a crash mid-cascade must not strand half of it --
+    // ahead of the best-effort gatekeeper notifies. The record itself survives: replay
+    // re-establishes the chat binding from the recorded tool call (see agent.ts), and the
+    // vendor's dead-session error explains the rejection better than a missing binding would.
+    let record = this.impl.storage.gatekeepers.get(action.gatekeeperId);
+    if (record?.creation?.actionId === id) {
+      let siblings = Array.from(
+          this.impl.storage.actions.pendingByGatekeeper.get(action.gatekeeperId))
+          .filter(sibling => sibling.type === "action");
+      for (let sibling of siblings) this.#markActionRejected(sibling, profile);
```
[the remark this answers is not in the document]
ndisidore: accepted tradeoff. the outcome is the platform's uniform broken-connection behavior: the gadget's calls fail with the vendor's curated `DEAD: … do not retry` explanation

packages/workshop-backend/src/overseer.ts
```diff
@@ -8835,6 +9160,102 @@ class OverseerImpl implements AgentHooks {
...
   }
 
+  // Create a brand-new external resource (createExternalResource tool). Unlike requestConnection,
+  // no user action gates the binding: the gatekeeper simulates the resource locally, and the
+  // provider-side creation is an ordinary pending action (captured for this chat, so its card
+  // lands in the transcript at the step barrier). `created: false` is a fixable rejection — the
+  // agent should adjust and retry in the same turn.
+  async createExternalResource(chatId: number, input: CreateExternalResourceInput,
+      initiator: AiChatAuthorInfo): Promise<CreatedResourceOutput | string> {
+    let vendors = await this.#listGatekeeperVendorsCached();
```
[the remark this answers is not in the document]
ndisidore: Right now catalogs don't vary per user so this point is moot.

packages/workshop-backend/src/overseer.ts
```diff
@@ -5349,7 +5576,16 @@ class OverseerImpl implements AgentHooks {
...
     // interleave between the record appearing and the block taking effect.
     if (creationSpec && "vendorId" in creationSpec) {
-      if (this.#restartIfSessionsAffected(
+      if (creation) {
+        // Deferred: an immediate abort would kill the minting turn before its barrier records
+        // the tool call, and the resumed replay would re-issue the mint -- an abort/resume loop
+        // for as long as the collaborator keeps reconnecting. Quarantine in the same synchronous
+        // block instead (unverified sessions can't reach the id, see assertGatekeeperUsable) and
+        // restart once the log backs the creation (addChatMessages' unstamp branch).
+        if (this.#hasCollaboratorSession("build")) {
```
ndisidore, line 5585: The comment above has more context on this.

the restart is a full DO abort, but the tool call that recorded this mint isn't durable until the step barrier commits - abort before that and the workspace comes back with no memory of the mint, the resumed turn re-creates it, and you're in a remint loop. so instead in these cases we defer the restart until the call is actually in the log

no collaborator session, no quarantine, no restart - all this gets skipped

packages/workshop-shared/src/api.ts
```diff
@@ -3233,8 +3233,59 @@ export type AiToolCall = {
...
+     * Vendor-specific creation parameters (flat scalars), as documented by the creatable
+     * type's description; see ResourceCreationOptions.
+     */
+    options?: ResourceCreationOptions;
+  };
+
+  /**
+   * On success, the full tool result: the created gatekeeper workpiece, its provisional
+   * resourceUrl, and the message shown to the model — replay re-binds and re-renders from it
+   * instead of re-creating, like createGadget. A string output is a fixable rejection returned
```
kentonv, line 3269: Shouldn't a rejection use the common `error` property instead?

(I'm not sure why `requestConnection` does it differently -- maybe it should also change to use `error`.)

packages/workshop-shared/src/api.ts
```diff
@@ -3233,8 +3233,59 @@ export type AiToolCall = {
...
+/** Success output of a createExternalResource call (see the AiToolCall member). */
+export type CreatedResourceOutput = {gatekeeperId: WorkpieceId, resourceUrl: string, message: string};
+
+/**
+ * True iff a createExternalResource output is the structured success shape — the only shape under
+ * which a binding was made. A string output is a fixable rejection; it is not an error (no
+ * `error` is set on the call), so error-based filters do not catch it. Every consumer that scans
+ * recorded tool calls must gate on this, not on `output !== undefined`.
+ */
+export function isCreatedResourceSuccess(output: CreatedResourceOutput | string | undefined)
```
kentonv, line 3284: I'm dubious this helper is worth defining, but it could definitely go away if we change to use `error` for errors (then `output` would be `output?: CreatedResourceOutput` so could trivially be tested with `if (tool.output)`).

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -593,6 +614,32 @@ export interface GatekeeperUser extends WorkerEntrypoint {
...
+   * Throws with an agent-readable message when the account cannot create this resource type
+   * (e.g. its authorization does not cover the needed scopes) or when `input.options` carries
+   * unknown or invalid entries (see ResourceCreationOptions); callers surface the message.
+   */
+  createResource?(resourceUrlPattern: string,
+                  input: {title: string, options?: ResourceCreationOptions}): Promise<{
+    class: DurableObjectClass<Gatekeeper<any>>;
+    resource: SupportedResource;
+    /** Provisional URL of the new resource; replaced by the real URL once created. */
+    resourceUrl: string;
```
kentonv, line 640: What is the use of a provisional URL? It's fake, you can't actually visit it, and in most cases it won't even match the final URL. Could we just drop this property altogether?

packages/workshop-backend/src/overseer.ts
```diff
@@ -8835,6 +9160,102 @@ class OverseerImpl implements AgentHooks {
...
+    // Mint the provisional gatekeeper class through the *initiator's* user DO (the admin-check
+    // chokepoint) -- connected accounts are per-user, so a collaborator-driven turn creates the
+    // resource under (and enumerates) the collaborator's accounts, not the owner's. The same
+    // initiator.id resolution as listAvailableBlueprints. Its failures are agent-readable by
+    // contract: no usable account, ambiguous accounts, missing authorization.
+    let minted;
+    try {
+      let userStub = wrapDoStubForTelemetry(
+          this.users.get(this.users.idFromName(initiator.id)), this.logger);
+      minted = await userStub.createResourceGatekeeper(
```
ndisidore, line 9197: Flagging again for reviewers: this goes through the *turn initiator’s User DO* on purpose, since connected accounts are per-user. a collaborator-driven chat creates the resource with that collaborator’s selected account, while gadget/scheduled turns resolve the initiator to the gadget owner, so the normal scheduled-task case still uses the owner’s account. the returned gatekeeper class is permanently imbued with that account’s credentials - whoever later clicks approve does not swap in their own credentials

I'm not deadset on this approach so I welcome pushback here

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -593,6 +614,32 @@ export interface GatekeeperUser extends WorkerEntrypoint {
...
+   * the resource locally, and describe() must not call the provider. addObserver() likewise
+   * MUST NOT require the provider resource to exist: build collaborators are verified against
+   * the binding while the creation is still pending, so admit observers on the simulated
+   * state's own policy.
+   *
+   * Throws with an agent-readable message when the account cannot create this resource type
+   * (e.g. its authorization does not cover the needed scopes) or when `input.options` carries
+   * unknown or invalid entries (see ResourceCreationOptions); callers surface the message.
+   */
+  createResource?(resourceUrlPattern: string,
```
ndisidore, line 635: this method is intentionally not the provider-side create call - it only mints a provisional local capability. the real create is queued as the first normal approval action, while the gatekeeper simulates the resource so the agent can keep working and queue dependent edits. that’s what lets something like “create today’s doc, then append to it” happen in one turn without bypassing approval. once that first action is approved, the same gatekeeper starts talking to the real resource instead

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -738,6 +785,24 @@ export interface Gatekeeper<Session> extends DurableObject {
...
+   * Ordering is the gatekeeper's responsibility: the platform applies auto-approvals in order,
+   * but a manual approval can target any pending action, so the gatekeeper must itself reject
+   * applyAction() of any action that depends on the resource existing until the creation has
+   * been applied. Only gatekeepers reachable via createResource() need implement this.
+   *
+   * If this call fails, or a crash orphans the mint, the platform settles the queued actions
+   * and removes the gatekeeper WITHOUT delivering rejectAction(): the facet's storage is
+   * destroyed with it, and any state staged outside the facet must tolerate orphaned entries.
+   */
+  submitCreationAction?(approvalQueue: RpcStub<ApprovalQueue>): Promise<void>;
```
kentonv, line 804: > &lt;thinking&gt;
> Hmm, as currently designed, `ResourceCreationOptions` would have to be embedded into the props of the gatekeeper class returned by `createResource()`. This strikes me as a bit odd -- those props are permanent, but creation options may not be needed anymore after creation time. Perhaps the options should be passed here instead?
>
> I'm also realizing that in the existing design, `getGatekeeperClassFor()` needs to embed the resource URL into the props. This is problematic because a gatekeeper created by `createResource()` doesn't necessarily have a URL until finalized, but we would probably like to share the same implementation class between created resources and existing ones... but you can't change the props after the fact.
>
> I wonder if we need to transition the gatekeeper API to be two-step for all gatekeepers, that is, you obtain the class you want first, then you start the facet and call a method on the class to initialize it. When connecting to an existing resource, you'd pass the URL on that method. For creations, you'd pass the ResourceCreationOptions.
>
> Existing gatekeepers that embed their URL into `props` would need a migration to move the information into local storage instead.
>
> Alternatively, I guess another approach we could take is to keep connecting to existing resources as-is, but for creations, the `props` starts out containing null values, but when the creation is finalized, there's some API for the facet to return a replacement `DurableObjectClass` to use from that point forward. So that way it actually is able to embed the URL into the `props` after the fact.
>
> OK, I guess I like this idea a bit better because it allows read-only gatekeepers to remain totally stateless. (Whereas if we move props to dynamic storage, the facet MUST initialize its sqlite database to store the resource URL at least...)
> &lt;/thinking&gt;

OK, I think what we actually want here is:
* Remove this method. I think it's not really needed: The `createResource()` call could return the appropriate ActionDescription.
* Add a different method, `applyCreation()`, to be called to _actually create_ the resource, instead of having `applyAction()` do it. This new method should return a new version of the `DurableObjectClass` (and final `resourceUrl`), and the facet should then be restarted using that new class before anything else happens.

I kind of wonder if `applyCreation()` should be hooked with an opportunity for the user to be presented with some UI to choose final creation options. E.g. for GitHub, this could give the user a chance to change the repo name and select an org to put it in -- things the agent might not have known if it was given generic instructions to create a repo.

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -556,6 +568,15 @@ export interface GatekeeperConnectCallback extends WorkerEntrypoint {
...
 }
 
+/**
+ * Vendor-specific creation parameters for GatekeeperUser.createResource(), authored by the agent
+ * and untrusted like `title`. Flat bounded scalars only (e.g. a parent folder id). The resource's
+ * `creatable.description` documents the accepted keys; the vendor MUST reject unknown or invalid
+ * entries with an agent-readable message, and MUST reflect consequential entries (e.g. placement)
+ * in the creation action's description so the user approves what will actually happen.
+ */
+export type ResourceCreationOptions = Record<string, string | number | boolean>;
```
kentonv, line 578: How does the agent discover what options are available?

packages/workshop-shared/src/gatekeeper.ts
kentonv, line 635: Maybe worth being extra-clear in the comments that this method does not actually create anything or have any side effects. It just returns a class. You have to call methods on the facet for anything to happen.

PR #440, opened 2026-09-03
Add a unified Google Drive Folder resource

ndisidore:
Replaces the separate Shared Drive resource with one Google Drive Folder resource, read-only.
One selection covers an ordinary folder in My Drive, a folder someone else shared with you,
or a subfolder inside a shared drive.

A folder binding sees its direct children only, not everything beneath it: listing and
provider-side search are scoped to that one folder, and a child folder is reached by opening
it as its own capability rather than by recursive traversal.

It requests the same three read-only scopes as the account and exact-file resources. The picker
is one `corpora=allDrives` search, which finds every folder kind above without an account-wide
grant. It returns one provider page of suggestions, not a complete list, so a shared drive's own
root or a folder that page misses is still connectable by its URL, which prefills the form. An
incomplete search from Drive fails rather than looking complete.

Folder-derived Doc and Sheet sessions re-prove ancestry on every method call, not only
at open, so a file moved out of the folder stops answering through a session that is
already open. Drive has no ancestry-plus-content transaction, so a move landing after
the final check still returns and the next read denies.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/440" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


ndisidore:
@ask-bonk review this

maxwellpeterson (approved):

PR #441, opened 2026-09-03
feat(gatekeeper-kit): replayable runs with account-adjudicated expiry

ndisidore:
`CredentialSource.run` gains a `replayable` option: when the provider rejects credentials, the account decides whether they are `expired`, `superseded`, or `unavailable`, and a replayable operation may retry once with a same-connection successor without crossing a reconnect.

This is an improved version of `fetchWithAuthRetry` that many gatekeepers hand roll

Derived-bearer ports can heal a rejected current credential inside `reportCredentialsRejected`, with identity-keyed single-flight coalescing concurrent mints; grant-death ports instead notify the Workshop and return expired.

Each attempt receives its own `{ identity, generation }` fence, while named `CredentialsChangedError` and `CredentialsExpiredError` results survive RPC boundaries and stale reads, reports, and refetches cannot re-establish cache authority.


ndisidore:
> P2 packages/gatekeeper-kit/src/credentials.ts:341: A reconnect during notify() leaves snapshot() rethrowing the old grant’s expiry despite new credentials being current. Recheck identity after notification.

Not worth actioning. The window is the notify RPC only; fresh()'s #overtaken already covers the wide mint-await window. The delta is error flavor for one caller — it sees expired instead of the reconnect that landed mid-notify — and its re-entry refetches the live credentials. No false notify is created (the grant was confirmed dead when notify fired; the latch handles dedup).

> P2 packages/gatekeeper-kit/src/credentials.ts:635: A reconnect adopted while the rejection report is pending is unconditionally cleared; a failed report then falsely expires the live connection. Reconcile the verdict with any adopted successor before clearing authority.

The scenario needs a triple coincidence: 401 + a concurrent fetch adopting a successor inside the ask window + the report reply lost (fail-closed "expired"). Net damage: the one reporting caller gets CredentialsExpiredError instead of CredentialsChangedError. Consider this accepted

PR #450, opened 2026-09-04
Support multi-tab Google Docs

ndisidore:
Updates the Docs session to support tabbed Docs. What's a tab in a doc, you ask? 

image
These things

One document can hold a tree of tabs, each with its own body and its own character indices. We were asking Google for all of it and then erroring out if there was more than one tab, which blocked the whole document, including read-only Docs opened through Drive.

Now `listTabs()` returns that tree flattened depth-first, and `getContent()`, `replaceText()` and `appendText()` each take an optional `tabId`. Every operation reads or edits exactly one tab; we never merge them. You can leave `tabId` off only when the doc has exactly one tab, and an ID we don't recognize errors instead of quietly falling back to the first tab. Since each tab has its own index space, every write coordinate we send now carries the tab ID.

This avoids being a breaking api change while also not introducing weird markers in the markdown.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/450" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/packages/gatekeeper-google/src/docs-read-types.d.ts
+++ b/packages/gatekeeper-google/src/docs-read-types.d.ts
@@ -7,11 +7,47 @@ export type DocMetadata = {
   lastModified: Date;
 }
 
-/** Read-only access to one native Google Doc. */
+/**
+ * One tab of a native Google Doc.
+ *
+ * A document is a tree of tabs; `listTabs()` returns that tree flattened depth-first, parents
+ * before children, so `parentTabId` reconstructs the hierarchy without nested objects.
+ */
+export type GoogleDocTab = {
+  /** Immutable tab ID. Pass it to `getContent()` and to edits to target this tab. */
+  id: string;
+
+  /** Tab name, as shown in the document's tab list. */
+  title: string;
+
+  /** The tab that contains this one, absent for a top-level tab. */
+  parentTabId?: string;
+
+  /** Position among the tabs sharing this parent, counting from 0. */
+  index: number;
+
+  /** Depth in the tab tree: 0 for a top-level tab, 1 for its child, and so on. */
+  nestingLevel: number;
+}
+
+/**
+ * Read-only access to one native Google Doc.
+ *
+ * Each tab holds its own content. Call `listTabs()` before `getContent()` to see which tabs
+ * exist, then read one of them.
+ */
 export interface GoogleDocReadSession {
   /** Return current document metadata. Works with any number of tabs. */
   getMetadata(): Promise<DocMetadata>;
 
-  /** Return the document body as Markdown. Requires exactly one tab. */
-  getContent(): Promise<string>;
+  /** Return every tab in the document, flattened depth-first with parents before children. */
+  listTabs(): Promise<GoogleDocTab[]>;
+
+  /**
+   * Return one tab's content as Markdown.
+   *
+   * This reads exactly one tab and never combines tabs, so pass an ID returned by `listTabs()`.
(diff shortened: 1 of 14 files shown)


packages/gatekeeper-google/src/docs-read-types.d.ts
```diff
@@ -7,11 +7,47 @@ export type DocMetadata = {
...
+  /** Return every tab in the document, flattened depth-first with parents before children. */
+  listTabs(): Promise<GoogleDocTab[]>;
+
+  /**
+   * Return one tab's content as Markdown.
+   *
+   * This reads exactly one tab and never combines tabs, so pass an ID returned by `listTabs()`.
+   * Omitting `tabId` is valid only when `listTabs()` returns exactly one tab.
+   */
+  getContent(tabId?: string): Promise<string>;
```
ndisidore, line 52: This has been around for half a year so trying to avoid a breaking change here. My thought is tabs are probably less common (at least I've never used them) so I'm less inclined to optimize for that path.

The alternatives here could be
- adding a new `getTabbedContent()` method that returns a structured object but we would still need the agent to do a `hasTabs` or `listTabs` call.
- having special "markers" in the markdown string around tabs - this is prob my least favorite

Personally I don't really mind much either way

packages/gatekeeper-google/src/google.ts
```diff
@@ -1281,58 +1367,106 @@ function invalidateGoogleDocAction(
...
 
-    case "appendText": {
-      let insertAt = snapshot.bodyEndIndex - 1;
-      return markdownToDocRequests("\n" + action.markdown, insertAt);
-    }
+    case "appendText":
+      return {
+        tab,
+        requests: markdownToDocRequests(
+            "\n" + action.markdown, tab.bodyEndIndex - 1, tab.tabId),
```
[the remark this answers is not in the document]
ndisidore: This is byte identical to what's on main and also a real docs body cannot be empty (it always has a section break)

packages/gatekeeper-google/src/google.ts
```diff
@@ -1748,28 +1891,61 @@ class GoogleDocSessionImpl extends RpcTarget implements GoogleDocSession {
...
+    await this.#approvalQueue.authorizeObservation({
+      title: "Read Google Doc content",
+      description:
+        `Read the full simulated content of tab ${googleDocTabLabel(selected.tab)} as Markdown.`,
+    });
+    return selected.markdown;
   }
 
-  async replaceText(oldMarkdown: string, newMarkdown: string): Promise<void> {
+  async replaceText(oldMarkdown: string, newMarkdown: string, tabId?: string): Promise<void> {
```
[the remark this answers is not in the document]
ndisidore: This is also pre-existing on main. Deferred.

packages/gatekeeper-google/src/docs-read-types.d.ts
maxwellpeterson: Makes sense to me! If anything it could be looser so calling `getContent()` without `tabId` always returns content from the first tab (regardless of how many tabs there are), but this feels like a good middle ground.

maxwellpeterson (approved):

packages/gatekeeper-google/src/docs-read-types.d.ts
ndisidore: Agreed! We could always depreciate the tab Id param later and add a new `getTabbedContent` in the future if we decide it is worth the lift

PR #451, opened 2026-09-05
fix: preserve custom resource URL schemes

### LINKED ISSUE #390: normalizeResourceUrl mangles custom-scheme resource URLs (gitlab://x → https://gitlab://x)
## Description

`normalizeResourceUrl()` in `workshop-frontend/src/resourceMatching.ts` prepends `https://` to any resource URL that does not already start with `http(s)://`:

```ts
if (normalized && !/^https?:\/\//i.test(normalized)) {
  normalized = 'https://' + normalized
}
```

Custom-vendor resources with their own schemes are first-class elsewhere in the platform — a blueprint's `gatekeeperName` + `typeUrlPattern` may reference any `GATEKEEPER_*` vendor, and `matchesResourceUrlPattern()` deliberately handles non-special schemes — but any custom-scheme URL passing through the blueprint flow gets mangled on the way to the backend: our `gitlab://inspekter-estate` reaches `account.getGatekeeperClassFor()` as `https://gitlab://inspekter-estate` (both via the auto-suggestion path and via `handleSaveActiveBinding`), and the UI displays the mangled form ("Using: https://gitlab://inspekter-estate").

Observed on our pinned build (`6478a144`) and still reproducing as of 2026-08-30; the regex is unchanged on current `main`.

## Steps observed

1. Deploy a custom gatekeeper whose vendor advertises a custom-scheme resource, e.g. `getSupportedResources()` → `{ urlPattern: "gitlab://*" }`.
2. Build a blueprint with a binding declaring `typeUrlPattern: "gitlab://*"`, `resourceUrl: "gitlab://inspekter-estate"`.
3. Import the blueprint and save the binding.
4. The vendor's `getGatekeeperClassFor()` receives `https://gitlab://inspekter-estate`; without a vendor-side workaround the bind fails (in our case: `This gatekeeper serves gitlab:// resources, not https://gitlab://inspekter-estate`), and the UI shows the mangled URL.

## Expected behavior

Prepend `https://` only when the URL has no scheme at all — e.g. test `/^[a-z][a-z0-9+.-]*:\/\//i` instead of `/^https?:\/\//i`.

## Workaround we use

Our vendors' `getGatekeeperClassFor()` strips a leading `https://` when it fronts the vendor's own scheme. That makes binding work, but the mangled form still persists in stored specs and in the UI, so it's tolerance rather than a fix.

Since the change looks like a one-line regex swap, we're happy to send it as a small PR per CONTRIBUTING.md if that's the direction you'd want — otherwise this is just the report.

atirna:
## What does this change?

`normalizeResourceUrl()` now preserves resource URLs with a custom `scheme://` prefix instead of prepending `https://`. Bare resource hosts still receive the existing HTTPS default, and trailing `/*` is still removed.

Fixes #390

The one shared normalizer now recognizes the URL scheme form already used by custom resource patterns. Its existing default only applies when that prefix is absent. The focused test covers both the reported `gitlab://` path and the bare-host default.

## Checklist

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/451" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/packages/workshop-frontend/src/resourceMatching.test.ts
+++ b/packages/workshop-frontend/src/resourceMatching.test.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest'
+import { normalizeResourceUrl } from './resourceMatching'
+
+describe('resource URL normalization', () => {
+  it('preserves custom schemes while supplying HTTPS for bare hosts', () => {
+    expect(normalizeResourceUrl('gitlab://inspekter-estate/*')).toBe('gitlab://inspekter-estate')
+    expect(normalizeResourceUrl('inspekter-estate/*')).toBe('https://inspekter-estate')
+  })
+})
--- a/packages/workshop-frontend/src/resourceMatching.ts
+++ b/packages/workshop-frontend/src/resourceMatching.ts
@@ -208,13 +208,13 @@ export function matchesResource(search: string, resource: SupportedResource): bo
 }
 
 /**
- * Normalize a resource URL for sending to the backend. Prepends https:// if no protocol
+ * Normalize a resource URL for sending to the backend. Prepends https:// if no scheme
  * is present, and strips a trailing /* wildcard (which is just "everything under here").
  */
 export function normalizeResourceUrl(url: string): string {
   let normalized = url.trim()
-  // Default to https:// if no protocol specified.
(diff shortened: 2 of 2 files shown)


atirna:

PR #461, opened 2026-09-08
proposal: gatekeeper resource sharing API

maxwellpeterson:
This PR sketches out an extension of the gatekeeper API that supports gatekeeper resource sharing between Cloudflare OS users. The initial use case for this API is supporting collection sharing within the context gatekeeper.

### DIFF
--- a/packages/workshop-shared/src/gatekeeper.ts
+++ b/packages/workshop-shared/src/gatekeeper.ts
@@ -80,11 +80,77 @@ export type VendorDescription = {
 }
 
 /**
- * Per-open context the Workshop passes to GatekeeperUser.startAppUi(). `isAdmin` is supplied fresh
- * each time rather than baked into the account, since a user's admin status can change over time.
+ * User directory capability passed from the workshop backend to a gatekeeper to
+ * faciliate sharing.
+ */
+export interface GatekeeperSharingDirectory extends RpcTarget {
+  /**
+   * Returns active workshop users that a gatekeeper resource can be shared
+   * with.
+   */
+  searchRecipients(query: string): Promise<GatekeeperShareRecipientSearchResult[]>;
+}
+
+/**
+ * Potential sharing recipient with presentation metadata and a delivery
+ * capability.
+ */
+export type GatekeeperShareRecipientSearchResult = {
+  /** Presentation metadata used for rendering the gatekeeper sharing UI. */
+  display: ShareRecipientDisplayInfo;
+  /** Capability for delivering a sharing grant to this recipient. */
+  recipient: RpcStub<GatekeeperShareRecipient>;
+};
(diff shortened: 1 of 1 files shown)


packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -80,11 +80,77 @@ export type VendorDescription = {
...
+  avatar?: AvatarImage;
+};
+
+/** Sharing capability bound to one user and one gatekeeper vendor. */
+export interface GatekeeperShareRecipient extends RpcTarget {
+  /**
+   * Routes the grant to the bound recipient. After the account accepts the
+   * grant, return a persistent Workshop-owned capability for reading the
+   * recipient's current presentation metadata. If the recipient does not have a
+   * corresponding gatekeeper account, the share fails.
```
maxwellpeterson, line 120: Instead of failing at delivery time, the user directory capability passed to the gatekeeper could omit users that don't meet this prerequisite. This makes the directory implementation more complicated, but is probably better UX here.

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -180,6 +246,12 @@ export type AccountDescription = {
    * surfaces it as a nav entry / page using this title.
    */
   providesUi?: { title: string; icon?: AvatarImage };
+
+  /**
+   * If true, this account implements GatekeeperUser.acceptShare(), and the Workshop may provide
+   * its management UI with a sharing directory.
```
ndisidore, line 252: nit: this may be context library specifics leaking into this contract?
is `directory` portable?  `supportsSharing` should advertise implementation support, not authorize directory access

maxwellpeterson: "directory" refers user directory here, not the filesystem type of directory. We might want to call the user directory something else to avoid confusion.

packages/workshop-shared/src/gatekeeper.ts
```diff
@@ -80,11 +80,77 @@ export type VendorDescription = {
...
- * each time rather than baked into the account, since a user's admin status can change over time.
+ * User directory capability passed from the workshop backend to a gatekeeper to
+ * faciliate sharing.
+ */
+export interface GatekeeperSharingDirectory extends RpcTarget {
+  /**
+   * Returns active workshop users that a gatekeeper resource can be shared
+   * with.
+   */
+  searchRecipients(query: string): Promise<GatekeeperShareRecipientSearchResult[]>;
```
kentonv: Could this return a `GatekeeperUserVerifier`, which the gatekeeper can then unwrap itself? It could then implement the sharing internally -- no need to spec out a `deliver()` RPC.

PR #468, opened 2026-09-09
Preserve Cloudflare connections and credential updates across sign-in

Ankcorn:
### Problems

1. **Signing in again could replace the more powerful connection with the basic one**, breaking features you had already connected.
2. **Later credential changes weren’t passed along**, so Cloudflare OS could miss that the connection had expired or been restored.

### Fix

Keep existing connections with resource permissions when users sign in again. Billing-only connections still refresh during sign-in; connections with resource permissions use the existing reconnect flow, including when their credentials have expired.

Forward credential expiry and restoration updates to the connected account after login finishes, so its status stays up to date.

Includes regression tests for repeat sign-in and credential updates. Lint/build, 775 backend tests, and 3 integration tests pass (4 existing integration tests skipped).


Ankcorn:

PR #479, opened 2026-09-11
implement kernel changes required for context collection sharing

maxwellpeterson:
This PR adds a generic user selection API that gatekeepers can use to support resource sharing and other features that require interacting with workshop user accounts. The initial use case for this API is supporting collection sharing in the context gatekeeper. The end to end flow for the context gatekeeper would look like this:

1. The iframed gatekeeper app UI calls `GatekeeperAppHost.selectUsers(collectionId)`. This request is delivered to the workshop frontend over the MessagePort RPC session. `collectionId` is an opaque string that is passed back to the gatekeeper backend later in the process to identify the context in which a user was selected. If the Workers runtime supported RPC stub unwrapping, this string could be replaced by an opaque stub that was unwrapped by the gatekeeper backend instead.
2. The workshop frontend opens the user picker modal over the iframed gatekeeper app UI. The user picker modal is owned by the workshop frontend and runs outside of the iframe. For Cloudflare OS deployments with user search enabled, the user picker modal supports searching for other users on the platform. If user search is disabled, users can be selected by entering their exact username or email.
3. When a user is selected, the workshop frontend calls `AuthenticatedApi.selectGatekeeperUser(contextGatekeeperId, selectedUserId, collectionId)` which calls `UserDurableObject.deliverSelectedUser(accountId, selectedUserId, collectionId)` on the DO of the user who made the selection.
4. The user DO creates a `GatekeeperUserVerifier` for the selected user and passes that to the gatekeeper backend by calling `receiveSelectedUser(collectionId, selectedUserVerifier, profile)` on the gatekeeper app account responsible for the UI. The gatekeeper backend would then verify that the user can share `collectionId`, and use `selectedUserVerifier` to deliver the share to the selected user. `profile` is a read-only capability that the gatekeeper can use to display profile information about the selected user in the gatekeeper app UI.
5. Returning to step 1, `GatekeeperAppHost.selectUsers()` returns `Promise<void>`. This promise resolves when the user closes the user picker modal, and indicates that the gatekeeper app UI should refresh to pull in any server-side state that was updated while the modal was open.

### Caveats

User selection eligibility isn't checked until the user is selected from search results, so users that do not have a corresponding context gatekeeper account will still appear in search results and fail with a helpful error message when selected. This lets us re-use the same user search API used for workspace sharing. If we wanted to scope search results to just eligible users, the user directory would need to become aware of each user's connected accounts and connection state, which I'm not sure we want.


packages/workshop-backend/src/user.ts
```diff
@@ -1971,6 +1997,27 @@ export class UserDurableObject extends DurableObject<Cloudflare.Env> {
...
+    let config = await readAdminConfig(this.env);
+    if (config.disabledGatekeepers.includes(vendorId)) return null;
+
+    await this.#ensureAutoProvisionedAccounts();
+
+    let matches = [...this.#connectedAccountRecords()].filter((record) =>
+      record.vendorId === vendorId && areCredentialsValid(record) &&
+      (!record.autoProvisioned || ambientGatekeeperMode(config, vendorId) !== "disabled")
+    );
+    return matches.length === 1 ? await matches[0].account.getVerifier() : null;
```
maxwellpeterson, line 2018: In the case of multiple accounts, we could also return multiple matches and let the the user pick between them. This would expose the existence of those accounts to the user doing the picking, which we might not want.

For the initial context gatekeeper use case, users will never have multiple accounts, so we could defer this until there's a concrete need to handle multiple accounts here. This is all self-contained in the workshop, so adding support for multiple accounts in the future would not require changing the gatekeeper-facing API.

maxwellpeterson: Alternatively, we could also pass multiple accounts to the gatekeeper and let it decide what to do with them.

kentonv:
> 3. When a user is selected, the workshop frontend calls `AuthenticatedApi.selectGatekeeperUser(contextGatekeeperId, selectedUserId, collectionId)` which calls `UserDurableObject.deliverSelectedUser(accountId, selectedUserId, collectionId)` on the DO of the user who made the selection.
> 4. The user DO creates a `GatekeeperUserVerifier` for the selected user and passes that to the gatekeeper backend by calling `receiveSelectedUser(collectionId, selectedUserVerifier, profile)` on the gatekeeper app account responsible for the UI. The gatekeeper backend would then verify that the user can share `collectionId`, and use `selectedUserVerifier` to deliver the share to the selected user. `profile` is a read-only capability that the gatekeeper can use to display profile information about the selected user in the gatekeeper app UI.
> 5. Returning to step 1, `GatekeeperAppHost.selectUsers()` returns `Promise<void>`. This promise resolves when the user closes the user picker modal, and indicates that the gatekeeper app UI should refresh to pull in any server-side state that was updated while the modal was open.

This seems a bit round-about. This seems to mean gatekeeper needs to keep state server-side related to a UI flow that is ongoing on the client. E.g. the client-side UI may be a form that isn't complete yet and isn't intended to have any effect until the user clicks "sumbit". Now there needs to be a server-side DO holding the form state, so that it can be updated when `receiveSelectedUser()` is invoked (`collectionId` I guess points at this DO).

How about this alternative: When the user selects a contact, some sort of token or ticket is created and returned as the result of the original `selectUsers()` call. The gatekeeper app can then send this ticket to the server (perhaps not until later, when the user clicks "submit"). At that time, the ticket can be exchanged for a GatekeeperUserVerifier server-side.

maxwellpeterson:
@kentonv I was modeling the UX here on the current workspace sharing modal, where users are added as soon as they are selected and there is no "submit" action, you just close the modal when you're done adding people. The API you're describing makes sense, but would mean the gatekeeper user picker has different UX from the workspace user picker. If we implemented the ticket-based API I think we should also update the workspace user picker to follow the same deferred submission pattern. I don't feel strongly about either approach, but I think we should be consistent across both of these. What do you think?

PR #489, opened 2026-09-13
Updated agent prompt: avoid unnecessary implementation details in response, don't always create gadgets

jonesphillip:
Noticing a couple of things in agent behavior that can be improved: 
- Responses often include phrases like like "binding" or "X (bindings)'s API". Those are implementation details that describe why agent is calling tools. We don't need the agent to describe implementation details to users unless prompted to do so.
- We are a little too quick to create gadgets. If a user wants help writing an email, we should respond in text instead of building gadget.

The goal of this PR is to make the agent communicate clearly for nontechnical users, avoid unnecessary implementation details, and use better judgment about when to create a gadget.

I ran 14 scenarios across Sonnet 4.6, GPT-5.6 Sol, and GLM 5.3, comparing the original and updated prompts (summary below). Overall saw improvement. Specifically saw fewer unnecessary gadgets and more direct answers in chat where warranted. Also saw less unprompted narration about bindings and APIs describing how agent was getting answers for users.

| Scenario | Result | What changed |
|---|---|---|
| Simple greeting | Improved | All three models gave a neutral greeting without steering users toward building gadgets. |
| Summarize supplied information | Baseline | — |
| One-off calculation | Baseline | — |
| Read a connected spreadsheet | Improved | Sonnet removed setup narration and delivered the analysis directly. |
| Request document access | Baseline | — |
| Explicit gadget creation | Improved | GLM described useful features without explaining storage or callbacks. |
| Update an existing gadget | Baseline | — |
| Technical details when requested | Baseline | — |
| Vague request for expense help | Improved | All three avoided creating an app immediately. |
| Spawned agent reporting results | Improved | Sonnet led with the result and removed narration about bindings and APIs. |
| Draft a email follow-up | Improved | Sonnet drafted the email instead of creating an email generator; GLM dropped its unsolicited gadget offer. |
| Query top CRM opportunities | Improved | GLM returned the correct results without suggesting a dashboard gadget. |
| Create a slide deck | Improved | Sonnet and GLM gave shorter confirmations focused on the slides they created. |
| Interpret a bare company name | Improved | Sol and GLM asked about the user’s broader intent instead of assuming they wanted an app. |


packages/workshop-backend/src/agent.ts
```diff
@@ -664,18 +664,28 @@ export interface AgentHooks {
 // =======================================================================================
 // Agent system prompt and tool descriptions
 
+const COMMUNICATION_GUIDANCE = `
+# Communicating with users
+
+Be helpful, direct, and friendly. Users may not know or care about implementation details. Lead with the answer or outcome, keep explanations concise, and use familiar document, account, and gadget names. Do not mention source filenames in routine progress updates or confirmations. Refer to the gadget and the change.
```
AshishKumar4: > Users may not know or care about implementation details.

We can avoid this, I fear this may make the agent care less about the code/implementation quality if it thinks that user doesn't care either. 

Instead, it can simply be "avoid unnecessary jargon or technical speak unless user is interested. Be concise and clear"

ndisidore (approved):
Makes sense to lean the prompt less coding specific

packages/workshop-backend/src/agent.ts
```diff
@@ -664,18 +664,28 @@ export interface AgentHooks {
...
 
 # Workspaces
 
 You are working within a "workspace". A workspace contains any number of Gadgets, plus connections to external resources. Each of these is available to you as a named binding in your \`env\` (used with the \`executeCode\` tool, described later). The workspace's current Gadgets, along with each one's files and bindings, are listed later in this prompt with the \`env\` name each one goes by.
 
-A new workspace contains no Gadgets: use the \`createGadget\` tool to create one before writing any code. Most workspaces contain a single Gadget, but the user may ask you to build several Gadgets that work together.
+A new workspace contains no Gadgets. You can answer questions, read connected resources, and perform one-off tasks with \`executeCode\` without creating a Gadget. Create one only when the user's request or established context clearly calls for a new application or saved output. An empty workspace or a task that needs code is not by itself a reason to create one.
```
ndisidore: We lost a reference to the `createGadget` tool here.

```suggestion
A new workspace contains no Gadgets. You can answer questions, read connected resources, and perform one-off tasks with \`executeCode\` without creating a Gadget. Create one (via the `createGadget` tool) only when the user's request or established context clearly calls for a new application or saved output. An empty workspace or a task that needs code is not by itself a reason to create one.
```

AshishKumar4:
LGTM

PR #490, opened 2026-09-14
Mint MCP request ids from a strict-server-safe charset

johncarpenter:
`McpClient` mints JSON-RPC ids as `\${uuid}:\${n}` (`packages/mcp-shared/src/client.ts`). JSON-RPC 2.0 allows any string id, but the official Ruby MCP SDK validates string ids against `/\A[a-zA-Z0-9_-]+\z/` in `json_rpc_handler` **before** dispatching a method, and answers `-32600 Invalid Request` otherwise.

The colon fails that pattern, so every call to a Ruby-SDK server fails — starting with `initialize`. The server is unreachable, not degraded, and both `gatekeeper-mcp` and `gatekeeper-mcp-portal` are affected.

Reproduced against Harvest (`https://api.harvestapp.com/mcp`, a Rails app — `WWW-Authenticate: Bearer realm="Rack::OAuth2 Protected Resources"`): OAuth completes, the redirect back to the connector then fails with

```
McpProtocolError: MCP server rejected "initialize": Invalid Request
```

and reproduced locally by running the real `McpClient` against a server applying that same id rule.

**Change:** the separator, `:` → `-`. The per-instance UUID prefix that keeps concurrent clients sharing one transport session from colliding on id `1` is unchanged; only the joining character moves inside the accepted set.

**Test:** `mints request ids from characters strict servers accept` in `packages/mcp-shared/__tests__/client-pagination.test.ts`, beside the existing id-uniqueness test. Fails before the change (`expected '2484468d-…-27b1cae665c9:1' to match /^[A-Za-z0-9_-]+$/`), passes after.

Upstream-strictness note: the Ruby SDK's `DEFAULT_ALLOWED_ID_CHARACTERS` is narrower than JSON-RPC 2.0 requires, so there is an argument that it should be relaxed there too — but the client side is what unblocks users of every already-deployed Ruby-SDK server.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/490" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


johncarpenter:

PR #493, opened 2026-09-14
Re-check compaction after every agent step

AshishKumar4:
The compaction check ran once, at the start of a turn. A turn whose tool results grew the prompt past the provider limit failed mid-turn with a context error; the user pressed retry, and only that second turn compacted. Robert hit this on Sol at roughly 512K tokens.

`runAgent` now owns the loop the Overseer used to run around it: load the durable history, run one pass, publish any checkpoint the pass produced, go again until a pass finishes. The Overseer's side is one `await runAgent(...)` plus two hooks, `loadChatHistory` and `commitChatCompaction`, which are the two private helpers it already had. A pass ends early either because it compacted instead of prompting the model, as before, or, new here, because a persisted tool step left the next request over the compaction trigger. That decision sits in `shouldStopAfterTurn` beside the other stop reasons: the step's measured prompt plus its tool results as the model will see them (not the `details` copy), or, when the provider reported no usage, a reload regardless so the turn-start estimator measures the whole prompt, the same rule that check already follows. Cancellation, connection requests and pending approvals still end the turn ahead of compaction.

The 30-step cap on a turn is gone. It came from the ai-sdk's `stepCountIs(30)` and was carried over when the loop moved to pi-agent-core; nothing else needed it.

GPT-5.6 Sol, Luna and Terra also gain a `compactionInputBudget` of 272K, the point past which OpenAI prices their input higher. Compaction sizes against that instead of the 1.05M window. The window and output limit are unchanged; `getModelTokenLimits` returns the smaller of the two as `inputBudget`, so nothing else moves.

Tests: the GPT-5.6 limits in `agent-compaction.test.ts`, plus the full backend suite and integration. The post-step decision is five lines over two functions the suite already covers (`estimateProjectionTokens`, `shouldCompactChat`); the scripted-model harness reports a fixed token count and has no summarizer, so an end-to-end test of it would mean building that, and the incident is the end-to-end check. Build and lint clean.


AshishKumar4:
On the two Bonk points:

P2, chars/4 undercounting CJK: that is the estimator the turn-start check has always used; the post-step check adds the same estimate to the provider-measured total for the step, so the two checks agree and each step re-anchors on real usage. Not changing the estimator here. #494 caps what one result can contribute, which bounds the exposure.

P3, the active-file marker surviving a reload: real. runAgent now clears it before returning reloadForCompaction, the same thing a non-edit tool start does, so the rerun's fresh preview manager starts from the state the client sees.

AshishKumar4:
The SUGGESTED_MODELS doc: moved onto the export, with the per-field docs on the entry type where they belong.

AshishKumar4:
Bonk P2 (19:37) and Devin on the zero-usage case: both right, and I was wrong to remove it. The turn-start check at `agent.ts:2619` already treats `measuredTokens > 0` as the condition for trusting provider usage and estimates the whole prompt otherwise, so the post-step check dropping that rule made the two disagree. Restored.

While there, the decision moved out of its own exported helper and into `shouldStopAfterTurn` beside the other stop reasons, which is where it reads best; the helper's unit test went with it, for the reason in the PR body.

packages/workshop-backend/src/agent.ts
```diff
@@ -294,8 +294,21 @@ export type CompactionContext = {
 
   /** The total tokens reported for the last measured model step, or zero if none are available. */
   measuredTokens: number;
+
+  /**
+   * Model steps still allowed in this turn, decremented here. The caller owns it because a turn
+   * spans every compaction reload: a fresh count per `runAgent` call would let an agent that keeps
+   * reading run 30 steps, compact, and run 30 more without end.
+   */
+  stepBudget: {remaining: number};
```
kentonv, line 303: Do we actually even want a step limit? I think it was mandatory with the old AI SDK code, which is the only reason we set it. We should just remove it and get rid of all this enforcement logic.

packages/workshop-backend/src/agent.ts
```diff
@@ -294,8 +294,21 @@ export type CompactionContext = {
...
 };
 
+/** Why one call to runAgent returned to the Overseer's rerun loop. */
```
kentonv, line 306: Why does the overseer need to concern itself with the "rerun loop"? Could we encapsulate that within `runAgent()`  itself? (That is, can `runAgent()` contain the loop where it calls `runAgentInner()` or something?)

packages/workshop-backend/src/agent.ts
AshishKumar4: Done!

packages/workshop-backend/src/agent.ts
AshishKumar4: Done!

kentonv (approved):

PR #496, opened 2026-09-14
Account for system prompt growth when sizing compaction

AshishKumar4:
## What does this change?

Compaction sizing starts from the tokens the provider reported for the last measured step and
estimates only the messages added since (`agent.ts`, the `contextTokens` computation). That assumed
the system prompt was unchanged between turns. It is not: the prompt is rebuilt every turn from the
workspace's gadget list, connectable vendors and output formats, and #267 adds the ambient catalogs.
Growth there never entered the estimate, so a chat near its window could skip compaction and send a
prompt the provider rejects. Because a failed request records no usage, the stale baseline then
repeats on the next turn.

The step's system prompt length is now stored beside its token total
(`StoredChatMetadata.measuredSystemPromptChars`, written in `addChatMessages`, cleared at both
`totalTokens` reset sites, stripped from client deliveries in `chatMetaForClient`). The next turn
adds the growth against the current length. A shrink is not credited: the usage is exact and the
four-characters-per-token ratio is a heuristic, so subtracting could under-count; over-counting only
compacts earlier, and the next measured step re-baselines. A step measured before the length was
recorded prices the prompt as unchanged, which is today's behavior.

The estimate moves out of `runAgent` into `estimateContextTokens` in `agent-compaction.ts`.

The helper reproduces the previous estimate exactly (same measured gate, same post-measurement
filter including the measured record's tool results) and adds one non-negative term. The unit tests
cover the unmeasured path, the measured path, growth, shrink, and a measurement taken before the
length was recorded. The stored field is written only with `totalTokens`, cleared wherever
`totalTokens` is, and never reaches clients.

Run: `pnpm --filter @gadgets/workshop-backend exec vitest run __tests__/agent-compaction.test.ts`

## Checklist

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/496" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


AshishKumar4:
Closing. The gap this fixes is bounded by the 15 percent trigger headroom and the output reservation, and it self-corrects at the next measured step, which includes the grown prompt. Prompt growth per turn is tens to a few hundred tokens; even a whole Context catalog at its caps is under 10k. No incident motivated it, and the fix adds a persisted field, a threaded `measured` object, and a heuristic on top of an exact count for a case nobody has hit. Reopen if a context-length failure is ever traced to system prompt growth.

PR #505, opened 2026-09-15
Fix blueprint configurator readiness

dcartertwo:
Blueprint resource connections using a configurator without a custom readiness check could not be saved because the generated iframe sent no readiness signal.

These configurators now report ready after a successful render. Failed or changing configurator state remains blocked from saving.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/cloudflare-os/pull/505" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


Maximo-Guk (approved):

ndisidore (approved):
Good find. Looks like a true RCA fix and not a baid-aid.
✔️

